PR #338, opened 2024-04-10
feat: add user invite management

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


packages/api/src/router/invite.ts
```diff
@@ -0,0 +1,70 @@
...
+        creator: {
+          columns: {
+            id: true,
+            name: true,
+          },
+        },
+      },
+    });
+    return dbInvites;
+  }),
```
SeDemal, line 27: Aren't we missing the "byId" procedure?
Or is the check done by going against all existing invites?

Meierschlumpf: It's not yet implemented that you can join Homarr with an invite url. This is only the management of it

apps/nextjs/src/app/[locale]/manage/users/invites/_components/invite-copy-modal.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export const InviteCopyModal = createModal<RouterOutputs["invite"]["create"]>(
+  ({ actions, innerProps }) => {
+    const t = useScopedI18n("management.page.user.invite");
+    const inviteUrl = useInviteUrl(innerProps);
+
+    return (
+      <Stack>
+        <Text>{t("action.copy.description")}</Text>
+        {/* FIXME: When next-international v2 is released the descriptions bold element can be implemented, see https://github.com/QuiiBz/next-international/pull/361 for progress */}
```
Comment (manuel-rw) [inline apps/nextjs/src/app/[locale]/manage/users/invites/_components/invite-copy-modal.tsx:17]: Rename to TODO?

packages/api/src/router/invite.ts
```diff
@@ -0,0 +1,70 @@
...
+import { TRPCError } from "@trpc/server";
+
+import { asc, createId, eq } from "@homarr/db";
+import { invites } from "@homarr/db/schema/sqlite";
+import { z } from "@homarr/validation";
+
+import { createTRPCRouter, protectedProcedure } from "../trpc";
+
+export const inviteRouter = createTRPCRouter({
+  all: protectedProcedure.query(async ({ ctx }) => {
```
manuel-rw, line 11: Action is unclear. Rename to ``getAll``

packages/api/src/router/invite.ts
```diff
@@ -0,0 +1,70 @@
...
+          columns: {
+            id: true,
+            name: true,
+          },
+        },
+      },
+    });
+    return dbInvites;
+  }),
+  create: protectedProcedure
```
manuel-rw, line 28: Rename to ``createInvite``

packages/api/src/router/invite.ts
```diff
@@ -0,0 +1,70 @@
...
+        creatorId: ctx.session.user.id,
+        token,
+      });
+
+      return {
+        id,
+        token,
+      };
+    }),
+  delete: protectedProcedure
```
manuel-rw, line 50: Rename to ``deleteBoard``

manuel-rw (changes requested):

Meierschlumpf:
Regarding the renaming of those router endpoints. I understand why you want to change it but on the same time it is kind of unnecessary as you'll always call them with `clientApi.invite.create` which already contains what you create and so it's kind of a duplicate

manuel-rw:
> Regarding the renaming of those router endpoints. I understand why you want to change it but on the same time it is kind of unnecessary as you'll always call them with `clientApi.invite.create` which already contains what you create and so it's kind of a duplicate

No, this is not my point. Consider the following endpoints: ``clientApi.invite.create`` and ``clientApi.invite.createTemporaryToken`` (note: these are illustrative names / ideas). So what does the first one do? Yes, the ``invite`` router does indicate it roughly, but since there are now to names, it is no longer clear.

manuel-rw (approved):

PR #464, opened 2024-05-10
fix: reject already existing usernames #443

### LINKED ISSUE #443: bug: multiple users with the same name can exist

-

### Describe the bug

There can be multiple users with the same name:
image

### Link to reproduction

-

### To reproduce

Create two users with the same name or rename an existing one to a name already occupied by another one

### Additional information

_No response_

ajnart:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

### Motivation
Closes #443 

### Changes
- Add a check when creating a user to see if a user with that `name` already exists. If that's the case then an error is thrown. ( https://github.com/homarr-labs/homarr/commit/781247de5141f4b0532801194ac904763db7da39 ) 
- Force lowercase usernames ( https://github.com/homarr-labs/homarr/commit/400fa3d480842ed099cfd0ae6b29e74d2d82d446 )


apps/nextjs/src/app/[locale]/manage/users/[userId]/_components/profile.accordion.tsx
```diff
@@ -5,22 +5,46 @@ import { Button, Stack, TextInput } from "@mantine/core";
...
-    },
-  });
+  const { mutate, isPending, isError, error } =
+    clientApi.user.editProfile.useMutation({
+      onError(error) {
+        showErrorNotification({
+          title: t(
+            "management.page.user.edit.section.profile.editProfile.title",
+          ),
+          message: error.message,
```
Comment (Meierschlumpf) [inline apps/nextjs/src/app/[locale]/manage/users/[userId]/_components/profile.accordion.tsx:31]: I would like it more if we would show the user a specific error message that we can translate

apps/nextjs/src/components/utils.tsx
```diff
@@ -0,0 +1,33 @@
+import type { AlertProps } from "@mantine/core";
```
Meierschlumpf, line 1: I'm not so much a fan of this:
image

I like a simple notification more, do we need such an error display?

apps/nextjs/src/components/utils.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+interface ErrorDisplayProps extends AlertProps {
+  title?: string;
+  hidden?: boolean;
+  message?: string;
+  icon?: React.ReactNode;
+}
+
+export function ErrorDisplay({
+  title = "There was an error",
```
Meierschlumpf, line 13: Here a translation would be missing

packages/api/src/router/user.ts
```diff
@@ -32,6 +32,15 @@ export const userRouter = createTRPCRouter({
   create: publicProcedure
     .input(validation.user.create)
     .mutation(async ({ ctx, input }) => {
+      const user = await ctx.db.query.users.findFirst({
+        where: eq(users.name, input.username),
+      });
+      if (user !== undefined) {
+        throw new TRPCError({
+          code: "FORBIDDEN",
```
Meierschlumpf, line 40: I would suggest the code `CONFLICT`

packages/api/src/router/user.ts
```diff
@@ -32,6 +32,15 @@ export const userRouter = createTRPCRouter({
   create: publicProcedure
     .input(validation.user.create)
     .mutation(async ({ ctx, input }) => {
+      const user = await ctx.db.query.users.findFirst({
+        where: eq(users.name, input.username),
```
Meierschlumpf, line 36:
```suggestion
        where: eq(users.name, input.username.toLowerCase()),
```

packages/api/src/router/user.ts
```diff
@@ -81,7 +90,16 @@ export const userRouter = createTRPCRouter({
         .from(users)
         .where(eq(users.id, input.userId))
         .limit(1);
+      const existingUser = await ctx.db.query.users.findFirst({
+        where: eq(users.name, input.form.name),
```
Meierschlumpf, line 94:
```suggestion
        where: eq(users.name, input.form.name.toLowerCase()),
```

packages/api/src/router/user.ts
```diff
@@ -81,7 +90,16 @@ export const userRouter = createTRPCRouter({
...
+      });
 
+      if (existingUser !== undefined) {
+        throw new TRPCError({
+          code: "FORBIDDEN",
```
Meierschlumpf, line 99: I would suggest the code `CONFLICT`

packages/api/src/router/user.ts
```diff
@@ -81,7 +90,16 @@ export const userRouter = createTRPCRouter({
         .from(users)
         .where(eq(users.id, input.userId))
         .limit(1);
+      const existingUser = await ctx.db.query.users.findFirst({
+        where: eq(users.name, input.form.name),
+      });
 
+      if (existingUser !== undefined) {
```
Meierschlumpf, line 97: I would not show an error when you rename to your current username, you can simply check if the existing users id is not the updated users id

Meierschlumpf (dismissed):

apps/nextjs/src/components/utils.tsx
ajnart: I can reformat it to an error at put it in the lib files close to the `showSuccessNotification` . I also think a notification makes more sense but I'm afraid some people might miss it. I will try some experiments to make it really visible

apps/nextjs/src/app/[locale]/manage/users/[userId]/_components/profile.accordion.tsx
Comment (ajnart) [reply apps/nextjs/src/app/[locale]/manage/users/[userId]/_components/profile.accordion.tsx:31]: This handles every error, I think translating server-side error messages a bit too overkill (the already user exist comes from the procedure, I also got a "fetch failed" error when the server was killed)

packages/api/src/router/user.ts
ajnart: It's not there that I checked, it's below in the condition that makes sure the form is dirty before it allows pressing the save button (i.e. the values are not the default ones anymore)

apps/nextjs/src/app/[locale]/manage/users/[userId]/_components/profile.accordion.tsx
```diff
@@ -50,7 +68,11 @@ export const ProfileAccordion = ({ user }: ProfileAccordionProps) => {
           label={t("user.field.email.label")}
           {...form.getInputProps("email")}
         />
-        <Button type="submit" disabled={!form.isValid()} loading={isPending}>
+        <Button
+          type="submit"
+          disabled={!form.isValid() || !form.isDirty()}
```
Comment (ajnart) [inline apps/nextjs/src/app/[locale]/manage/users/[userId]/_components/profile.accordion.tsx:73]: Here to disallow submitting an edit request when the form is not touched

Comment (Meierschlumpf) [reply apps/nextjs/src/app/[locale]/manage/users/[userId]/_components/profile.accordion.tsx:73]: You may have to run `setInitialValues` on the form with the new values after a successful update so it's no longer dirty and has the correct initial values

apps/nextjs/src/app/[locale]/manage/users/[userId]/_components/profile.accordion.tsx
Comment (Meierschlumpf) [reply apps/nextjs/src/app/[locale]/manage/users/[userId]/_components/profile.accordion.tsx]: Yes but errors that we handle ourself (Like existing user) would make sense to give the user feedback in their own language

packages/api/src/router/user.ts
```diff
@@ -81,7 +90,16 @@ export const userRouter = createTRPCRouter({
         .from(users)
         .where(eq(users.id, input.userId))
         .limit(1);
+      const existingUser = await ctx.db.query.users.findFirst({
+        where: eq(users.name, input.form.name.toLowerCase()),
```
manuel-rw, line 94: Is this case insensitive?

manuel-rw (dismissed):

packages/api/src/router/user.ts
ajnart: not sure, kind of want to change the create user method to .toLowercase() usernames just in case

ajnart:
@manuel-rw @Meierschlumpf we fix conflicts then merge ?

Meierschlumpf:
when my part is okay then yes

Meierschlumpf (approved):

Meierschlumpf:
> Please use squash commits instead of merge commits for features 😉

I was able to restrict pull requests to dev to only allow squash

PR #468, opened 2024-05-10
feat: add pre-commit hook

ajnart:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>
## Changes
Adds husky pre-commit hook that will lint/format/typecheck the codebase before allowing a commit

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+turbo run typecheck lint format && pnpm test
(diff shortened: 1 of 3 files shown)


Meierschlumpf:
Is there a way to get arround the pre commit hook? Like when I have an unfinished state that I want to commit and push. Generally I like the idea to push stuff that is not quite ready as I want to have savepoints

manuel-rw:
I agree with @Meierschlumpf here. I would hate to loose code because the hooks prevented me from pushing unfinished code

ajnart:
> Is there a way to get arround the pre commit hook? Like when I have an unfinished state that I want to commit and push. Generally I like the idea to push stuff that is not quite ready as I want to have savepoints

you can --no-verify or -n like `git commit -nm "Hello world"`

manuel-rw:
Most GUIs will verify by default. IMO (code) analysis should primarily be done on server. Is there a way to deactivate this by default? Then I would not be against it

PR #523, opened 2024-05-19
perf: improve typescript performance

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

image
As it can be seen above typescript spends a lot of type comparing the two types of databases. That's the reason I used the `as unknown as never` conversation so it hsa not to do that


manuel-rw (approved):
this is ridicilous.... 😅

PR #526, opened 2024-05-20
chore: Update trpc logger info

ajnart:
It now includes the username of the person doing the tRPC request (because the user id is not enough)


Meierschlumpf (approved):

ajnart:
👍🏼

PR #580, opened 2024-05-30
feat: add simple app ping

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

image
image
image


apps/tasks/src/lib/queue/worker.ts
```diff
@@ -14,7 +15,13 @@ export const queueWorkerAsync = async () => {
   for (const execution of executions) {
     const queue = queueRegistry.get(execution.name);
     if (!queue) continue;
-    await queue.callback(execution.data);
+
+    try {
+      await queue.callback(execution.data);
+    } catch (err) {
+      logger.error(err);
```
manuel-rw, line 22: Add more info (eg. that this error occurred in a worker)

apps/tasks/src/lib/cron-job/creator.ts
```diff
@@ -9,11 +10,19 @@ interface CreateCronJobOptions {
 export const createCronJob = (cronExpression: string, options: CreateCronJobOptions = { runOnStart: false }) => {
   return {
     withCallback: (callback: () => MaybePromise<void>) => {
+      const catchingCallbackAsync = async () => {
+        try {
+          await callback();
+        } catch (error) {
+          logger.error(error);
```
manuel-rw, line 17: Add more info (see other comment)

packages/ping/src/index.ts
```diff
@@ -0,0 +1,18 @@
+import { logger } from "@homarr/log";
+
+export const sendPingRequestAsync = async (url: string) => {
+  try {
+    return await fetch(url).then((response) => ({ statusCode: response.status }));
+  } catch (error) {
+    logger.error(error);
```
manuel-rw, line 7: Log where this error happened.

packages/ping/src/index.ts
```diff
@@ -0,0 +1,18 @@
...
+    if (error instanceof Error) {
+      return { error: error.message };
+    }
+
+    if (typeof error === "string") {
+      return { error };
+    }
+
+    return { error: "Unknown error" };
```
manuel-rw, line 16: Maybe it would be useful to put this in an exported function since it could be used elsewhere too?

packages/widgets/src/app/component.tsx
```diff
@@ -33,6 +35,19 @@ export default function AppWidget({ options, serverData, isEditMode, width, heig
     },
   );
 
+  const [pingResult, setPingResult] = useState<RouterOutputs["widget"]["app"]["ping"] | null>(null);
+
+  const shouldPingRun = Boolean(app?.href) && options.pingEnabled;
```
manuel-rw, line 40: Rename to shouldRunPing?

manuel-rw (changes requested):
Do I understand this correctly?

- The ping will only fail when there is an error within the fetch
- The status code is only used for the display in the frontend
- The ping will only be performed when it was registered over redis?

Can you attach also some screenshots?
Have you done some testing with invalid certificates, redirects, ...?

Meierschlumpf:
> Do I understand this correctly?
> 
> * The ping will only fail when there is an error within the fetch
> * The status code is only used for the display in the frontend
> * The ping will only be performed when it was registered over redis?
> 
> Can you attach also some screenshots? Have you done some testing with invalid certificates, redirects, ...? Do you want to push tests later?

The ping itself will only fail when there is an error with the fetch:
> Yes it will only show an error then. On the frontend it will check also if the statuscode was greater than 500 which would be a server error

The status code is only used for the display in the frontend:
> Yes and to check if it's greater than 500 to show a red dot if needed

The ping will only be performed when it was registered over redis:
> Yes it will only ping an url that was added to the redis list. If the websocket disconnects it will remove the url again.

I was not able to test it with invalid certificates and such. I'll try with a redirect within the next app now
I'll take a look on what tests make sense here within this pull request and add them accordingly

Meierschlumpf:
Okay redirect works

manuel-rw (approved):

manuel-rw (approved):

PR #649, opened 2024-06-09
test: add pi-hole test container integration test

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

### DIFF
--- a/packages/integrations/test/pi-hole.spec.ts
+++ b/packages/integrations/test/pi-hole.spec.ts
@@ -1,4 +1,5 @@
-import { GenericContainer, StartedTestContainer, Wait } from "testcontainers";
+import type { StartedTestContainer } from "testcontainers";
+import { GenericContainer, Wait } from "testcontainers";
 import { describe, expect, test } from "vitest";
 
 import { PiHoleIntegration } from "../src";
--- a/packages/integrations/test/pi-hole.spec.ts
+++ b/packages/integrations/test/pi-hole.spec.ts
@@ -1,33 +1,16 @@
-import { GenericContainer, Wait } from "testcontainers";
+import { GenericContainer, StartedTestContainer, Wait } from "testcontainers";
 import { describe, expect, test } from "vitest";
 
 import { PiHoleIntegration } from "../src";
 
+const DEFAULT_PASSWORD = "12341234";
+const DEFAULT_API_KEY = "3b1434980677dcf53fa8c4a611db3b1f0f88478790097515c0abb539102778b9"; // Some hash generated from password
+
 describe("Pi-hole integration", () => {
   test("getSummaryAsync should return summary from pi-hole", async () => {
     // Arrange
-    const password = "12341234";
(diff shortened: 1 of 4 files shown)


packages/integrations/test/pi-hole.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    const piholeContainer = await new GenericContainer("pihole/pihole:latest")
+      .withEnvironment({
+        WEBPASSWORD: password,
+      })
+      .withExposedPorts(80)
+      .start();
+
+    // We need to wait until pi hole has actually started and is ready to get requests
+    await sleep(5000);
```
manuel-rw, line 20: Isn't there a better way? (eg. healthchecks)
On fast systems this will be an artificial delay which will greatly add up if you run dozens of tests

packages/integrations/test/pi-hole.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+      ],
+      name: "Pi hole",
+      url: `http://${piholeContainer.getHost()}:${piholeContainer.getMappedPort(80)}`,
+    });
+
+    // Act
+    const result = await piHoleIntegration.getSummaryAsync();
+
+    // Assert
+    expect(result.adsBlockedToday).toBe(0);
```
manuel-rw, line 38: I would expect a test with a value other than 0 because 0 is a known default value in a lot of places

packages/integrations/test/pi-hole.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    // Cleanup
+    await piholeContainer.stop();
+  }, 20_000); // Timeout of 20 seconds
+});
+
+function sleep(ms: number) {
```
manuel-rw, line 45: Rename to async suffix

manuel-rw, line 45: Don't we have the same function in the project already?

manuel-rw (changes requested):

packages/integrations/test/pi-hole.spec.ts
Meierschlumpf: Yes there is a better way, I've added it now. Wasn't sure anymore how it has to be done and wanted to test if the test would run successfully on GitHub

packages/integrations/test/pi-hole.spec.ts
Meierschlumpf: Removed

Meierschlumpf: Removed

packages/integrations/test/pi-hole.spec.ts
Meierschlumpf: Yes that's true, this also why I made the domains blocked greater than one. The issue is that when you spin up a new instance of pi hole there is no blocked ad yet. I would suggest to leave it as it is with the 3 other values added now

manuel-rw (approved):

PR #663, opened 2024-06-13
feat: add calendar widget

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

image


apps/tasks/src/jobs/integrations/media-organizer.ts
```diff
@@ -0,0 +1,58 @@
...
+              },
+            },
+          },
+        },
+      },
+    },
+  });
+
+  for (const itemForIntegration of itemsForIntegration) {
+    const integration = itemForIntegration.integrations[0]?.integration;
```
Meierschlumpf, line 36: This should iterate through all iterations as well

Meierschlumpf: Maybe it also makes sense to create a map of all integrations with matching Id and then to convert it back to a list of integrations. Then it will also not fetch events for one integration multiple times:

```ts
const items = [{id: 1}, {id: 2}, {id: 1}, {id: 3}, {id: 4}, {id: 3}, {id: 2}, {id: 1}, {id: 4}, {id: 5}];
const itemsMap = new Map<number, Integration>();
items.forEach(item => {
    if (itemsMap.has(item.id)) return;

    itemsMap.set(item.id, item);
});
```

packages/api/src/router/widgets/calendar.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { createManyIntegrationMiddleware } from "../../middlewares/integration";
+import { createTRPCRouter, publicProcedure } from "../../trpc";
+
+export const calendarRouter = createTRPCRouter({
+  findAllEvents: publicProcedure
+    .unstable_concat(createManyIntegrationMiddleware("sonarr", "radarr", "readarr", "lidarr"))
+    .query(async ({ ctx }) => {
+      const data = ctx.integrations.map(async (integration) => {
+        const cache = createCacheChannel<CalendarEvent[]>(`calendar:${integration.id}`);
```
Meierschlumpf, line 12: I would suggest to move the creation of this cache channel to the redis package

packages/api/src/router/widgets/calendar.ts
```diff
@@ -0,0 +1,19 @@
...
+
+        return await cache.getAsync();
+      });
+      const a = await Promise.all(data);
```
Meierschlumpf, line 16: Return directly instead of assigning to a

packages/widgets/src/calendar/calender-day.tsx
```diff
@@ -0,0 +1,79 @@
...
+      opened={opened}
+      disabled={disabled}
+    >
+      <Popover.Target>
+        <Container
+          onClick={events.length > 0 && !opened ? open : close}
+          h="100%"
+          w="100%"
+          style={{
+            padding: "18% !important",
```
Meierschlumpf, line 40: Does this make sense to be percentage?

Meierschlumpf (changes requested):

apps/tasks/src/jobs/integrations/media-organizer.ts
```diff
@@ -0,0 +1,55 @@
...
+              },
+            },
+          },
+        },
+      },
+    },
+  });
+
+  for (const itemForIntegration of itemsForIntegration) {
+    for (const integration of itemForIntegration.integrations) {
```
Meierschlumpf, line 36: As already mentioned this will fetch for all integrations multiple times if items with the same integration exist, is this intentional?

Meierschlumpf (changes requested):

apps/tasks/src/jobs/integrations/media-organizer.ts
manuel-rw: No. Will fix 👍

manuel-rw: How do we deal with it if there are multiple widgets (items) that use this integration but with different options?
I would propose the following code for this Cron Job:
image
Simple options like numbers could be reduced to the max and then we could calculate in each procedure and exclude depending the item options - but that's quite a lot of effort.
What do you think?

packages/api/src/router/widgets/calendar.ts
manuel-rw: What exactly should I move? Can you elaborate

Meierschlumpf: Move 
```ts
export const createCalendarIntegrationChannel = (integrationId: string) => createCacheChannel<CalendarEvent[]>(`calendar:${integrationId}`);
```
to the redis package like the other channels we export there. Otherwise once the structure of the path changes it we could forget about one of them

apps/tasks/src/jobs/integrations/media-organizer.ts
manuel-rw: Maybe we should sent one per widget and one per integration instead? Distinct sadly also doesn't solve the problem (messages will be overwritten)

Meierschlumpf: Oh yeah you're right. Probably best to do so

packages/api/src/router/widgets/calendar.ts
```diff
@@ -0,0 +1,20 @@
...
+import { createManyIntegrationMiddleware } from "../../middlewares/integration";
+import { createTRPCRouter, publicProcedure } from "../../trpc";
+
+export const calendarRouter = createTRPCRouter({
+  findAllEvents: publicProcedure
+    .unstable_concat(createManyIntegrationMiddleware("sonarr", "radarr", "readarr", "lidarr"))
+    .query(async ({ ctx }) => {
+      return await Promise.all(
+        ctx.integrations.flatMap(async (integration) => {
+          for (const item of integration.items) {
```
Meierschlumpf, line 13: Shouldn't it just have another parameter for the itemId. Because it makes no sense to return the events for all items of this integration

Meierschlumpf (changes requested):

packages/api/src/router/widgets/calendar.ts
manuel-rw: I pushed another commit. Can you look whether this would be the correct approach?
Also, please comment on the TODO

packages/api/src/router/widgets/calendar.ts
```diff
@@ -0,0 +1,22 @@
+import type { CalendarEvent } from "@homarr/integrations/types";
+
+import { createItemWithIntegrationChannel } from "../../../../redis/src/lib/channel";
```
Meierschlumpf, line 3: Import should use redis package instead of relative path

packages/widgets/src/calendar/serverData.ts
```diff
@@ -0,0 +1,30 @@
...
+import type { RouterOutputs } from "@homarr/api";
+import { api } from "@homarr/api/server";
+
+import type { WidgetProps } from "../definition";
+
+export default async function getServerDataAsync({ integrationIds }: WidgetProps<"calendar">) {
+  try {
+    const data = await api.widget.calendar.findAllEvents({
+      integrationIds,
+      itemId: "" // TODO: Get actual item ID here
```
Meierschlumpf, line 12: You can add it in the `definition.ts` file of widgets package to the `WidgetProps` (just move the itemId from `WidgetComponentProps` up there). After that you only need to add it to the `server/runner.tsx` in widgets package

apps/tasks/src/jobs/integrations/media-organizer.ts
Meierschlumpf, line 1: Merge with tasks page pull request once that one is merged

Meierschlumpf (changes requested):

Meierschlumpf (approved):

PR #669, opened 2024-06-15
feat: test integration connection

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

- [x] Unit tests
- [x] System Integration tests for test connection implementations
- [x] Testing home assistant functionalities manually


packages/api/src/router/integration/integration-test-connection.ts
```diff
@@ -0,0 +1,95 @@
...
+  dbSecrets: {
+    kind: IntegrationSecretKind;
+    value: `${string}.${string}`;
+  }[] = [],
+) => {
+  const formSecrets = integration.secrets
+    .filter((secret) => secret.value !== null)
+    .map((secret) => ({
+      ...secret,
+      // Will never be null because of the check above, also will no longer be necessary in typescript 5.5
```
manuel-rw, line 25: Can you make a reminder / issue so we don't forget to do this?

packages/integrations/src/base/factory.ts
```diff
@@ -0,0 +1,16 @@
+import type { IntegrationKind } from "@homarr/definitions";
+
+import { HomeAssistantIntegration } from "../homeassistant/homeassistant-integration";
+import { PiHoleIntegration } from "../pi-hole/pi-hole-integration";
+import type { IntegrationInput } from "./integration";
+
+export const integrationFactory = (kind: IntegrationKind, integration: IntegrationInput) => {
```
manuel-rw, line 7: **NIT**
I dislike the name of this -> It is true that this function is a factory but I would expect this name to see as the class name. The process of instantiating is not called "integrationFactory".

packages/integrations/src/base/integration.ts
```diff
@@ -19,4 +28,86 @@ export abstract class Integration {
     }
     return secret.value;
   }
+
+  /**
+   * Test the connection to the integration
+   */
+  public abstract testConnectionAsync(): Promise<void>;
```
manuel-rw, line 35: Extend JSDoc: What happens if connection is invalid? Does it throw?

packages/integrations/src/homeassistant/homeassistant-integration.ts
```diff
@@ -53,21 +44,61 @@ export class HomeAssistantIntegration extends Integration {
...
+   * @param body the body of the request
+   * @returns the response from the API
+   */
+  private async postAsync(path: `/api/${string}`, body: Record<string, string>) {
+    return await fetch(appendPath(this.integration.url, path), {
+      headers: this.getAuthHeaders(),
+      body: JSON.stringify(body),
+      method: "POST",
+    });
+  }
```
manuel-rw, line 93: **NIT**
For me these functions do not bring much value. They abstract business logic into a function which makes it harder to read. Additionally abstraction seems not required here to me because the code is already very de-duplicated. You can leave this as it is but I would be careful with too much abstraction (you have to read the JSDoc to see what happens in this function)

packages/integrations/test/volumes/home-assistant-config.zip
manuel-rw, line 1: Have you ensured that no IPs from you are inside this file?

manuel-rw (changes requested):

packages/integrations/src/homeassistant/homeassistant-integration.ts
Meierschlumpf: IMO it makes it easier to use it and it's more centralized which allows to change less code when something changes

packages/integrations/test/volumes/home-assistant-config.zip
Meierschlumpf: No I'll check

Meierschlumpf: Checked it and it does not contain any IPs or location data. The long and lat are in amsterdam, the default city

manuel-rw (approved):

PR #672, opened 2024-06-16
feat: add jellyfin integration

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

image

image

### DIFF
--- a/packages/integrations/src/jellyfin/jellyfin-integration.ts
+++ b/packages/integrations/src/jellyfin/jellyfin-integration.ts
@@ -53,8 +53,8 @@ export class JellyfinIntegration extends Integration {
         sessionName: `${sessionInfo.Client} (${sessionInfo.DeviceName})`,
         user: {
           profilePictureUrl: `${this.integration.url}/Users/${sessionInfo.UserId}/Images/Primary`,
-          userId: sessionInfo.UserId ?? '',
-          username: sessionInfo.UserName ?? '',
+          userId: sessionInfo.UserId ?? "",
+          username: sessionInfo.UserName ?? "",
         },
         currentlyPlaying: nowPlaying,
       };
--- a/packages/integrations/src/jellyfin/jellyfin-integration.ts
+++ b/packages/integrations/src/jellyfin/jellyfin-integration.ts
@@ -53,8 +53,8 @@ export class JellyfinIntegration extends Integration {
         sessionName: `${sessionInfo.Client} (${sessionInfo.DeviceName})`,
         user: {
           profilePictureUrl: `${this.integration.url}/Users/${sessionInfo.UserId}/Images/Primary`,
-          userId: `${sessionInfo.UserId}`,
-          username: `${sessionInfo.UserName}`,
+          userId: sessionInfo.UserId ?? '',
+          username: sessionInfo.UserName ?? '',
         },
         currentlyPlaying: nowPlaying,
       };
--- a/packages/integrations/src/jellyfin/jellyfin-integration.ts
+++ b/packages/integrations/src/jellyfin/jellyfin-integration.ts
@@ -1,0 +1,68 @@
+import { Jellyfin } from "@jellyfin/sdk";
+import { getSessionApi } from "@jellyfin/sdk/lib/utils/api/session-api";
+import { getSystemApi } from "@jellyfin/sdk/lib/utils/api/system-api";
+
+import { Integration } from "../base/integration";
+import type { StreamSession } from "../interfaces/media-server/session";
+
+export class JellyfinIntegration extends Integration {
+  private readonly jellyfin: Jellyfin = new Jellyfin({
+    clientInfo: {
+      name: "Homarr",
+      version: "0.0.1",
+    },
+    deviceInfo: {
+      name: "Homarr",
+      id: "homarr",
+    },
+  });
+
+  public async testConnectionAsync(): Promise<void> {
+    const api = this.getApi();
+    const systemApi = getSystemApi(api);
+    await systemApi.getPingSystem();
+  }
+
+  public async getCurrentSessionsAsync(): Promise<StreamSession[]> {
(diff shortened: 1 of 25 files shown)


apps/tasks/src/jobs/integrations/media-server.ts
```diff
@@ -0,0 +1,46 @@
...
+    for (const integration of itemForIntegration.integrations) {
+      const jellyfinIntegration = new JellyfinIntegration({
+        ...integration.integration,
+        decryptedSecrets: integration.integration.secrets.map((secret) => ({
+          ...secret,
+          value: decryptSecret(secret.value),
+        })),
+      });
+      const streamSessions = await jellyfinIntegration.getCurrentSessionsAsync();
+      const cache = createCacheChannel(`media-server:${integration.integrationId}`);
```
Meierschlumpf, line 41: Does cache channel make sense here? Maybe pubSub channel makes mor sense?

apps/tasks/src/jobs/integrations/media-server.ts
```diff
@@ -0,0 +1,46 @@
+import { decryptSecret } from "@homarr/common";
+import { db, eq } from "@homarr/db";
+import { items } from "@homarr/db/schema/sqlite";
+import { JellyfinIntegration } from "@homarr/integrations";
+import { createCacheChannel } from "@homarr/redis";
+
+import { EVERY_5_SECONDS } from "~/lib/cron-job/constants";
+import { createCronJob } from "~/lib/cron-job/creator";
+
+export const mediaServerJob = createCronJob(EVERY_5_SECONDS).withCallback(async () => {
```
Meierschlumpf, line 10: Are you sure that you want to run this every 5 seconds. Seems quite expensive

packages/integrations/src/jellyfin/jellyfin-integration.ts
```diff
@@ -0,0 +1,57 @@
+import { Jellyfin } from "@jellyfin/sdk";
+import { getSessionApi } from "@jellyfin/sdk/lib/utils/api/session-api";
+
+import { Integration } from "../base/integration";
+import type { StreamSession } from "../interfaces/media-server/session";
+
+const jellyfin = new Jellyfin({
```
Meierschlumpf, line 7: Are you sure that you want to initalize a new instance of jellyfin here? This code will run always when anything from homarr/integrations is imported. If this connects to Jellyfin it should be moved inside the JellyfinIntegration class

packages/integrations/src/jellyfin/jellyfin-integration.ts
```diff
@@ -0,0 +1,57 @@
...
+          episodeCount: sessionInfo.NowPlayingItem.EpisodeCount,
+        };
+      }
+
+      return {
+        sessionId: `${sessionInfo.Id}`,
+        sessionName: `${sessionInfo.Client} (${sessionInfo.DeviceName})`,
+        user: {
+          profilePictureUrl: null,
+          userId: `${sessionInfo.UserId}`,
```
Meierschlumpf, line 50: Not sure but this seems completly unnecessary

packages/redis/src/lib/channel.ts
```diff
@@ -144,6 +144,20 @@ export const createCacheChannel = <TData>(name: string, cacheDurationMs: number
 
       return parsedData;
     },
+    subscribeAsync: async (callback: (data: TData) => void) => {
```
Meierschlumpf, line 147: As already mentioned pub sub channel makes IMO more sense

packages/widgets/src/media-server/component.tsx
```diff
@@ -0,0 +1,103 @@
...
+          const previousStreamIndex = previousCurrentStreams.findIndex(
+            (stream) => stream.sessionId === integrationPair.sessionId,
+          );
+          if (previousStreamIndex === -1) {
+            previousCurrentStreams.push(integrationPair);
+            continue;
+          }
+          previousCurrentStreams[previousStreamIndex] = integrationPair;
+        }
+        setCurrentStreams(previousCurrentStreams);
```
Meierschlumpf, line 61: Won't this never remove finished streams that do no longer exist?

Meierschlumpf (changes requested):

apps/tasks/src/jobs/integrations/media-server.ts
manuel-rw: Isn't Redis all memory? Writing every 5 seconds to memory doesn't seem excessive to me.

Meierschlumpf: Okay sure, just wanted to check because we add the 5 seconds thing to test it normally. But I guess it will be fine. Otherwise we can adjust it later easily

packages/widgets/src/media-server/component.tsx
manuel-rw, line 1: There are major performance issues with the widget. This should be resolved before we merge.

packages/redis/src/lib/channel.ts
```diff
@@ -1,6 +1,7 @@
 import superjson from "superjson";
 
 import { createId } from "@homarr/db";
+import { WidgetKind } from "@homarr/definitions";
```
Meierschlumpf, line 4: Fix import

Meierschlumpf (changes requested):

packages/integrations/src/jellyfin/jellyfin-integration.ts
manuel-rw: You mean this? All properties are nullable, even if the API doesn't seem to return null (unclear?)
image

Meierschlumpf: image
I think we should fallback to an empty string then

Meierschlumpf (approved):

PR #687, opened 2024-06-18
feat: add undici fetch interceptor

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

image

All credit goes to Meier- I only adjusted the format & logging

### DIFF
--- a/apps/tasks/package.json
+++ b/apps/tasks/package.json
@@ -33,7 +33,8 @@
     "@homarr/analytics": "workspace:^0.1.0",
     "dotenv": "^16.4.5",
     "node-cron": "^3.0.3",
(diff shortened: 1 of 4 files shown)


Meierschlumpf (approved):
Maybe `debug` log level makes more sense so the logs are not spammed

manuel-rw:
IMO info is fine here because...
- most people will not know how to change the log level
- network requests are crucial information
- we don't perform thousands of requests. The noise from Nextjs will be more than this one.
However, I'm open to discussions and we can easily change this in a future commit. I'll merge it for now so we can make use of it while developing.

PR #727, opened 2024-06-28
feat: improve calendar ux/ui

SeDemal:
image

### DIFF
--- a/packages/widgets/src/calendar/calender-day.tsx
+++ b/packages/widgets/src/calendar/calender-day.tsx
@@ -64,17 +64,24 @@ export const CalendarDay = ({ date, events, disabled }: CalendarDayProps) => {
   );
 };
 
-interface ArrIndicatorProps {
+interface NotificationIndicatorProps {
   events: CalendarEvent[];
 }
 
-const NotificationIndicator = ({ events }: ArrIndicatorProps) => {
+const NotificationIndicator = ({ events }: NotificationIndicatorProps) => {
   const notificationEvents = [...new Set(events.map((event) => event.links[0]?.notificationColor))].filter(String);
   return (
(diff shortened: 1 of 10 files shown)


packages/widgets/src/calendar/calender-day.tsx
```diff
@@ -59,20 +64,19 @@ export const CalendarDay = ({ date, events, disabled }: CalendarDayProps) => {
...
+interface ArrIndicatorProps {
   events: CalendarEvent[];
-  children: ReactNode;
-  position: IndicatorProps["position"];
 }
 
-const DayIndicator = ({ size, color, events, children, position }: DayIndicatorProps) => {
-  if (events.length === 0) return children;
-
+const NotificationIndicator = ({ events }: ArrIndicatorProps) => {
```
manuel-rw, line 71: Interface suggests that this component only applies for Arr apps. But this is not the case. Rename ``NotificationIndicator`` or ``ArrIndicatorProps``

packages/widgets/src/calendar/calender-day.tsx
```diff
@@ -59,20 +64,19 @@ export const CalendarDay = ({ date, events, disabled }: CalendarDayProps) => {
...
+  const notificationEvents = [...new Set(events.map((event) => event.links[0]?.notificationColor))].filter(String);
```
manuel-rw, line 72: Will there always be at least one link? I doubt

manuel-rw (changes requested):

packages/widgets/src/calendar/calender-day.tsx
SeDemal: No, and that will create an undefined which will get filtered out

manuel-rw (approved):

PR #750, opened 2024-07-05
fix: memory leak caused by many unclosed redis subscriptions

### LINKED ISSUE #744: bug: memory leak with cleaning up subscriptions and tRPC observables
When an observable is closed and terminated, the callback for the subscription is never cleaned up and the subscription will remain.
Since new observable are created constantly upon each reload, we also create new redis subscriptions each time on the same clients.
Clients are limited to 10'000 on Redis and subscriptions are limited to 2^32 which is a limit that we are unlikely to reach but the memory and CPU usage due to the constant if statements being run could increase nevertheless.
Only workaround would be to restart Homarr entirely.
Memory leak size is unknown.

See https://github.com/homarr-labs/homarr/pull/743

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Closes #744

### DIFF
--- a/packages/redis/src/lib/subscription-manager.ts
+++ b/packages/redis/src/lib/subscription-manager.ts
@@ -1,0 +1,83 @@
+import { randomUUID } from "crypto";
+
+import type { MaybePromise } from "@homarr/common/types";
+import { logger } from "@homarr/log";
+
+import { createRedisConnection } from "./connection";
+
+type SubscriptionCallback = (message: string) => MaybePromise<void>;
+
+export class SubscriptionManager {
+  private static subscriptions = new Map<string, Map<string, SubscriptionCallback>>();
+  private static redis = createRedisConnection();
+  private static listenerActive = false;
+
+  /**
+   * Subscribes to a channel.
+   * @param channelName name of the channel
+   * @param callback callback function to be called when a message is received
+   * @returns a function to unsubscribe from the channel
+   */
+  public static subscribe(channelName: string, callback: SubscriptionCallback) {
+    logger.debug(`Adding redis channel callback channel='${channelName}'`);
+
+    // We only want to activate the listener once
+    if (!this.listenerActive) {
+      this.activateListener();
+      this.listenerActive = true;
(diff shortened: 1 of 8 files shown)


packages/redis/src/lib/subscription-manager.ts
```diff
@@ -0,0 +1,83 @@
+import { randomUUID } from "crypto";
+
+import type { MaybePromise } from "@homarr/common/types";
+import { logger } from "@homarr/log";
+
+import { createRedisConnection } from "./connection";
+
+type SubscriptionCallback = (message: string) => MaybePromise<void>;
+
+export class SubscriptionManager {
```
manuel-rw, line 10: Add documentation what this does

manuel-rw, line 10: ``Manager`` is a poorly chosen name

packages/redis/src/lib/subscription-manager.ts
```diff
@@ -0,0 +1,83 @@
...
+    return () => {
+      logger.debug(`Removing redis channel callback channel='${channelName}' id='${id}'`);
+
+      const channelSubscriptions = this.subscriptions.get(channelName);
+      if (!channelSubscriptions) return;
+
+      channelSubscriptions.delete(id);
+
+      // If there are no subscriptions to the channel, unsubscribe from it
+      if (channelSubscriptions.size === 0) {
```
manuel-rw, line 54: Invert to reduce nesting?

manuel-rw (changes requested):

packages/redis/src/lib/subscription-manager.ts
Meierschlumpf: What do you think of `ChannelSubscriptionTracker`?
If not any better ideas?

manuel-rw: lgtm 👍

manuel-rw (approved):

PR #826, opened 2024-07-17
fix: Manage banner animation hitch fix

SeDemal:
fix the animation loop hitch in the banner of the management page.
Don't ask me why 50.8%, I just starred at my screen very long until the value caused the hitch to disappear.


manuel-rw (approved):
lol 😂

PR #844, opened 2024-07-21
feat: downloads widget

manuel-rw:
From PR #895

image
image
(Screenshot only include test files and linux ISO, therefore safe to use)

Usenet and Torrent downloads in 1 widget.
sabNZBd, NzbGet, Deluge, qBitTorrent, and transmission support.
Columns can be reordered in Edit mode.
Sorting enabled.
Time uses Dayjs with auto translation.
Can pause/resume single items, clients, or all.
Can delete items (With option to delete assossiated files).
Clients list and details.
Include all filtering and processing for ratio from oldmarr torrent widget.
Invalidation of old data (older than 30 seconds) to show an integration is not responding anymore.

Misc (So many miscs):
Fixed validation error with multiText.
Fixed translation application for multiSelect to behave the same as select.
Added background to gitignore (I needed to add a background to visually test opacity, probably will in the future too)
Added setOptions to frontend components so potential updates made from the Dashboard can be saved.
Extracted background and border color to use in widgets.
humanFileSize function based on the si format (powers of 1024, not 1000).
Improved integrationCreatorByKind by @Meierschlumpf.
Changed integrationCreatorByKind to integrationCreator so it functions directly from the integration.
Added integrationCreatorFromSecrets to directly work with secrets from db.
Added getIntegrationKindsByCategory to get a list of integrations sharing categories.
Added IntegrationKindByCategory type to get the types possible for a category (Great to cast on integration.kind that isn't already properly limited/typed but for which we know the limitation)
Added a common AtLeastOneOf type. Applied to TKind and IntegrationSecretKind[] where it was already being used and Added to the getIntegrationKindsByCategory's output to be more freely used.
Added the Modify type, instead of omiting to then add again just to change a parameters type, use the modify instead. Applied code wide already.
Hook to get list of integration depending on permission level of user. (By @Meierschlumpf)


packages/widgets/src/usenet-downloads/component.tsx
```diff
@@ -0,0 +1,52 @@
+"use client";
+
+import type {WidgetComponentProps} from "../definition";
+import {clientApi} from "@homarr/api/client";
+import type {UsenetQueueItem} from "@homarr/integrations";
+import {Stack} from "@mantine/core";
+import {useListState} from "@mantine/hooks";
+import {useMemo} from "react";
+
+export default function VideoWidget({integrationIds, serverData}: WidgetComponentProps<"usenet-downloads">) {
```
hillaliy, line 10: Change VideoWidget to UsenetDownloadsWidget

Meierschlumpf:
The build error is happening because of a upstream dependency issue with crypto-hash, that we provided a suggestion to fix it in the used dependency for qBittorrent: https://github.com/scttcper/torrent-file/pull/77

SeDemal (approved):

packages/api/src/router/widgets/media-requests.ts
```diff
@@ -11,7 +12,9 @@ import { createTRPCRouter, protectedProcedure, publicProcedure } from "../../trp
 
 export const mediaRequestsRouter = createTRPCRouter({
   getLatestRequests: publicProcedure
-    .unstable_concat(createManyIntegrationOfOneItemMiddleware("query", "overseerr", "jellyseerr"))
+    .unstable_concat(
+      createManyIntegrationOfOneItemMiddleware("query", ...getIntegrationKindsByCategory("mediaRequest")),
```
Meierschlumpf, line 16: What about putting this into a seperate function as well?

packages/api/src/router/widgets/media-server.ts
```diff
@@ -8,7 +9,7 @@ import { createTRPCRouter, publicProcedure } from "../../trpc";
 
 export const mediaServerRouter = createTRPCRouter({
   getCurrentStreams: publicProcedure
-    .unstable_concat(createManyIntegrationMiddleware("query", "jellyfin", "plex"))
+    .unstable_concat(createManyIntegrationMiddleware("query", ...getIntegrationKindsByCategory("mediaService")))
```
Meierschlumpf, line 12: What about putting this into a seperate function as well?

packages/api/src/router/widgets/smart-home.ts
```diff
@@ -26,17 +27,17 @@ export const smartHomeRouter = createTRPCRouter({
     });
   }),
   switchEntity: publicProcedure
-    .unstable_concat(createOneIntegrationMiddleware("interact", "homeAssistant"))
+    .unstable_concat(createOneIntegrationMiddleware("interact", ...getIntegrationKindsByCategory("smartHomeServer")))
```
Meierschlumpf, line 30: What about putting this into a seperate function as well?

packages/integrations/test/nzbget.spec.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    // Assert
+    await expect(getAsync()).resolves.toMatchObject({ items: [{ ...item, state: "paused" }] });
+    await expect(actAsync()).resolves.not.toThrow();
+    await expect(getAsync()).resolves.toMatchObject({ items: [{ ...item, state: "downloading" }] });
+
+    // Cleanup
+    await startedContainer.stop();
+  }, 20_000); // Timeout of 20 seconds
+  // */
```
Meierschlumpf, line 161: Please remove this code, it will still be in the history if needed at some point

Meierschlumpf (changes requested):

Meierschlumpf (approved):

SeDemal (approved):

PR #871, opened 2024-07-27
feat: restrict non credential provider interactions

### LINKED ISSUE #712: feat: restrict non credentials interactions

The following things should be restricted to credentials only accounts:

- [x] Invites should only be available if credentials login enabled
- [x] Any manage functionallity on users (except language selection) should be disabled for non credentials accounts
- [x] Users of non credentials provider can not be added / removed from groups manually
- [x] Hide pages like `Security` page in user preferences for non credentials users

### Describe the solution you'd like to see

- [x] Add field `provider` to user in database
- [x] Implement above mentioned restrictions
- [x] Add alerts to inform user on why certain fields can not be modified (user preferences) / certain users membership on groups can not be managed.

### Additional information

_No response_

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Closes #712

image

image


apps/nextjs/src/app/[locale]/manage/users/[userId]/general/page.tsx
```diff
@@ -53,27 +54,37 @@ export default async function EditUserPage({ params }: Props) {
...
-        </Box>
-        <Box w={{ base: "100%", lg: 200 }}>
-          <UserProfileAvatarForm user={user} />
-        </Box>
-      </Group>
+      <Alert variant="light" color="yellow" icon={<IconExclamationCircle size="1rem" stroke={1.5} />}>
+        {t("management.page.user.fieldsDisabledExternalProvider")}
+      </Alert>
+
+      <Stack>
```
Comment (Meierschlumpf) [inline apps/nextjs/src/app/[locale]/manage/users/[userId]/general/page.tsx:65]: Inner stack can be removed

packages/api/src/router/user.ts
```diff
@@ -7,9 +7,12 @@ import { groupMembers, groupPermissions, groups, invites, users } from "@homarr/
 import { validation, z } from "@homarr/validation";
 
 import { createTRPCRouter, protectedProcedure, publicProcedure } from "../trpc";
+import { assertCredentialsEnabled } from "./invite/asserts";
 
 export const userRouter = createTRPCRouter({
   initUser: publicProcedure.input(validation.user.init).mutation(async ({ ctx, input }) => {
+    assertCredentialsEnabled();
```
manuel-rw, line 14: assert is a name that should only be used for test.
I suggest ``throwIfCredentialsDisabled()``

packages/api/src/router/user.ts
```diff
@@ -190,26 +212,34 @@ export const userRouter = createTRPCRouter({
...
-      }
+    if (dbUser.provider !== "credentials") {
+      throw new TRPCError({
+        code: "FORBIDDEN",
+        message: "Password can not be changed for users with external providers",
+      });
+    }
 
+    // Admins can change the password of other users without providing the previous password
+    const isPreviousPasswordRequired = ctx.session.user.id === input.userId;
```
manuel-rw, line 240: Can you log an information whether this was required or not?

manuel-rw (changes requested):

packages/api/src/router/user.ts
Meierschlumpf: I don't full agree. In my opinion it should be placed whereever we want to restrict the continous execution to something that's also why typescript code like this exists:

```ts
const assertNotNullable = <T>(input: T): asserts input is NonNullable<T> => {
  if (input === null || input === undefined) {
    throw new Error("Value is null or undefined");
  }
}
```

but as I don't use this assertion I'll rename it

manuel-rw: Okay, I see your point.
But regardless, ``throwIfXXXX()`` seems more descriptive than ``assertXXX`` in my opinion.
Up to you what name you want to keep

manuel-rw (approved):

PR #881, opened 2024-07-28
fix: fetch timeout for external requests to small

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


packages/common/src/fetch-with-timeout.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @returns fetch response
+ */
+export const fetchWithTimeout = (...[url, requestInit]: Parameters<typeof fetch>) => {
+  const controller = new AbortController();
+
+  // 10 seconds timeout:
+  const timeoutId = setTimeout(() => controller.abort(), 10000);
+
+  return fetch(url, { signal: controller.signal, ...requestInit }).then((response) => {
+    clearTimeout(timeoutId);
```
manuel-rw, line 14: Don't we need to clear the timeout on ``finally()`` instead?

manuel-rw (approved):

packages/common/src/fetch-with-timeout.ts
Meierschlumpf: Yes you're probably right!

manuel-rw (approved):

PR #901, opened 2024-08-01
fix: apply size to style directly to prevent warnings on size

SeDemal:
Settings tabler icons-react SVG's size parameter using relative units causes warning in console.
Use style instead. (Only affects SVG/icons)


packages/widgets/src/rssFeed/component.tsx
```diff
@@ -66,7 +66,7 @@ export default function RssFeed({ serverData, options }: WidgetComponentProps<"r
 
 const InfoDisplay = ({ date }: { date: string }) => (
   <Group gap="2.5cqmin">
-    <IconClock size="2.5cqmin" />
+    <IconClock style={{ height: "3cqmin", width: "3cqmin" }} />
```
manuel-rw, line 69: Can you provide more context why size doesn't work here and why you use the style attribute instead of height and width directly?

SeDemal: SVG's don't officially support relative sizing, says so in it's spec. Browsers knows that, found it stupid and allows for relative sizing when using the <svg /> component, but because it's not supposed to support it, it'll flood an error in the console of the browser. so to circumvent that, you need to set the size using the style parameter, which is anoying, especially since the size parameter exist but you just can't use it properly without it making a mess in the logs.

manuel-rw: Shouldn't we contribute this fix upstream then? (eg. Tabler Icons?)

SeDemal: Yeah... I agree, just out of laziness I didn't want to bother. But since I don't know how long this will take, and I don't want my console to be flooded, I'd suggest we do this in the meantime.

Meierschlumpf:
What is the state of this pr @SeDemal @manuel-rw ?

manuel-rw:
I still think that we should contribute upstream. My opinion hasn't changed

SeDemal:
Leave as a keepsake that we need to do it at some point.

Meierschlumpf:
I close this pr as we no longer use cqmin stuff

PR #915, opened 2024-08-04
feat: add support for app url variables

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


manuel-rw (dismissed):
lgtm, I'm generally not a fan of these variables though. All routing / differences in hosts should be handeled by the load balancer or proxies - not Homarr. However, since 0.15 a lot of users are relying on this feature so we should not remove it from the user.

manuel-rw (approved):

PR #930, opened 2024-08-06
feat: add column count and is public options to board creation modal

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

image


Meierschlumpf:
Will fix tests tomorrow

SeDemal (commented):
Should we really force a maximum?
We've had the comment that x amount of columns wasn't enough and I'm sure we'll get more.
I really do like the use of the slider more than an unlimited NumberInput.
I suggest limiting to 100 so people can think of it as percents. (That would be a big slider though)
What do you think?
(Also there's an unrelated error, not sure what's that about)

Meierschlumpf:
I would leave it for now at 24. When somebody really want's more we can add this feature later. I mean it's not really hard to do so

manuel-rw:
I think 24 is a good default to start with. We can still adjust this value later. Our early adopters will provide feedback and say whether they like this size

manuel-rw (approved):

PR #953, opened 2024-08-09
fix: DnsHole controls timer

hillaliy:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

### DIFF
--- a/packages/integrations/src/adguard-home/adguard-home-integration.ts
+++ b/packages/integrations/src/adguard-home/adguard-home-integration.ts
@@ -123,8 +123,7 @@ export class AdGuardHomeIntegration extends Integration implements DnsHoleSummar
     }
   }
 
-  public async disableAsync(duration?: number): Promise<void> {
-    duration = duration ? duration * 1000 : 0;
+  public async disableAsync(duration = 0): Promise<void> {
     const response = await fetch(`${this.integration.url}/control/protection`, {
       method: "POST",
       headers: {
@@ -133,7 +132,7 @@ export class AdGuardHomeIntegration extends Integration implements DnsHoleSummar
       },
       body: JSON.stringify({
         enabled: false,
-        duration: duration,
+        duration: duration * 1000,
       }),
     });
     if (!response.ok) {
--- a/packages/integrations/src/adguard-home/adguard-home-integration.ts
+++ b/packages/integrations/src/adguard-home/adguard-home-integration.ts
@@ -124,6 +124,7 @@ export class AdGuardHomeIntegration extends Integration implements DnsHoleSummar
   }
 
   public async disableAsync(duration?: number): Promise<void> {
+    duration = duration ? duration * 1000 : 0;
     const response = await fetch(`${this.integration.url}/control/protection`, {
       method: "POST",
       headers: {
--- a/packages/widgets/src/dns-hole/controls/TimerModal.tsx
+++ b/packages/widgets/src/dns-hole/controls/TimerModal.tsx
@@ -8,11 +8,11 @@ import { useI18n } from "@homarr/translation/client";
 interface TimerModalProps {
   opened: boolean;
   close: () => void;
-  integrationIds: string[];
+  selectedIntegration: string[];
   disableDns: (data: { duration: number; integrationId: string }) => void;
 }
 
-const TimerModal = ({ opened, close, integrationIds, disableDns }: TimerModalProps) => {
+const TimerModal = ({ opened, close, selectedIntegration, disableDns }: TimerModalProps) => {
   const t = useI18n();
   const [hours, setHours] = useState(0);
   const [minutes, setMinutes] = useState(0);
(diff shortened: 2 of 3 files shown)


packages/widgets/src/dns-hole/controls/TimerModal.tsx
```diff
@@ -8,11 +8,11 @@ import { useI18n } from "@homarr/translation/client";
 interface TimerModalProps {
   opened: boolean;
   close: () => void;
-  integrationIds: string[];
+  selectedIntegration: string[];
```
manuel-rw, line 11: This is the singular form. However, this is an array (hence pluaral should be used).
Also, ``integration`` sounds like an object. Please name it ``integrationIds`` since it passes the IDs

packages/widgets/src/dns-hole/controls/component.tsx
```diff
@@ -90,7 +91,17 @@ export default function DnsHoleControlsWidget({ options, integrationIds }: Widge
           </Tooltip>
 
           <Tooltip label={t("widget.dnsHoleControls.controls.setTimer")}>
-            <Button onClick={open} disabled={allDisabled} variant="light" color="yellow" fullWidth h="2rem">
+            <Button
+              onClick={() => {
+                setSelectedIntegration(integrationIds);
```
manuel-rw, line 96: Again, please use plural

packages/widgets/src/dns-hole/controls/component.tsx
```diff
@@ -149,7 +169,17 @@ const ControlsCard = (
...
             </UnstyledButton>
-            <ActionIcon disabled={!isEnabled} size={20} radius="xl" top="2.67px" variant="default" onClick={open}>
+            <ActionIcon
+              disabled={!isEnabled}
+              size={20}
+              radius="xl"
+              top="2.67px"
+              variant="default"
+              onClick={() => {
+                setSelectedIntegration([integrationId]);
```
manuel-rw, line 179: Use plural

manuel-rw (changes requested):

hillaliy:
I think it's can be merge

manuel-rw (approved):
LGTM. Please **wait for a second approval**

packages/integrations/src/adguard-home/adguard-home-integration.ts
```diff
@@ -124,6 +124,7 @@ export class AdGuardHomeIntegration extends Integration implements DnsHoleSummar
   }
 
   public async disableAsync(duration?: number): Promise<void> {
+    duration = duration ? duration * 1000 : 0;
```
SeDemal, line 127: Small possible improvement, you can just set the default value of duration right in the function's parameter definition. Instead of "(duration?: number)", you can do "(duration = 0)".
You then won't need to test for undefined.
And instead of doing the *1000 here, apply it down in the body of the request.
(0*1000 is still 0 so it's fine)

packages/widgets/src/dns-hole/controls/component.tsx
```diff
@@ -90,7 +91,17 @@ export default function DnsHoleControlsWidget({ options, integrationIds }: Widge
           </Tooltip>
 
           <Tooltip label={t("widget.dnsHoleControls.controls.setTimer")}>
-            <Button onClick={open} disabled={allDisabled} variant="light" color="yellow" fullWidth h="2rem">
+            <Button
+              onClick={() => {
+                setSelectedIntegrationIds(integrationIds);
```
SeDemal, line 96: We shouldn't disable integrations that are already disabled.
Oldmarr has a neet function that filters the integrations between the 2 subset.

Also, you did something that was quite nice in oldmarr, instead of setting the list of integrations, you just by default assumed it was all disabled integrations if a specific integration ID wasn't given. But that's not important, the first part matters more.

packages/widgets/src/dns-hole/controls/component.tsx
SeDemal: I unmarked the "resolve".
Either make the function accept only single element instead of an array, or make it plural, just because you use it with a single item doesn't make it singular. you still pass an array to it and It could be used as is otherwise.

Like I mentioned before though, if you made the state for single items to begin with and assume all integration when that state is unset, then you can set all of them to single and won't need to do specific things as such.

SeDemal (dismissed):

packages/widgets/src/dns-hole/controls/component.tsx
```diff
@@ -114,11 +123,24 @@ export default function DnsHoleControlsWidget({ options, integrationIds }: Widge
 
       <Stack gap="2.5cqmin" flex={1} justify={options.showToggleAllButtons ? "flex-end" : "space-evenly"}>
         {data.map((integrationData) =>
-          ControlsCard(integrationData.integrationId, integrationData.integrationKind, toggleDns, status, open, t),
+          ControlsCard(
```
Meierschlumpf, line 126: This should be a jsx component and not a function call

Meierschlumpf (changes requested):

Meierschlumpf (approved):
Now look good to me

PR #965, opened 2024-08-10
feat: Prowlarr integration

hillaliy:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


packages/integrations/src/interfaces/indexer-manager/indexer.ts
```diff
@@ -0,0 +1,7 @@
+export interface Indexer {
+  id: number;
+  name: string;
+  url: string;
+  enable: boolean;
```
manuel-rw, line 5: Rename to ``enabled``

packages/integrations/src/interfaces/indexer-manager/indexer.ts
```diff
@@ -0,0 +1,7 @@
...
+  status: boolean;
```
manuel-rw, line 6: What is the difference between ``status`` and ``enable``?

manuel-rw (changes requested):

packages/integrations/src/interfaces/indexer-manager/indexer.ts
hillaliy: Enable: when the user enable / disable the indexer

Status: when there is an error with the indexer site.

On old homarr we did it the same.
If one of the options are false the indexer is off.

manuel-rw: Can you document this using JSDoc?
https://jsdoc.app/about-getting-started

hillaliy: Yes

packages/integrations/src/prowlarr/prowlarr-integration.ts
```diff
@@ -0,0 +1,83 @@
+import { Integration } from "../base/integration";
+import { IntegrationTestConnectionError } from "../base/test-connection-error";
+import type { Indexer } from "../interfaces/indexer-manager/indexer";
+import { indexerResponseSchema, statusResponseSchema } from "./prowlarr-types";
+
+export class ProwlarrIntegration extends Integration {
+  public async getIndexersAsync(): Promise<Indexer[]> {
+    const apiKey = super.getSecretValue("apiKey");
+
+    const indexerResponse = await fetch(`${this.integration.url}/api/v1/indexer?apikey=${apiKey}`);
```
Meierschlumpf, line 10: I would suggest to use the `x-Api-Key` header that is supported by prowlarr:
https://prowlarr.com/docs/api/
image

packages/integrations/src/prowlarr/prowlarr-integration.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (!indexerResponse.ok) {
+      throw new Error(
+        `Failed to fetch indexers for ${this.integration.name} (${this.integration.id}): ${indexerResponse.statusText}`,
+      );
+    }
+
+    const statusResponse = await fetch(`${this.integration.url}/api/v1/indexerstatus?apikey=${apiKey}`);
```
Meierschlumpf, line 17: Does this one only return disabled indexers?

packages/integrations/src/prowlarr/prowlarr-integration.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    const disabledIndexerIds = new Set(statusResult.data.map((status: { id: number }) => status.id));
+
+    const indexers: Indexer[] = indexersResult.data.map((indexer) => ({
+      id: indexer.id,
+      name: indexer.name,
+      url: indexer.indexerUrls[0] ?? "",
+      enabled: indexer.enable,
+      status: disabledIndexerIds.has(indexer.id),
```
Meierschlumpf, line 47: Either the enabled flag or disabledIndexerIds name make no sense as then this would be interpreted as the same

Meierschlumpf (dismissed):

packages/integrations/src/prowlarr/prowlarr-integration.ts
hillaliy: Yes

packages/integrations/src/interfaces/indexer-manager/indexer.ts
```diff
@@ -0,0 +1,12 @@
...
+  name: string;
+  url: string;
+  enabled: boolean;
+  status: boolean;
+}
+/**
+ * Enabled: when the user enable / disable the indexer.
+ * Status: when there is an error with the indexer site.
+ * If one of the options are false the indexer is off.
+ */
```
manuel-rw, line 12: Almost correct.
Put this on a new line before the enabled row.
Like this:

```ts
interace MyInterface  {
	/**
	* This is a JsDoc comment
	*/
	name: string;
}
```

manuel-rw (changes requested):

SeDemal:
When do you think you'll be done making the widget for it?
Please set your PR's as draft if they are not done.

hillaliy:
> When do you think you'll be done making the widget for it? Please set your PR's as draft if they are not done.

The widget will be ready a day or two.
About draft, OK

Meierschlumpf:
LGTM, what about you @SeDemal @manuel-rw

SeDemal:
If there's a wish to have a separate PR for the widget I fine with it, that's how we've been doing with Mani until now and it make sense.
Like mentioned, I'm strongly against having code without purpose waiting for the widget to come. Especially since it means that it cannot be tested practically.
Especially if the integration changes depending on the use of the widget.
To note that this is specifically for new widget structure and that is not something that will happen often or at all once we have all the current widget ported over.
Stacked is good and shows the work that is being done and has been done.

hillaliy:
That was my intention a separate PR for the widget.

SeDemal:
> That was my intention a separate PR for the widget.

Stacked PR ≠= Separate PR.
Stacked PR would be to make a second PR based on the first branch without merging the first PR.

hillaliy:
Ok i'll do it next time

manuel-rw:
So do we merge now or not? IMO @hillaliy you should create a stacked PR. Can you do that?

hillaliy:
I have a problem to make stacked PR, I make pull to dev and I now when I go to Prowlarr branch I get an error. something with the DB.

manuel-rw (dismissed):

manuel-rw (approved):

PR #1053, opened 2024-08-31
feat: radarr integration

hillaliy:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

**Thank you for your contribution. Please ensure that your pull request meets the following pull request:**

- [x] Builds without warnings or errors (``pnpm buid``, autofix with ``pnpm format:fix``)
- [x] Pull request targets ``dev`` branch
- [x] Commits follow the [conventional commits guideline](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] No shorthand variable names are used (eg. ``x``, ``y``, ``i`` or any abbrevation)

image

PR #1057, opened 2024-09-02
feat: indexer manager widget

hillaliy:
Co-authored-by: homarr-renovate[bot] <158783068+homarr-renovate[bot]@users.noreply.github.com>

<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


hillaliy:
Screenshot 2024-09-02 at 17 27 07

packages/api/src/router/widgets/indexer-manager.ts
```diff
@@ -0,0 +1,67 @@
...
+import { createManyIntegrationMiddleware } from "../../middlewares/integration";
+import { createTRPCRouter, publicProcedure } from "../../trpc";
+
+export const indexerManagerRouter = createTRPCRouter({
+  indexers: publicProcedure
+    .unstable_concat(createManyIntegrationMiddleware("query", "prowlarr"))
+    .query(async ({ ctx }) => {
+      const results = await Promise.all(
+        ctx.integrations.map(async (integration) => {
+          let client;
```
manuel-rw, line 15: Commented out code is bad practice. At this time, the second integration has not been implemented and it makes no sense to already partially implement related logic here.

packages/api/src/router/widgets/indexer-manager.ts
```diff
@@ -0,0 +1,67 @@
+import { TRPCError } from "@trpc/server";
+
+import { ProwlarrIntegration } from "@homarr/integrations";
+import { logger } from "@homarr/log";
+
+import { createManyIntegrationMiddleware } from "../../middlewares/integration";
+import { createTRPCRouter, publicProcedure } from "../../trpc";
+
+export const indexerManagerRouter = createTRPCRouter({
+  indexers: publicProcedure
```
manuel-rw, line 10: This also get's the status, correct? If yes, I strongly suggest you to rename to getIndexerStatus or getIndexers

packages/translation/src/lang/en.ts
```diff
@@ -1060,6 +1060,20 @@ export default {
         unknown: "Unknown",
       },
     },
+    indexerManager: {
+      name: "Indexer manager status",
+      description: "Status about your indexers",
```
manuel-rw, line 1065: Status **of**

packages/translation/src/lang/en.ts
```diff
@@ -1060,6 +1060,20 @@ export default {
...
+      option: {
+        openIndexerSiteInNewTab: {
+          label: "Open Indexer Site In New Tab",
```
manuel-rw, line 1068: Grammar - Only the O should the uppercase here

packages/widgets/src/indexer-manager/component.tsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+  const t = useI18n();
+  const data = useMemo(() => (serverData?.initialData ?? []).flatMap((indexers) => indexers.indexers), [serverData]);
+
+  const [loading, { toggle }] = useDisclosure();
+  const { mutate: testAll } = clientApi.widget.indexerManager.testAllIndexers.useMutation({
+    onMutate: () => {
+      toggle();
+    },
+    onSettled: () => {
```
manuel-rw, line 28: Just as a question, doesn't this trigger always? So maybe the second one is not required?

manuel-rw (changes requested):
Looks pretty good. Congratulations. Your code is improving a lot 🙌🎉

manuel-rw:
Also, please make sure that Deepsource and all other builds pass before you mark as ready for review

packages/widgets/src/indexer-manager/component.tsx
hillaliy: It's toggle on / off so if I remove onSettled it's stay loading

packages/widgets/src/indexer-manager/component.tsx
```diff
@@ -0,0 +1,72 @@
...
+  integrationIds,
+  serverData,
+}: WidgetComponentProps<"indexerManager">) {
+  if (integrationIds.length === 0) {
+    throw new NoIntegrationSelectedError();
+  }
+  const t = useI18n();
+  const data = useMemo(() => (serverData?.initialData ?? []).flatMap((indexers) => indexers.indexers), [serverData]);
+
+  const [loading, { toggle }] = useDisclosure();
```
Meierschlumpf, line 23: Instead you can use `isPending` from the `useMutation` hook below. It reflects the loading state

packages/api/src/router/widgets/indexer-manager.ts
```diff
@@ -0,0 +1,61 @@
...
+    .unstable_concat(createManyIntegrationMiddleware("query", "prowlarr"))
+    .query(async ({ ctx }) => {
+      const results = await Promise.all(
+        ctx.integrations.map(async (integration) => {
+          let client;
+          switch (integration.kind) {
+            case "prowlarr":
+              client = new ProwlarrIntegration(integration);
+              break;
+          }
```
Meierschlumpf, line 20: As already described [here](https://github.com/homarr-labs/homarr/pull/1053#discussion_r1739918181) you can use the existing function that will be typesafe and not show a deepsource issue 😄

Meierschlumpf (changes requested):

manuel-rw:
> > Also, please make sure that Deepsource and all other builds pass before you mark as ready for review
> 
> The DeepSource error is because I didn't add default to "switch" but we don't need default the widget will support only two integrations (Prowlarr & Jackett)

This is untrue. You should always have a default case in a switch.
It's good practice to throw an error there because it should never run that code - and if it does something critical went wrong.
But as @Meierschlumpf mentioned, you can remove this switch anyway if you use the typesafe method he suggested.

SeDemal:
> > > Also, please make sure that Deepsource and all other builds pass before you mark as ready for review
> > 
> > 
> > The DeepSource error is because I didn't add default to "switch" but we don't need default the widget will support only two integrations (Prowlarr & Jackett)
> 
> This is untrue. You should always have a default case in a switch. It's good practice to throw an error there because it should never run that code - and if it does something critical went wrong. But as @Meierschlumpf mentioned, you can remove this switch anyway if you use the typesafe method he suggested.

Does it matter? Once dl widget gets merged we can implement the selector function meier made to all the widgets anyway.

packages/cron-jobs/src/jobs/integrations/indexer-manager.ts
```diff
@@ -0,0 +1,45 @@
...
+              },
+            },
+          },
+        },
+      },
+    },
+  });
+
+  for (const itemForIntegration of itemsForIntegration) {
+    const integration = itemForIntegration.integrations[0]?.integration;
```
manuel-rw, line 31: Check ``media-organizer.ts`` job for example for iteration. We should not take an index -> this will limit the user to one integration only. As a reminder, we want them to have multiple

packages/widgets/src/indexer-manager/component.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const { mutate: testAll, isPending } = clientApi.widget.indexerManager.testAllIndexers.useMutation();
+
+  clientApi.widget.indexerManager.subscribeIndexersStatus.useSubscription(
+    { integrationIds },
+    {
+      onData() {
+        setIndexersData((data) => {
+          return [...data];
+        });
```
manuel-rw, line 35: Don't you want to take the data from the ``onData(T)`` function?
Right now there will not be any data being updated. Please fix.

manuel-rw (changes requested):

Meierschlumpf (dismissed):

manuel-rw (approved):

Meierschlumpf (approved):

PR #1108, opened 2024-09-11
refactor: Remove typed-rpc dependency and used fetch instead

SeDemal:
Fetch method POST achieves basically the same as typed-rpc.
Could actually use all that was already made and just swap it.
Allows to remove a dependency.

Closes #1104 
To delete once merged

### DIFF
--- a/packages/integrations/src/download-client/nzbget/nzbget-integration.ts
+++ b/packages/integrations/src/download-client/nzbget/nzbget-integration.ts
@@ -1,24 +1,21 @@
 import dayjs from "dayjs";
-import { rpcClient } from "typed-rpc";
 
 import type { DownloadClientJobsAndStatus } from "../../interfaces/downloads/download-client-data";
 import { DownloadClientIntegration } from "../../interfaces/downloads/download-client-integration";
 import type { DownloadClientItem } from "../../interfaces/downloads/download-client-items";
 import type { DownloadClientStatus } from "../../interfaces/downloads/download-client-status";
-import type { NzbGetClient } from "./nzbget-types";
+import type { NzbGetGroup, NzbGetHistory, NzbGetStatus } from "./nzbget-types";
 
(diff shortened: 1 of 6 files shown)


packages/integrations/src/download-client/nzbget/nzbget-integration.ts
```diff
@@ -64,39 +61,54 @@ export class NzbGetIntegration extends DownloadClientIntegration {
...
+    if (progress === 1) {
+      await this.nzbGetApiCallAsync("editqueue", "GroupFinalDelete", "", [Number(id)]);
     } else {
-      await client.editqueue("HistoryFinalDelete", "", [Number(id)]);
+      await this.nzbGetApiCallAsync("editqueue", "HistoryFinalDelete", "", [Number(id)]);
     }
   }
 
-  private getClient() {
+  private async nzbGetApiCallAsync(method: string, ...params: unknown[]): Promise<unknown> {
```
Meierschlumpf, line 93: I would suggest to make this generic maybe?

packages/integrations/src/download-client/nzbget/nzbget-integration.ts
```diff
@@ -64,39 +61,54 @@ export class NzbGetIntegration extends DownloadClientIntegration {
...
     const url = new URL(this.integration.url);
     url.pathname += `${this.getSecretValue("username")}:${this.getSecretValue("password")}`;
```
Meierschlumpf, line 95: Is there a better way to pass username / password?

Meierschlumpf (changes requested):

packages/integrations/src/download-client/nzbget/nzbget-integration.ts
SeDemal: Unfortunately no...

Meierschlumpf (approved):

PR #1134, opened 2024-09-16
fix: indexer manager UI behaviour + class naming

SeDemal:
Make indexer manager scaling behave like the other widget.
Fixed improper base flex behaviour.
Added class naming to all components.

### DIFF
--- a/packages/widgets/src/indexer-manager/component.tsx
+++ b/packages/widgets/src/indexer-manager/component.tsx
@@ -16,9 +16,6 @@ export default function IndexerManagerWidget({
   integrationIds,
   serverData,
 }: WidgetComponentProps<"indexerManager">) {
-  if (integrationIds.length === 0) {
-    throw new NoIntegrationSelectedError();
-  }
   const t = useI18n();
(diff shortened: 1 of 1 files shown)


packages/widgets/src/indexer-manager/component.tsx
```diff
@@ -39,26 +36,49 @@ export default function IndexerManagerWidget({
     },
   );
 
+  const iconStyle = { height: "7.5cqmin", width: "7.5cqmin" };
+
+  if (integrationIds.length === 0) {
+    throw new NoIntegrationSelectedError();
+  }
```
manuel-rw, line 43: Move to top of function

manuel-rw (approved):

packages/widgets/src/indexer-manager/component.tsx
SeDemal: Why? Doesn't this cause hooks issues? I remember needing to do that for the media request widgets with Meier.

manuel-rw: Ah, sorry. Didn't see the hook. Approved.

PR #1168, opened 2024-09-23
fix: issue with color scheme in layout

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

This fixes the bug that sometimes the whole screen is black and there are no colors. The reason was, that we only used the color scheme when signed in and otherwise it was not set. Now I use a cookie and when you log in it's set to that one saved for the specific user


manuel-rw (dismissed):
Thanks. Can you fix the builds before merging?

manuel-rw (approved):

PR #1172, opened 2024-09-24
fix: various issues with docker image workflow

MaienM:
The conditional statements were all in the form of `input || true`, presumably in an effort to treat this as `true` if the input was not set as it would be when the trigger is a push rather than a manual dispatch (`null || true` => `true`). However `false || true` also results in `true`, so this would just always result in `true`.

Changing this to `!= false` does result in the desired behaviour:

- Enabled: `true != false` => `true`
- Disabled: `false != false` => `false`
- Unset: `null != false` => `true`

---

It was always looking at the `dev` branch for the commit messages in the versioning step, which meant that running it against a different branch would not have the desired result.

### DIFF
--- a/.github/workflows/deployment-docker-image.yml
+++ b/.github/workflows/deployment-docker-image.yml
@@ -38,7 +38,7 @@ jobs:
         node-version: [20]
     steps:
       - name: Discord notification
-        if: ${{ github.events.inputs.send-notifications || true }}
+        if: ${{ github.events.inputs.send-notifications != false }}
         env:
           DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK }}
         uses: Ilshidur/action-discord@master
@@ -52,7 +52,7 @@ jobs:
           token: ${{ github.token }}
           branch: dev
       - name: Discord notification
-        if: ${{ github.events.inputs.send-notifications || true }}
+        if: ${{ github.events.inputs.send-notifications != false }}
         env:
           DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK }}
         uses: Ilshidur/action-discord@master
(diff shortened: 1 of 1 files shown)


.github/workflows/deployment-docker-image.yml
```diff
@@ -50,9 +50,9 @@ jobs:
         uses: ietf-tools/semver-action@v1
         with:
           token: ${{ github.token }}
-          branch: dev
+          branch: ${{ github.ref_name }}
```
manuel-rw, line 53: We always want to use ``dev`` here

manuel-rw (approved):

.github/workflows/deployment-docker-image.yml
MaienM: It's going to get the list of commits between the latest tag and whatever `branch` is set to and using this to determine the version it should use for the build. In my case the latest tag was pointing to the latest commit on `dev`, while the branch I was running the action for had some more commits. This resulted in [the build failing](https://github.com/MaienM/homarr/actions/runs/10985329633/job/30497035910) with the message `Couldn't find any commits between branch HEAD and latest tag.`.

Is this expected? I would expect it to consider the state of the branch selected when triggering the workflow, not `dev`, but maybe I'm misunderstanding how this workflow is intended to be used.

manuel-rw: I expect this error to happen. We should cancel the workflow run when there is no diff. Do you disagree?
``dev`` is our main development branch. We will never push to master or ``dev`` directly. Hence we should compare on the ``dev`` branch and this workflow should not be executed on different branches.

MaienM: In my case there was a diff though, but it wasn't seeing this diff because it was looking at `dev` instead of the branch it was actually building.

If the intent is that this workflow can only ever work on `dev` this change isn't needed though. It might be worth considering explicitly checking that in a step (as `workflow_dispatch` allows you to select what ref to run for), but that's outside of the scope of this PR.

manuel-rw (approved):

.github/workflows/deployment-docker-image.yml
manuel-rw: I agree, we should either remove the dispatch or allow runs from other refs. But this PR can be merged as is -all other actions are out of scope. @homarr-labs/maintainers what do you think?

PR #1173, opened 2024-09-24
fix: container image always being on AMD64 arch

### LINKED ISSUE #1125: bug: Not working on ARM64

I'm trying to use the docker image, but it's failing so early in the process that I cannot open a shell in this and run the requested command inside the container. I did run it on the machine that I'm attempting to run the container on, but keep in mind that a lot of this information is useless/misleading; the listed binaries are obviously not actually the ones that will be used in the container. 

```
  System:
    OS: Linux 6.1 NixOS 24.11 (Vicuna) 24.11 (Vicuna)
    CPU: (4) arm64 Cortex-A76
    Memory: 4.73 GB / 7.84 GB
    Container: Yes
    Shell: 5.2.26 - /nix/store/w7fy8y8a296hi61nibl62xgnfwd2nfw5-bash-interactive-5.2p26/bin/bash
  Binaries:
    Node: 22.4.1 - /nix/store/2c0qimdjslxal42s4m10zj95f9r11l4l-nodejs-22.4.1/bin/node
    npm: 10.8.1 - /nix/store/2c0qimdjslxal42s4m10zj95f9r11l4l-nodejs-22.4.1/bin/npm
```

### Describe the bug

The `arm64` image doesn't actually appear to be valid `arm64`, and running it on a machine with that architecture results in `exec /usr/local/bin/docker-entrypoint.sh: exec format error`.

### Link to reproduction

N/A

### To reproduce

Attempt to start the image on a machine with an `arm64` architecture.

### Additional information

I suspect that this is due to the dockerfile specifically using an `amd64` image regardless of what platform it's being built for:

https://github.com/homarr-labs/homarr/blob/7059a3d88f73ddceb70d037bb0f362fe50bda225/Dockerfile#L1

This is probably fine for the first two steps (`builder` and `installer`), but this same image is also used for the `runner` part which I think results in the final image being based on an `amd64` image even for the `arm64` image:

https://github.com/homarr-labs/homarr/blob/7059a3d88f73ddceb70d037bb0f362fe50bda225/Dockerfile#L61

MaienM:
The container image always used an amd64 image as base, both for the build steps as well as for the runtime. For the runner this resulted in the containers just not booting with the error `exec /usr/local/bin/docker-entrypoint.sh: exec format error`.

I first tried fixing this by adjusting this only for the runner, but this resulted in a segfault later on, presumably due to some binary component in the build/dependencies that was now build for the wrong architecture, so I just ended up switching the entire thing over to use the target architecture.

Fixes #1125.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 node:20.17.0-alpine AS base
+FROM node:20.17.0-alpine AS base
 
 FROM base AS builder
 RUN apk add --no-cache libc6-compat


manuel-rw (changes requested):
Thanks for the PR. We tried this hacky "solution" because building an image for arm64 will slow down the build until GitHub actions terminates the job. This is caused by poor emulation performance with quemu and docker builds for arm64. We haven't had time to fix this issue yet. Removing this parameter will fix nothing.
GitHub will soon offer arm64 runners: https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/
We could either wait for these new runners to become available or we could attempt to build differently (perhaps using buildx?)

MaienM:
While it definitely takes a long time to build it doesn't take so long that it times out, at least my [run just now](https://github.com/MaienM/homarr/actions/runs/11037629666/job/30658954748) didn't. I'm not sure what the timeout is though, perhaps this pushes us close enough to it that it sometimes works and sometimes times out?

If that is the case we should just remove `arm64` from the list of platforms in the workflow file though, as this attempted workaround results in images that aren't functional, and it's more useful to just not have an arm64 image than it is to have a broken arm64 image.

manuel-rw:
> While it definitely takes a long time to build it doesn't take so long that it times out, at least my [run just now](https://github.com/MaienM/homarr/actions/runs/11037629666/job/30658954748) didn't. I'm not sure what the timeout is though, perhaps this pushes us close enough to it that it sometimes works and sometimes times out?
> 
> If that is the case we should just remove `arm64` from the list of platforms in the workflow file though, as this attempted workaround results in images that aren't functional, and it's more useful to just not have an arm64 image than it is to have a broken arm64 image.

Maybe they have improved the performance 🤔 . When we tried this, it took 6 hours (default timeout for GitHub actions). I will approve and then we'll just see how it goes. 24m is nowhere near acceptable but better than a broken image. Hopefully this will get better and decrease to < 10 minutes.

manuel-rw (approved):

Meierschlumpf:
@manuel-rw the arm64 one did always take about 20 - 30 minutes. It only was the armv7 that timed out because it had an infinite loop during `pnpm install`

PR #1187, opened 2024-09-26
feat: add healthchecks

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


apps/nextjs/src/app/api/health/live/route.ts
```diff
@@ -0,0 +1,93 @@
...
+      await db.query.serverSettings.findFirst();
+      const after = performance.now();
+      return {
+        latency: after - before,
+      };
+    }),
+    executeHealthCheckSafelyAsync("redis", async () => {
+      const before = performance.now();
+      await handshakeAsync();
+      const after = performance.now();
```
Meierschlumpf, line 40: You could actually simplify this by only running the handshakeAsync in here and put the before and after with latency calculation into the `executeHealthCheckSafelyAsync` method

apps/nextjs/src/app/api/health/live/route.ts
```diff
@@ -0,0 +1,93 @@
...
+      return {
+        latency: after - before,
+      };
+    }),
+  ];
+
+  const healthCheckResults = await Promise.all(healthChecks);
```
Meierschlumpf, line 47: Use Promise.allSettled instead, otherwise it will not respond anything when unhealthy

Meierschlumpf (changes requested):

apps/nextjs/src/app/api/health/live/route.ts
manuel-rw: ``Promise.allSettled`` returns a list of ``PromiseSettledResult<HealthCheckResult>[]``. That's quite inconvinient because I always have to filter for the resolved promises and TypeScript isn't smart enough that iterations will have the type of the resolved promise. Also, since I catch any errors in the safe method, which returns the promise, it should never return a rejected promise, no?
Can you make a suggestion otherwise with ``allSettled`` and suggestions for the above mentioned problems?

Meierschlumpf: Okay haven't seen that you have a try catch in there, then it's fine

apps/nextjs/src/app/api/health/live/route.ts
```diff
@@ -0,0 +1,91 @@
...
+}> => {
+  const healthChecks = [
+    executeHealthCheckSafelyAsync("database", async () => {
+      // sqlite driver does not support raw query execution. this is for a heartbeat check only - it doesn't matter if data is returned or not
+      await db.query.serverSettings.findFirst();
+      return {};
+    }),
+    executeHealthCheckSafelyAsync("redis", async () => {
+      await handshakeAsync();
+      return {};
```
Meierschlumpf, line 35: Is there a reason it returns an empty object?

apps/nextjs/src/app/api/health/live/route.ts
```diff
@@ -0,0 +1,91 @@
...
+): Promise<HealthCheckResult> => {
+  try {
+    const currentTimeBeforeCallback = performance.now();
+    const values = await callback();
+    return {
+      name,
+      status: "healthy",
+      values: {
+        ...values,
+        latency: currentTimeBeforeCallback,
```
Meierschlumpf, line 71: I guess you should do a calculation here instead

Meierschlumpf (changes requested):

apps/nextjs/src/app/api/health/live/route.ts
manuel-rw: Yes, because this method could return some data - but currently it does not. We could return additional data that is important for the healtcheck. I think we can leave it empty for now

Meierschlumpf (dismissed):

PR #1246, opened 2024-10-05
fix: rtl common translation unnecessary

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

**Thank you for your contribution. Please ensure that your pull request meets the following pull request:**

- [x] Builds without warnings or errors (``pnpm buid``, autofix with ``pnpm format:fix``)
- [x] Pull request targets ``dev`` branch
- [x] Commits follow the [conventional commits guideline](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] No shorthand variable names are used (eg. ``x``, ``y``, ``i`` or any abbrevation)

image

PR #1247, opened 2024-10-05
feat: RTL option to RSS widget

hillaliy:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


Meierschlumpf:
I think this should work already when merging #1246

hillaliy:
Before:
Screenshot 2024-10-19 at 7 48 12

After:
Screenshot 2024-10-19 at 7 49 31

packages/old-import/src/widgets/definitions/rss.ts
```diff
@@ -4,6 +4,7 @@ export type OldmarrRssDefinition = CommonOldmarrWidgetDefinition<
   "rss",
   {
     rssFeedUrl: string[];
+    anableRtl: boolean;
```
Meierschlumpf, line 7: Typo, would be `enableRtl`

Okay I first thought why we would need an option for that as we already could use `useDirection`, but it makes sense as the source of the RSS feed has obviously not to be in the same language as the user choose.

Meierschlumpf (changes requested):

Meierschlumpf (approved):

PR #1248, opened 2024-10-05
feat: add week start day to calendar widget

hillaliy:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

### DIFF
--- a/packages/translation/src/lang/en.ts
+++ b/packages/translation/src/lang/en.ts
@@ -1009,6 +1009,9 @@ export default {
       name: "Calendar",
       description: "Display events from your integrations in a calendar view within a certain relative time period",
       option: {
+        weekStartOnSanday: {
+          label: "Week start on sanday",
+        },
         filterPastMonths: {
           label: "Start from",
         },
--- a/packages/translation/src/lang/en.ts
+++ b/packages/translation/src/lang/en.ts
@@ -1010,7 +1010,7 @@ export default {
       description: "Display events from your integrations in a calendar view within a certain relative time period",
       option: {
         weekStartOnSanday: {
-          label: "Week start on sanday",
+          label: "Week start on sunday",
         },
         filterPastMonths: {
           label: "Start from",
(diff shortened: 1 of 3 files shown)


packages/translation/src/lang/en.ts
```diff
@@ -1009,6 +1009,9 @@ export default {
       name: "Calendar",
       description: "Display events from your integrations in a calendar view within a certain relative time period",
       option: {
+        weekStartOnSanday: {
+          label: "Week start on sanday",
```
SeDemal, line 1013: typo: "sunday"

SeDemal (changes requested):

Meierschlumpf:
Bro :joy: I worked at exactly the same thing, but rather than putting it into the calendar I put it in user settings :joy:
What's your opinion on that, should this be user specific or defined in the widget itself?

SeDemal:
> Bro 😂 I worked at exactly the same thing, but rather than putting it into the calendar I put it in user settings 😂 What's your opinion on that, should this be user specific or defined in the widget itself?
> 
> Mine would be #1249

Omg I thought I would have thought of that when I made the widget, so that's why it wasn't in already.
I think, either you do it for both the clock and the calendar, or we leave it both in options

SeDemal:
> > Omg I thought I would have thought of that when I made the widget, so that's why it wasn't in already. I think, either you do it for both the clock and the calendar, or we leave it both in options
> 
> What do you mean in the clock? Is there something related to first day of week?


time representation with 24h format or am/pm, date order between month and day.
All of them are localizations proper to the user in the end.

Meierschlumpf:
Maybe we can create a seperate issue for that? I think it's okay as we have it right now for now

SeDemal:
Yeah we can make a different issue/PR for that, but to me it's the same, a representation of time that is proper to a user's preference.
Overall though, making it a user setting and not a board setting seems like a good idea.
@manuel-rw thoughts?

hillaliy:
Ok my friend, on settings it will be better. I'll close this PR

PR #1265, opened 2024-10-07
refactor: replace serverdata with suspense query

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


packages/widgets/src/rssFeed/component.tsx
```diff
@@ -3,15 +3,29 @@ import { Card, Flex, Group, Image, ScrollArea, Stack, Text } from "@mantine/core
...
 
-export default function RssFeed({ serverData, options }: WidgetComponentProps<"rssFeed">) {
-  if (serverData?.initialData === undefined) {
-    return null;
-  }
+export default function RssFeed({ options, itemId }: WidgetComponentProps<"rssFeed">) {
+  const [rssFeeds] = clientApi.widget.rssFeed.getFeeds.useSuspenseQuery(
+    {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      itemId: itemId!,
```
manuel-rw, line 15: Can we always expect this to be not null?

manuel-rw (approved):

manuel-rw:
Great changes. How's performance?
Code is much cleaner now. Love it ❤️

packages/widgets/src/rssFeed/component.tsx
Meierschlumpf: Always except when you are on the try out page. IMO it would make generally more sense to just send the options to the server

Meierschlumpf:
> Great changes. How's performance? Code is much cleaner now. Love it ❤️

Performance is a little faster, but it's mostly faster in that sense that not everything has to wait for the slowest response and rather is shown as soon as it's resolved

manuel-rw (approved):

PR #1322, opened 2024-10-18
feat: add everyone group

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


apps/nextjs/src/app/[locale]/manage/users/groups/[id]/_reserved-group-alert.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const t = await getI18n();
+
+  return (
+    <Alert variant="light" color="yellow" icon={<IconExclamationCircle size="1rem" stroke={1.5} />}>
+      {t("group.reservedNotice.message", {
+        checkoutDocs: (
+          <Anchor
+            size="sm"
+            component={Link}
+            href="https://deploy-preview-113--homarr-docs.netlify.app/docs/management/users/#special-groups"
```
Comment (manuel-rw) [inline apps/nextjs/src/app/[locale]/manage/users/groups/[id]/_reserved-group-alert.tsx:17]: **NEVER** do this. It will **always** be forgotten.
Introduce a constant and use it as a base for all links. Additionally, I'm unsure whether we should use the correct URL here or not. We'll probably forget to set this to ``https://homarr.dev`` when we publish 1.0

packages/api/src/router/group.ts
```diff
@@ -236,6 +237,19 @@ const checkSimilarNameAndThrowAsync = async (db: Database, name: string, ignoreI
   }
 };
 
+const throwIfGroupNameIsReservedAsync = async (db: Database, id: string) => {
+  const group = await db.query.groups.findFirst({
```
manuel-rw, line 241: Can't you filter server side here using an ANY? As an example, return the count instead of the group to reduce bandwidth

manuel-rw (approved):

PR #1394, opened 2024-10-31
feat: add support for all languages from old homarr

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

All translations from non english were migrated from old homarr with a script. It may contain incorrect translations. Overall 300 / 1050 translations were copied.

Created issue #1402 for crowdin integration and made a prototype with a project on my own to verify it would work with the current structure


ajnart (commented):
Great work ! How did you manage to port everything ? It's impressive

- Do you plan to add back the live translation feature (Crowdin) ? I made it so that the script is added to the page depending on the user preference 
- I saw you're switching to next-intl (the translation thing from crowdin), they support a localstorage adapter if I recall correctly so you could make it so that translations are only passed once, also could have translations come from a CDN to synchronize them with the app (in a chron job?) with Crowdin's OTA translations feature. 
- Last comment the i18n ally vscode extension works really well with this plugin so maybe add a config into the `.vscode` folder that takes care of this, then you can have auto-add keys/inline keys replacement (it will show the translations) and also a warning when a text is not translated

Meierschlumpf:
> Great work ! It's impressive
> 
> * Did you make sure that this format is compatible with crowdin ? It might require to have a `.json` file and import it in the `.ts` as a valid directly (no need for validation)
> * Do you plan to add back the live translation feature (Crowdin) ? I made it so that the script is added to the page depending on the user preference
> * I saw you're switching to next-intl (the translation thing from crowdin), they support a localstorage adapter if I recall correctly so you could make it so that translations are only passed once then saved in localstorage, also could utilize translations come from Crowdin's CDN to synchronize them with the app (in a chron job?) with Crowdin's OTA translations feature. (I believe they gave us a Pro plan for open-source so we have pretty much all features for free)
> * Last comment the i18n ally vscode extension works really well with this plugin so maybe add a config into the `.vscode` folder that takes care of this, then you can have auto-add keys/inline keys replacement (it will show the translations) and also a warning when a text is not translated

1. Yeah TS is not working great, it wants to translate the import statements 😂 
image

3. Yes it's planned to add this again
4. Okay nice, I think CDN is a little bit overkill for now, especially as we will release automatically every 7 days, but the locale storage thing could be quite nice
5. I think you already added it, not sure if it's working when I changed it from next-international to next-intl tough.

Meierschlumpf:
Test issue: https://github.com/vitest-dev/vitest/issues/5101
Fixing it right now by adding default coverage excludes to `vitest.config.ts`

manuel-rw:
> Yeah TS is not working great, it wants to translate the import statements 😂

So... what do we do about this then?

> Okay nice, I think CDN is a little bit overkill for now, especially as we will release automatically every 7 days, but the locale storage thing could be quite nice

I agree, let's plan this for >1.0.

Meierschlumpf:
> So... what do we do about this then?

I already changed it to JSON

manuel-rw:
> > So... what do we do about this then?
> 
> I already changed it to JSON

Stupid question. Would we have an advantage storing as JSON but converting to TS on the fly / on commit?
If not, I'm okay with this solution.

manuel-rw (approved):

Issue #1414, opened 2024-11-03
bug: invalid_type zod sabnzbd


manuel-rw:
Investigate:
```
@homarr/tasks:dev:   {
@homarr/tasks:dev:     "code": "invalid_type",
@homarr/tasks:dev:     "expected": "number",
@homarr/tasks:dev:     "received": "boolean",
@homarr/tasks:dev:     "path": [
@homarr/tasks:dev:       "history",
@homarr/tasks:dev:       "slots",
@homarr/tasks:dev:       0,
@homarr/tasks:dev:       "retry"
@homarr/tasks:dev:     ],
@homarr/tasks:dev:     "message": "Expected number, received boolean"
@homarr/tasks:dev:   }
@homarr/tasks:dev: ]"
```

SeDemal:
will do

PR #1415, opened 2024-11-03
fix: remove unused properties in SabNZBd schema

SeDemal:
fixes: #1414


SeDemal:
@manuel-rw if you test it real quick, knowing them there might be something else needing to be changed, and if so I need to investigate more

packages/integrations/src/download-client/sabnzbd/sabnzbd-schema.ts
```diff
@@ -87,7 +87,7 @@ export const historySchema = z.object({
         size: z.string(),
         category: z.string(),
         pp: z.string(),
-        retry: z.number(),
+        retry: z.number().or(z.boolean()),
```
manuel-rw, line 90: Why not use transform to transform from Number to boolean?

manuel-rw (dismissed):

packages/integrations/src/download-client/sabnzbd/sabnzbd-schema.ts
SeDemal: Because we don't use the value. It's only there because I made the full schema.
I could but I feel that the only important thing here is for it to pass the schema validation.

Meierschlumpf: I think we can remove it then if we don't use it

manuel-rw: You should remove all values that we don't use. We don't need them. They just eat up network bandwidth and RAM

SeDemal: IIRC we were talking multiple times about publishing our API controllers, I must have done it like so with this in mind.
The bandwidth usage is going to be the same no matter what but it is true that we could save up a little on processing power in this case.
If you both think I should remove it then I'll do so in this PR

manuel-rw: > I think we can remove it then if we don't use it
I agree. Remove it 👍

Meierschlumpf (approved):

PR #1498, opened 2024-11-18
fix: user is undefined when using use-suspense-query

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Thanks to esavitskiy I've found a pretty straight forward solution. We just pass the cookie string as header when suspense-query is executed server side. See https://github.com/apollographql/apollo-client-nextjs/issues/85#issuecomment-1952430214
Using cookies() directly does not work as it's a client component

**Before**
image
image

**After**
image


SeDemal (approved):
lgtm, gj

PR #1595, opened 2024-12-03
fix: remove unnecessary complex copy and build steps in dockerfile

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

We misused the prune functionality from turborepo which allows to only include the necessary things for installation and build. But as we need everything in the same image, we can simply skip this mess and build normally with all files present and then copy the build files to the run step.
This also resolves the issue with docker build in #1586 which occured because the patched dependency was no longer part of the package.json after prune.

I tested the image in this state locally, size is about 330 MB and it works as expected

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,45 +6,14 @@ RUN apk update
 
 # Set working directory
 WORKDIR /app
-COPY . .
-RUN npm i -g turbo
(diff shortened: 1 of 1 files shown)


manuel-rw (approved):
😎

PR #1626, opened 2024-12-07
feat: add update indicator

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

image

### DIFF
--- a/apps/nextjs/src/components/user-avatar-menu.tsx
+++ b/apps/nextjs/src/components/user-avatar-menu.tsx
@@ -77,7 +77,7 @@ export const UserAvatarMenu = ({ children, availableUpdates }: UserAvatarMenuPro
               leftSection={<IconBellRinging size="1rem" />}
             >
               <Text fw="bold" size="sm">
-                {t("updateAvailable", { countUpdates: availableUpdates.length, tag: availableUpdates[0].tag_name })}
+                {t("updateAvailable", { countUpdates: availableUpdates.length, tag: availableUpdates[0].tagName })}
               </Text>
             </Menu.Item>
             <Menu.Divider />
--- a/apps/nextjs/src/components/user-avatar-menu.tsx
+++ b/apps/nextjs/src/components/user-avatar-menu.tsx
@@ -31,7 +31,7 @@ import { CurrentLanguageCombobox } from "./language/current-language-combobox";
 
 interface UserAvatarMenuProps {
   children: ReactNode;
-  availableUpdates: RouterOutputs["updateChecker"]["getAvailableUpdates"];
+  availableUpdates?: RouterOutputs["updateChecker"]["getAvailableUpdates"];
 }
 
 export const UserAvatarMenu = ({ children, availableUpdates }: UserAvatarMenuProps) => {
--- a/apps/nextjs/src/components/user-avatar-menu.tsx
+++ b/apps/nextjs/src/components/user-avatar-menu.tsx
@@ -7,6 +7,7 @@ import { useRouter } from "next/navigation";
 import { Center, Menu, Stack, Text, useMantineColorScheme } from "@mantine/core";
 import { useHotkeys, useTimeout } from "@mantine/hooks";
 import {
+  IconBellRinging,
   IconCheck,
   IconHome,
   IconLogin,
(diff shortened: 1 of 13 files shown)


apps/nextjs/src/components/user-avatar-menu.tsx
```diff
@@ -64,6 +68,21 @@ export const UserAvatarMenu = ({ children }: UserAvatarMenuProps) => {
     // We use keepMounted so we can add event listeners to prevent navigating away without saving the board
     <Menu width={300} withArrow withinPortal keepMounted>
       <Menu.Dropdown>
+        {availableUpdates && availableUpdates.length > 0 && availableUpdates[0] && (
```
Meierschlumpf, line 71: Maybe show the available updates only the admin? Or what do you think?

packages/api/src/router/update-checker.ts
```diff
@@ -0,0 +1,16 @@
+import { createSubPubChannel } from "../../../redis/src/lib/channel";
```
Meierschlumpf, line 1: Is it possible to import `@homarr/redis` here instead?

packages/cron-jobs/src/jobs/update-checker.ts
```diff
@@ -0,0 +1,58 @@
...
+import { compareSemVer, isValidSemVer } from "semver-parser";
+
+import { EVERY_HOUR } from "@homarr/cron-jobs-core/expressions";
+import { logger } from "@homarr/log";
+
+import packageJson from "../../../../package.json";
+import { createSubPubChannel } from "../../../redis/src/lib/channel";
+import { createCronJob } from "../lib";
+
+export const updateCheckerJob = createCronJob("updateChecker", EVERY_HOUR, {
```
Meierschlumpf, line 11: What about using a `request handler` instead?

packages/translation/src/lang/en.json
```diff
@@ -733,7 +733,8 @@
         "logout": "Logout",
         "login": "Login",
         "homeBoard": "Your home board",
-        "loggedOut": "Logged out"
+        "loggedOut": "Logged out",
+        "updateAvailable": "({countUpdates}) updates available: {tag}"
```
Meierschlumpf, line 737: IMO move the brackets `5 updates available (v1.2.3)` makes more sense

Meierschlumpf (changes requested):

packages/cron-jobs/src/jobs/update-checker.ts
manuel-rw: All other cron jobs for non-integrations do not use it. Why should we here? Isn't it exclusively for the integrations?

Meierschlumpf: Your correct, right now we only use it with the cached-integration-request-handler, but I build it that way that all requests can be handled centralized and cached. For example my plan is to move the ping results to this system as well. You can simply use the createCachedRequestHandler method provided

apps/nextjs/src/components/user-avatar-menu.tsx
```diff
@@ -60,10 +64,27 @@ export const UserAvatarMenu = ({ children }: UserAvatarMenuProps) => {
     });
   }, [logoutUrl, openModal, router]);
 
+  const isAdmin = session.data?.user.permissions.includes("admin");
+
   return (
     // We use keepMounted so we can add event listeners to prevent navigating away without saving the board
     <Menu width={300} withArrow withinPortal keepMounted>
       <Menu.Dropdown>
+        {isAdmin && availableUpdates && availableUpdates.length > 0 && availableUpdates[0] && (
```
Meierschlumpf, line 73: If we only show them to the admin we should also only show the indicator then. So it's probably easier if we only fetch it in the user.tsx if the user is admin and then pass RouterOutputs[*] | undefined down. Then the logic for available updates does not need to change

packages/cron-jobs/src/jobs/update-checker.ts
```diff
@@ -0,0 +1,58 @@
+import { Octokit } from "octokit";
+import { compareSemVer, isValidSemVer } from "semver-parser";
+
+import { EVERY_HOUR } from "@homarr/cron-jobs-core/expressions";
+import { logger } from "@homarr/log";
+
+import packageJson from "../../../../package.json";
+import { createSubPubChannel } from "../../../redis/src/lib/channel";
```
Meierschlumpf, line 8: Wrong import for channel

Meierschlumpf (changes requested):

Meierschlumpf (approved):

PR #1637, opened 2024-12-10
refactor: Move integration search flag to categories

SeDemal:
While adding the encodeURI function to the query (Spaces were not working) I noticed the implementation was not using the category system, after discussing it with Mani it was asserted that it was only for personal preference/meta reasons.
I am suggesting this PR as a change to keep the system centralized using the generalized category system instead.
(This also implements the encodeURI fix, which I can make a new PR for if this one doesn't get accepted).

### DIFF
--- a/packages/api/src/router/integration/integration-router.ts
+++ b/packages/api/src/router/integration/integration-router.ts
@@ -14,15 +14,16 @@ import {
 } from "@homarr/db/schema/sqlite";
 import type { IntegrationSecretKind } from "@homarr/definitions";
 import {
+  getIntegrationKindsByCategory,
   getPermissionsWithParents,
   integrationDefs,
   integrationKinds,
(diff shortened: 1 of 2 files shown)


manuel-rw:
As discussed via Discord with Sen, I vote to keep the current system. For me, it makes more sense on a meta-level

SeDemal:
For me on the meta level, integration searching being a function of integrations, it makes more sense to use it exactly the same way we would any other integrations.

Meierschlumpf (dismissed):

manuel-rw:
Seems that ajnart isn't going to respond soon. I took some time to think about this change again and I think I agree with it. I will take over, resolve conflicts and merge, because the majority now voted for this change.

Meierschlumpf (approved):

PR #1657, opened 2024-12-14
feat: add tdarr integration

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

image


packages/api/src/router/widgets/media-transcoding.ts
```diff
@@ -0,0 +1,28 @@
...
+    .unstable_concat(createIndexerManagerIntegrationMiddleware("query"))
+    .query(async ({ ctx, input }) => {
+      const innerHandler = mediaTranscodingRequestHandler.handler(ctx.integration, {
+        pageOffset: input.pageOffset,
+        pageSize: input.pageSize,
+      });
+      const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
+
+      return {
+        integrationId: ctx.integration.id,
```
SeDemal, line 24: We should return all data from integrations (except secrets), so it's easier to handle in the front end.
Mostly used so we can show what data belongs to what integration when there are multiple.

packages/api/src/router/widgets/media-transcoding.ts
SeDemal, line 1: Did you think about adding subscription? It's useful for the list of current tasks aka "queue".

packages/widgets/src/media-transcoding/panels/statistics.panel.tsx
```diff
@@ -0,0 +1,139 @@
...
+          <Text size="xs">{t("healthChecks")}</Text>
+        </Stack>
+      </Group>
+      <Group
+        style={{
+          flex: 1,
+        }}
+        justify="apart"
+        align="center"
+        wrap="nowrap"
```
SeDemal, line 88: Seeing the screenshot I think it might just be missing the w="100%" here?
I know it's not really worth it but I feel like it's starring at me.

SeDemal (dismissed):

packages/api/src/router/widgets/media-transcoding.ts
manuel-rw: At this time, we don't need to return all data - it's not needed. As soon as we do need it, we can return it

packages/api/src/router/widgets/media-transcoding.ts
manuel-rw: I will create a follow up issue

manuel-rw: See #1659

packages/widgets/src/media-transcoding/panels/queue.panel.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <ScrollArea style={{ flex: "1" }}>
+      <Table style={{ tableLayout: "fixed" }}>
+        <thead>
+          <tr>
+            <th>{t("table.file")}</th>
+            <th style={{ width: 80 }}>{t("table.size")}</th>
+          </tr>
+        </thead>
+        <tbody>
```
Meierschlumpf, line 34: Is there a reason you use the default html elements and not `Table.THead` and such? Then you could also use `w={80}`

packages/widgets/src/media-transcoding/panels/workers.panel.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <Center style={{ flex: "1" }}>
+        <Title order={3}>{t("empty")}</Title>
+      </Center>
+    );
+  }
+
+  return (
+    <ScrollArea style={{ flex: "1" }}>
+      <Table style={{ tableLayout: "fixed" }}>
+        <thead>
```
Meierschlumpf, line 25: Same for table arguments, please use those of mantine

Meierschlumpf (changes requested):

Meierschlumpf (dismissed):

Meierschlumpf (approved):

PR #1759, opened 2024-12-23
feat(docker): add support for pgid and puid env variables

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


IIIdefconIII:
following this thread

scripts/entrypoint.sh
```diff
@@ -7,4 +7,9 @@ mkdir -p /appdata/redis
 
 chown -R nextjs:nodejs /appdata
 
+DOCKER_SOCKET=/var/run/docker.sock
+if [ -f "$DOCKER_SOCKET" ]; then
+    chown -R nextjs:nodejs $DOCKER_SOCKET
+fi
+
```
manuel-rw, line 14: Add a comment, why this chown is required

manuel-rw (changes requested):

scripts/entrypoint.sh
```diff
@@ -7,4 +7,9 @@ mkdir -p /appdata/redis
 
 chown -R nextjs:nodejs /appdata
 
+DOCKER_SOCKET=/var/run/docker.sock
```
Meierschlumpf, line 10:
```suggestion
# Change owner to allow homarr to read docker socket when mounted
DOCKER_SOCKET=/var/run/docker.sock
```

darkcloud784:
Won't chowning the socket also impact other things using said socket? IE portainer. If so, might want to think about setting a user field that someone can run the container as that will have access to said socket.

IIIdefconIII:
> Won't chowning the socket also impact other things using said socket? IE portainer. If so, might want to think about setting a user field that someone can run the container as that will have access to said socket.

mma also think so we need to get PUID and GUID ad environment variables

Meierschlumpf:
> Won't chowning the socket also impact other things using said socket? IE portainer. If so, might want to think about setting a user field that someone can run the container as that will have access to said socket.

You are right, I didn't know that. I guess then we'll need to add `PUID` and `PGID`
Reference for me: https://docs.linuxserver.io/general/understanding-puid-and-pgid/#why-use-these

Meierschlumpf:
What do you think of that solution? We allow the user to change PUID and PGID to whatever they need and chown the volumes the same way and internally still use the `internal_user` to run the processes as we don't want to change the owner of all files in the `/app` directory when starting up the container. But that user is member of the `homarr` group which is owner of the volumes

scripts/entrypoint.sh
```diff
@@ -1,10 +1,30 @@
...
 mkdir -p /appdata/redis
 
-chown -R nextjs:nodejs /appdata
+chown -R $PUID:$PGID /appdata
+# Allow homarr group to write to appdata folder
+chmod -R g+w /appdata
+
+# Change owner to allow homarr to read docker socket when mounted
+DOCKER_SOCKET=/var/run/docker.sock
+if [ -r "$DOCKER_SOCKET" ]; then
```
manuel-rw, line 24: What happens if the socket is read-only? At least run a test

manuel-rw (changes requested):

scripts/entrypoint.sh
IIIdefconIII: image
cant wait for it to fly in working, alread just changed the compose file

Dockerfile
```diff
@@ -25,51 +25,47 @@ RUN corepack enable pnpm && pnpm build
 FROM base AS runner
 WORKDIR /app
 
-# gettext is required for envsubst
-RUN apk add --no-cache redis nginx bash gettext su-exec
+# gettext is required for envsubst, shadow for usermod, su-exec for running as non-root
+RUN apk add --no-cache redis nginx bash gettext su-exec shadow
```
Meierschlumpf, line 29: `shadow` can be removed

Dockerfile
```diff
@@ -25,51 +25,47 @@ RUN corepack enable pnpm && pnpm build
...
-RUN addgroup --system --gid 1001 nodejs
-RUN adduser --system --uid 1001 nextjs
-
 # Enable homarr cli
-COPY --from=builder --chown=nextjs:nodejs /app/packages/cli/cli.cjs /app/apps/cli/cli.cjs
+COPY --from=builder /app/packages/cli/cli.cjs /app/apps/cli/cli.cjs
 RUN echo $'#!/bin/bash\ncd /app/apps/cli && node ./cli.cjs "$@"' > /usr/bin/homarr
 RUN chmod +x /usr/bin/homarr
 
+ENV APPLICATION_PGID=7575
```
Meierschlumpf, line 40: Can be removed

Dockerfile
```diff
@@ -25,51 +25,47 @@ RUN corepack enable pnpm && pnpm build
...
+RUN addgroup --system --gid $APPLICATION_PGID homarr_app
```
Meierschlumpf, line 41: Can be removed

manuel-rw (approved):

manuel-rw (approved):

PR #1770, opened 2024-12-24
fix(board): tiles order not respected on firefox

### LINKED ISSUE #1102: bug: tiles order not respected on firefox

Development Environment
Firefox

### Describe the bug

Upon entering a board right after starting `pnpm dev`, the items won't be in the right place.
Editing the board in this state will result in all kinds of weird placement issues.

Actually self corrects directly after refreshing the page it seems.
Don't know if there's another way to get to that state that may cause problem to the standard user.

### To reproduce

pnpm dev and reload the board page while it's starting up.

### Additional information

image

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Closes #1102 

The bug happens because the order in the virtual DOM (React) and in the real dom (Modified by Gridstack) do not match. The reason for this is that Gridstack creates the items itself and places them in the correct x / y order into the gridstack section. While we did not respect the correct order during the rendering of the virtual DOM it resulted in a mixed up order which then was rendered in the wrong order because suspense only had a reference to the elements at the specific places

### DIFF
--- a/apps/nextjs/src/components/board/sections/content.tsx
+++ b/apps/nextjs/src/components/board/sections/content.tsx
@@ -19,7 +19,7 @@ export const SectionContent = () => {
         return itemA.xOffset - itemB.xOffset;
       }
 
-      return itemA.yOffset - itemB.xOffset;
+      return itemA.yOffset - itemB.yOffset;
     });
   }, [section.items, innerSections]);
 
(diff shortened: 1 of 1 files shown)


SeDemal (approved):

manuel-rw (approved):
Good job 🚀

PR #1775, opened 2024-12-24
fix(media-server): redesign widget

hillaliy:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

image
image


packages/translation/src/lang/en.json
```diff
@@ -1457,7 +1457,12 @@
     "mediaServer": {
       "name": "Current media server streams",
       "description": "Show the current streams on your media servers",
-      "option": {}
+      "option": {},
+      "items": {
+        "user": "User:",
+        "name": "Name:",
+        "id": "Id:"
```
Meierschlumpf, line 1464: Remove the `:` and move it to the component itself

packages/widgets/src/media-server/component.tsx
```diff
@@ -1,17 +1,24 @@
...
 import { clientApi } from "@homarr/api/client";
+import { integrationDefs } from "@homarr/definitions";
 import type { StreamSession } from "@homarr/integrations";
+import { useScopedI18n } from "@homarr/translation/client";
 import { useTranslatedMantineReactTable } from "@homarr/ui/hooks";
 
 import type { WidgetComponentProps } from "../definition";
 
 export default function MediaServerWidget({ integrationIds, isEditMode }: WidgetComponentProps<"mediaServer">) {
+  const [clickedIndex, setClickedIndex] = useState<number>(0);
```
Meierschlumpf, line 20:
```suggestion
  const [clickedIndex, setClickedIndex] = useState(0);
```

packages/widgets/src/media-server/component.tsx
```diff
@@ -83,49 +93,147 @@ export default function MediaServerWidget({ integrationIds, isEditMode }: Widget
...
     mantineTableContainerProps: {
       style: {
-        flexGrow: 5,
+        height: "100%",
       },
     },
+    mantineTableBodyCellProps: ({ row }) => ({
+      onClick: () => {
+        setClickedIndex(row.index);
+        open();
```
Meierschlumpf, line 153: Put those two operations in a `startTransition`. See https://react.dev/reference/react/startTransition
It will only rerender the components once and do the two operations in parallel

packages/widgets/src/media-server/component.tsx
```diff
@@ -83,49 +93,147 @@ export default function MediaServerWidget({ integrationIds, isEditMode }: Widget
...
+      },
+    }),
   });
 
   return (
-    <Box h="100%">
+    <Stack gap={0} h="100%" display="flex" style={baseStyle}>
       <MantineReactTable table={table} />
-    </Box>
+      {flatSessions[0]?.integrationKind && (
```
Meierschlumpf, line 161: Does the widget only support one integration? If now we should show multiple here

packages/widgets/src/media-server/component.tsx
```diff
@@ -83,49 +93,147 @@ export default function MediaServerWidget({ integrationIds, isEditMode }: Widget
...
+              {item.currentlyPlaying?.episodeName} (
+              {item.currentlyPlaying?.seasonName && item.currentlyPlaying.seasonName})
+            </Title>
+          </Flex>
+          <NormalizedLine itemKey={t("user")} value={item.user.username} />
+          <NormalizedLine itemKey={t("name")} value={item.sessionName} />
+          <NormalizedLine itemKey={t("id")} value={item.sessionId} />
+        </Stack>
+      )}
+    </Modal>
```
Meierschlumpf, line 228: Instead of using the general Modal component you could use the modal system we have. Then you can also directly pass in innerProps and don't need to use the index and `useDisclosure`.

See `copy-api-key-modal.tsx` for an example. You will need the `createModal` method AND a `withOptions` after calling that function. If something is not clear, just ask!

packages/widgets/src/media-server/component.tsx
```diff
@@ -83,49 +93,147 @@ export default function MediaServerWidget({ integrationIds, isEditMode }: Widget
...
+    ) : item?.currentlyPlaying?.type === "video" ? (
+      <IconVideo size={20} />
+    ) : item?.currentlyPlaying?.type === "audio" ? (
+      <IconDeviceAudioTape size={20} />
+    ) : null;
+
+  return (
+    <Modal opened={opened} onClose={onClose} centered title={currentlyPlayingType} size="auto">
+      {item === undefined ? (
+        <Center>{"No item found"}</Center>
```
Meierschlumpf, line 213: Should probably be translated and does not need to be in `{"` and `"}`

packages/widgets/src/media-server/component.tsx
```diff
@@ -83,49 +93,147 @@ export default function MediaServerWidget({ integrationIds, isEditMode }: Widget
...
+      ) : (
+        <Stack align="center">
+          <Flex direction="column" gap="xs" align="center">
+            <Title>{item.currentlyPlaying?.name}</Title>
+            <Title order={3}>
+              {item.currentlyPlaying?.episodeName} (
+              {item.currentlyPlaying?.seasonName && item.currentlyPlaying.seasonName})
```
Meierschlumpf, line 220: If we show the name and episode / season name like this. It wouldn't make much sense to be able to open the modal when nothing is playing no?

Meierschlumpf (changes requested):

packages/widgets/src/media-server/component.tsx
hillaliy: if nothing is playing there is no option to open the modal.

packages/widgets/src/media-server/component.tsx
hillaliy: I build it like downloads widget, and there is not use with createModal. Why not to use Mantine modal component?

Meierschlumpf: Under the hood it's also using that, because then you don't need to render it in the dom yourself and can just call a function after using the useModalAction hook

hillaliy: need help with this

packages/widgets/src/media-server/component.tsx
```diff
@@ -83,49 +90,141 @@ export default function MediaServerWidget({ integrationIds, isEditMode }: Widget
...
       style: {
-        flexGrow: 5,
+        height: "100%",
       },
     },
+    mantineTableBodyCellProps: ({ row }) => ({
+      onClick: () => {
+        openModal({
+          item: row.original,
+        });
```
Meierschlumpf, line 153: You probably want to add the custom title or not?

Meierschlumpf: Also what happened to the icon?

hillaliy: I want that the title will be the icon but it can't be done. do you have a better idea?

hillaliy:
Screenshot 2024-12-28 at 9 48 31

packages/widgets/src/media-server/component.tsx
```diff
@@ -150,6 +151,16 @@ export default function MediaServerWidget({ integrationIds, isEditMode }: Widget
...
+          title:
+            row.original.currentlyPlaying?.type === "movie" ? (
+              <IconMovie size={36} />
+            ) : row.original.currentlyPlaying?.type === "tv" ? (
+              <IconDeviceTv size={36} />
+            ) : row.original.currentlyPlaying?.type === "video" ? (
+              <IconVideo size={36} />
+            ) : (
+              <IconDeviceAudioTape size={36} />
+            ),
```
Meierschlumpf, line 163:
```suggestion
        },
        {
          title:
            row.original.currentlyPlaying?.type === "movie" ? (
              <IconMovie size={36} />
            ) : row.original.currentlyPlaying?.type === "tv" ? (
              <IconDeviceTv size={36} />
            ) : row.original.currentlyPlaying?.type === "video" ? (
              <IconVideo size={36} />
            ) : (
              <IconDeviceAudioTape size={36} />
            ),
        },
```

I would use the official modal title instead of passing it as innerProps

Meierschlumpf (changes requested):

packages/widgets/src/media-server/component.tsx
hillaliy: i get an error - Type 'Element' is not assignable to type 'stringOrTranslation | undefined'.

Meierschlumpf: Okay I'll fix that real quick 👍🏼

Meierschlumpf: Okay I don't do it, I like the look of your implementation more

Meierschlumpf (approved):

PR #1995, opened 2025-01-17
fix: User options switch on load value always false

### LINKED ISSUE #1998: bug: Using symbols for pings, display error
### Describe the bug

Another thing I noticed is that when I check the Use icons for pings item and press save, the setting is saved and the item is also shown as active, but as soon as I refresh the page, the item appears as disabled again, even though the setting is still active. Should be a display error!

### Steps to reproduce

The bug will be there for everyone...

### Impact

Not a big impact. Homarr can be used as usual

### Additional information

_No response_

### Version

1.0.0-beta

### Installation method

Docker Run

### Browser

Chrome

SeDemal:
Switch needs "checked" to be set instead of value.
Maybe could have changed the form to set a different field instead.

closes #1998

### DIFF
--- a/apps/nextjs/src/app/[locale]/manage/users/[userId]/general/_components/_ping-icons-enabled.tsx
+++ b/apps/nextjs/src/app/[locale]/manage/users/[userId]/general/_components/_ping-icons-enabled.tsx
@@ -51,7 +51,10 @@ export const PingIconsEnabled = ({ user }: PingIconsEnabledProps) => {
   return (
     <form onSubmit={form.onSubmit(handleSubmit)}>
       <Stack gap="md">
-        <Switch {...form.getInputProps("pingIconsEnabled", { type: "checkbox" })} label={t("user.field.pingIconsEnabled.label")} />
+        <Switch
+          {...form.getInputProps("pingIconsEnabled", { type: "checkbox" })}
+          label={t("user.field.pingIconsEnabled.label")}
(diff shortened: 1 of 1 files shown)


apps/nextjs/src/app/[locale]/manage/users/[userId]/general/_components/_ping-icons-enabled.tsx
```diff
@@ -51,7 +51,11 @@ export const PingIconsEnabled = ({ user }: PingIconsEnabledProps) => {
   return (
     <form onSubmit={form.onSubmit(handleSubmit)}>
       <Stack gap="md">
-        <Switch {...form.getInputProps("pingIconsEnabled")} label={t("user.field.pingIconsEnabled.label")} />
+        <Switch
+          checked={form.getInputProps("pingIconsEnabled").value as boolean}
+          {...form.getInputProps("pingIconsEnabled")}
+          label={t("user.field.pingIconsEnabled.label")}
+        />
```
Comment (Meierschlumpf) [inline apps/nextjs/src/app/[locale]/manage/users/[userId]/general/_components/_ping-icons-enabled.tsx:58]:
```suggestion
        <Switch {...form.getInputProps("pingIconsEnabled", { type: "checkbox" })} label={t("user.field.pingIconsEnabled.label")} />
```

Meierschlumpf (changes requested):

Meierschlumpf (approved):

PR #2099, opened 2025-01-24
feat: weather widget - add wind speed, option to disable decimals for temperature & redesign dropdown forecast

hillaliy:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


packages/old-import/src/widgets/options.ts
```diff
@@ -77,6 +77,8 @@ const optionMapping: OptionMapping = {
     forecastDayCount: (oldOptions) => oldOptions.forecastDays,
     hasForecast: (oldOptions) => oldOptions.displayWeekly,
     isFormatFahrenheit: (oldOptions) => oldOptions.displayInFahrenheit,
+    disableTemperatureDecimals: (oldOptions) => oldOptions.disableTemperatureDecimals,
```
manuel-rw, line 80: Wait, we already had this in 0.15.10 and before? I didn't know that 😅

packages/widgets/src/weather/component.tsx
manuel-rw, line 1: Please post a screenshot

packages/old-import/src/widgets/options.ts
Meierschlumpf: No we did not, they should not be added to the `old-import` schema above (see packages/old-import/src/widgets/definitions/weather.ts)

manuel-rw: Ok, so please don't add it @hillaliy

packages/widgets/src/weather/component.tsx
hillaliy: Screenshot 2025-01-27 at 17 10 27

hillaliy:
The typecheck error because old homarr options

Meierschlumpf:
> The typecheck error because old homarr options

For options that were added you can add the callback as it requests but just return undefined (then the default value is used)
Like: `optionName: () => undefined`

PR #2133, opened 2025-01-26
fix: unable to update integration when encryption key changed

### LINKED ISSUE #2084: bug: unable to update secrets
### Describe the bug

I changed my Secret Encryption Key before upgrading, so homarr couldn't access my secrets. 

I tried to update the secrets (integration API keys) and got an error. 

I was told to make an issue so homarr doesn't decrypt when secrets are updated

From support: "we allow partial upgrades, I guess here we should add a try catch then and not reuse the secret when an error occurs"

### Steps to reproduce

1. change Secret Encryption Key in unraid template before upgrade 
2. after upgrade integrations not working 
3. because the secrets are encrypted 
4. try to change the secrets (integration API keys)
5. get the error 

### Impact

Unable to use integrations unless I delete and start over 

### Additional information

_No response_

### Version

1.1.0

### Installation method

Unraid Apps

### Browser

Firefox

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Closes #2084

### DIFF
--- a/packages/api/src/router/integration/integration-test-connection.ts
+++ b/packages/api/src/router/integration/integration-test-connection.ts
@@ -3,6 +3,7 @@ import type { Integration } from "@homarr/db/schema";
 import type { IntegrationKind, IntegrationSecretKind } from "@homarr/definitions";
 import { getAllSecretKindOptions } from "@homarr/definitions";
 import { integrationCreator, IntegrationTestConnectionError } from "@homarr/integrations";
+import { logger } from "@homarr/log";
 
 type FormIntegration = Integration & {
   secrets: {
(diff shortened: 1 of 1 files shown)


packages/api/src/router/integration/integration-test-connection.ts
```diff
@@ -28,11 +29,22 @@ export const testConnectionAsync = async (
...
+  const decryptedDbSecrets = dbSecrets
+    .map((secret) => {
+      try {
+        return {
+          ...secret,
+          value: decryptSecret(secret.value),
+          source: "db" as const,
+        };
+      } catch {
+        logger.warn(
```
manuel-rw, line 41: Do not swallow error

manuel-rw (changes requested):

packages/api/src/router/integration/integration-test-connection.ts
Meierschlumpf: What do you mean with this? What would you do and why?

manuel-rw: I mean that, in JavaScript you can't specify the exception to catch. Hence, any error thrown will trigger the catch block. It would be useful to log the error for troubleshooting. Swallowing is bad because we will never know what happened here

Meierschlumpf: Okay so I just add the error as additional parameter?

manuel-rw (approved):

PR #2226, opened 2025-02-03
feat: Support aria2 in download widget

hotrungnhan:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Change Logs: 

* Support Aria2 in download widget. 
* Download item detail:
     * Conditional render `upspeed`, `sent` and `ratio` to torrent's item only.
* Adjust the `DownloadClientStatus` interface to support the multiple download type client such aria2.

Limitations: 

 * Remove operation does not remove downloaded files (Aria2 RPC does not support this).
 * The waiting and finished entries only listing first 1000 items. 
 * Can only pause `downloading` and `leeching`.
 * The 'error' and `completed` entries can only be removed.

### DIFF
--- a/packages/definitions/src/integration.ts
+++ b/packages/definitions/src/integration.ts
@@ -49,6 +49,12 @@ export const integrationDefs = {
     iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons@master/svg/qbittorrent.svg",
     category: ["downloadClient", "torrent"],
   },
+  aria2: {
+    name: "Aria2",
+    secretKinds: [[], ["apiKey"]],
+    iconUrl: "https://cdn.jsdelivr.net/gh/PapirusDevelopmentTeam/papirus_icons@latest/src/system_downloads_3.svg",
+    category: ["downloadClient", "torrent", "miscellaneous"],
+  },
   sonarr: {
     name: "Sonarr",
     secretKinds: [["apiKey"]],
@@ -211,6 +217,7 @@ export type IntegrationCategory =
   | "downloadClient"
   | "usenet"
   | "torrent"
+  | "miscellaneous"
   | "smartHomeServer"
   | "indexerManager"
   | "healthMonitoring"
--- a/packages/integrations/src/download-client/aria2/aria2-integration.ts
+++ b/packages/integrations/src/download-client/aria2/aria2-integration.ts
@@ -1,0 +1,180 @@
+import path from "path";
+
+import { fetchWithTrustedCertificatesAsync } from "@homarr/certificates/server";
+
+import type { DownloadClientJobsAndStatus } from "../../interfaces/downloads/download-client-data";
+import { DownloadClientIntegration } from "../../interfaces/downloads/download-client-integration";
+import type { DownloadClientItem } from "../../interfaces/downloads/download-client-items";
+import type { Aria2Download, Aria2GetClient } from "./aria2-types";
+
+export class Aria2Integration extends DownloadClientIntegration {
+  public async getClientJobsAndStatusAsync(): Promise<DownloadClientJobsAndStatus> {
+    const client = this.getClient();
+    const keys: (keyof Aria2Download)[] = [
+      "bittorrent",
+      "uploadLength",
+      "uploadSpeed",
+      "downloadSpeed",
+      "totalLength",
+      "completedLength",
+      "files",
+      "status",
+      "gid",
+    ];
+    const [activeDownloads, waitingDownloads, stoppedDownloads, globalStats] = await Promise.all([
+      client.tellActive(),
+      client.tellWaiting(0, 1000, keys),
+      client.tellStopped(0, 1000, keys),
+      client.getGlobalStat(),
+    ]);
+
+    const downloads = [...activeDownloads, ...waitingDownloads, ...stoppedDownloads];
+    const allPaused = downloads.every((download) => download.status === "paused");
+
+    return {
+      status: {
+        types: ["torrent", "miscellaneous"],
+        paused: allPaused,
+        rates: {
+          up: Number(globalStats.uploadSpeed),
+          down: Number(globalStats.downloadSpeed),
+        },
+      },
+      items: downloads.map((download, index) => {
+        const totalSize = Number(download.totalLength);
+        const completedSize = Number(download.completedLength);
+        const progress = totalSize > 0 ? completedSize / totalSize : 0;
+
+        const itemName = download.bittorrent?.info?.name ?? path.basename(download.files[0]?.path ?? "Unknown");
+
+        return {
+          index,
+          id: download.gid,
+          name: itemName,
+          type: download.bittorrent ? "torrent" : "miscellaneous",
+          size: totalSize,
+          sent: Number(download.uploadLength),
+          downSpeed: Number(download.downloadSpeed),
+          upSpeed: Number(download.uploadSpeed),
+          time: this.calculateEta(completedSize, totalSize, Number(download.downloadSpeed)),
+          state: this.getState(download.status, Boolean(download.bittorrent)),
+          category: [],
+          progress,
+        };
+      }),
+    } as DownloadClientJobsAndStatus;
+  }
+  public async pauseQueueAsync(): Promise<void> {
+    const client = this.getClient();
+    await client.pauseAll();
+  }
+  public async pauseItemAsync(item: DownloadClientItem): Promise<void> {
+    const client = this.getClient();
+    await client.pause(item.id);
+  }
+  public async resumeQueueAsync(): Promise<void> {
+    const client = this.getClient();
+    await client.unpauseAll();
+  }
+  public async resumeItemAsync(item: DownloadClientItem): Promise<void> {
+    const client = this.getClient();
+    await client.unpause(item.id);
+  }
+  public async deleteItemAsync(item: DownloadClientItem, fromDisk: boolean): Promise<void> {
+    const client = this.getClient();
+    // Note: Remove download file is not support by aria2, replace with forceremove
+
+    if (item.state in ["downloading", "leeching", "paused"]) {
+      await (fromDisk ? client.remove(item.id) : client.forceRemove(item.id));
+    } else {
+      await client.removeDownloadResult(item.id);
+    }
+  }
+
+  public async testConnectionAsync(): Promise<void> {
+    const client = this.getClient();
+    await client.getVersion();
+  }
+
+  private getClient() {
+    const url = this.url("/jsonrpc");
+
+    return new Proxy(
+      {},
+      {
+        get: (target, method: keyof Aria2GetClient) => {
+          return async (...args: Parameters<Aria2GetClient[typeof method]>) => {
+            let params = [...args];
+            if (this.hasSecretValue("apiKey")) {
+              params = [`token:${this.getSecretValue("apiKey")}`, ...params];
+            }
+            const body = JSON.stringify({
+              jsonrpc: "2.0",
+              id: btoa(["Homarr", Date.now().toString(), Math.random()].join(".")), // unique id per request
+              method: `aria2.${method}`,
+              params,
+            });
+            return await fetchWithTrustedCertificatesAsync(url, { method: "POST", body })
+              .then(async (response) => {
+                const responseBody = (await response.json()) as { result: ReturnType<Aria2GetClient[typeof method]> };
+
+                if (!response.ok) {
+                  throw new Error(response.statusText);
+                }
+                return responseBody.result;
+              })
+              .catch((error) => {
+                if (error instanceof Error) {
+                  throw new Error(error.message);
+                } else {
+                  throw new Error("Error communicating with Aria2");
+                }
+              });
+          };
+        },
+      },
+    ) as Aria2GetClient;
+  }
+
+  private getState(aria2Status: Aria2Download["status"], isTorrent: boolean): DownloadClientItem["state"] {
+    return isTorrent ? this.getTorrentState(aria2Status) : this.getNonTorrentState(aria2Status);
+  }
+  private getNonTorrentState(aria2Status: Aria2Download["status"]): DownloadClientItem["state"] {
+    switch (aria2Status) {
+      case "active":
+        return "downloading";
+      case "waiting":
+        return "queued";
+      case "paused":
+        return "paused";
+      case "complete":
+        return "completed";
+      case "error":
+        return "failed";
+      case "removed":
+      default:
+        return "unknown";
+    }
+  }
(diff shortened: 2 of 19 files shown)


packages/definitions/src/integration.ts
```diff
@@ -49,6 +49,12 @@ export const integrationDefs = {
     iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons@master/png/qbittorrent.png",
     category: ["downloadClient", "torrent"],
   },
+  aria2: {
+    name: "Aria2",
+    secretKinds: [["apiKey"]],
+    iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons@master/png/aria2.png",
```
manuel-rw, line 55: This gives me ``Package size exceeded the configured limit of 50 MB. Try https://github.com/homarr-labs/dashboard-icons/tree/master/png/aria2.png instead.``

packages/definitions/src/integration.ts
```diff
@@ -49,6 +49,12 @@ export const integrationDefs = {
...
+    category: ["downloadClient", "torrent", "http(s)"],
```
manuel-rw, line 56: I dislike that the category has brackets in it's name. Just a suggestion: ``multiProtocolDownloadClient``?
@Meierschlumpf / @SeDemal WDYT?

packages/integrations/src/download-client/aria2/aria2-client.ts
```diff
@@ -0,0 +1,73 @@
+
+import { aria2, Conn, ReadyState, open, PreconfiguredSocket, OpenOptions } from "maria2/dist"
```
manuel-rw, line 2: Check lint errors. For future reference, please fix those before you request a review.

packages/integrations/src/download-client/aria2/aria2-client.ts
```diff
@@ -0,0 +1,73 @@
...
+import { fetch } from "undici";
+
+type Aria2Type = typeof aria2
+
+// Extract method argument types and return types
+type MethodArguments<T> = T extends (conn: Conn, ...args: infer A) => any ? A : never;
```
manuel-rw, line 8: Same applies here and next line. Please fix. See the ``lint`` action.

packages/integrations/src/download-client/aria2/aria2-client.ts
```diff
@@ -0,0 +1,73 @@
...
+                },
+            }).then(async (res) => {
+                this.dispatchEvent(new MessageEvent('message', { data: await res.text() }))
+            });
+        }
+    }) as any;
+}
+
+export async function CreateAria2(this: any, config: Aria2Config) {
+    const conn = await open(getSocket(config));
```
manuel-rw, line 58:
```suggestion
    const connection = await open(getSocket(config));
```

packages/integrations/src/download-client/aria2/aria2-client.ts
```diff
@@ -0,0 +1,73 @@
...
+    return new Proxy({ conn, config: config }, {
+        get: (target, prop: string) => {
+            if (prop === "then") return this;
```
manuel-rw, line 61: Add a comment why this needs to be done (or what this function does)

packages/integrations/src/download-client/aria2/aria2-client.ts
```diff
@@ -0,0 +1,73 @@
...
+            if (prop in aria2) {
+                return (...args: any[]) => {
+                    const method = (aria2 as any)[prop];
+                    return method(conn, ...args);
+                };
+            }
+            throw new Error(`Method ${prop} does not exist`);
+        },
+    }) as any as Aria2ProxyHandler;
```
manuel-rw, line 70: Are you sure about this cast?

packages/integrations/src/download-client/aria2/aria2-integration.ts
```diff
@@ -0,0 +1,151 @@
...
+import type { DownloadClientItem } from "../../interfaces/downloads/download-client-items";
+import { CreateAria2, Aria2DownloadState } from './aria2-client';
+import path from "path"
+
+
+export class Aria2Integration extends DownloadClientIntegration {
+  public async getClientJobsAndStatusAsync(): Promise<DownloadClientJobsAndStatus> {
+    const client = await this.getClientAsync();
+    // ?: Should recursive retrive all items or just few first item?
+    const [actives, waitings, stoppeds, globalstats] = await Promise.all([client.tellActive(), client.tellWaiting(0, 1000), client.tellStopped(0, 1000), client.getGlobalStat()]);
```
manuel-rw, line 14: So this will only fetch 1000 items?

packages/integrations/src/download-client/aria2/aria2-integration.ts
```diff
@@ -0,0 +1,151 @@
...
+import { DownloadClientIntegration } from "../../interfaces/downloads/download-client-integration";
+import type { DownloadClientItem } from "../../interfaces/downloads/download-client-items";
+import { CreateAria2, Aria2DownloadState } from './aria2-client';
+import path from "path"
+
+
+export class Aria2Integration extends DownloadClientIntegration {
+  public async getClientJobsAndStatusAsync(): Promise<DownloadClientJobsAndStatus> {
+    const client = await this.getClientAsync();
+    // ?: Should recursive retrive all items or just few first item?
```
manuel-rw, line 13: You're referring to a ``?`` but there is none on the next line

packages/integrations/src/download-client/aria2/aria2-integration.ts
```diff
@@ -0,0 +1,151 @@
...
+    return {
+      status: {
+        types: ["torrent", "http(s)"],
+        paused: false,
+        rates: {
+          up: Number(globalstats.uploadSpeed),
+          down: Number(globalstats.downloadSpeed),
+        }
+      },
+      items: downloadings.map((item, idx) => {
```
manuel-rw, line 27:
```suggestion
      items: downloadings.map((download, index) => {
```

packages/integrations/src/download-client/aria2/aria2-integration.ts
```diff
@@ -0,0 +1,151 @@
...
+  public async pauseItemAsync(item: DownloadClientItem): Promise<void> {
+    const client = await this.getClientAsync();
+    switch (item.state) {
+      case "downloading":
+      case "leeching":
+        await client.pause(item.id);
+        return;
+      default:
+        return;
+    }
```
manuel-rw, line 64: This is bad practice - there is only one branch and AFAIK you won't get any performance advantage (is this correct  @Meierschlumpf )? Use a normal ``if`` statement instead.

packages/integrations/src/download-client/aria2/aria2-integration.ts
```diff
@@ -0,0 +1,151 @@
...
+        return
+      default:
+        await client.removeDownloadResult(item.id)
+        return;
+    }
+  }
+
+  public async testConnectionAsync(): Promise<void> {
+    const client = await this.getClientAsync();
+    await client.getVersion()
```
manuel-rw, line 97:
```suggestion
    await client.getVersion();
```

packages/integrations/test/aria2.spec.ts
manuel-rw, line 1: Why an empty file?

manuel-rw (changes requested):

packages/definitions/src/integration.ts
Meierschlumpf: Yeah the icon does not exist and has to be added to the dashboard-icons repository

SeDemal:
I'll need to spend some more time later on for a proper review.
Few things to note though:
 - Properly check all the tests (pnpm -> fomrat:fix, lint, typecheck)
 - Make the integration fit the current template as much as possible (I'm unsure if it's relevant here, just a reminder)
 - If you're going to make your own client interpretation of the dependency, then why not get rid of it? It seems to me that a http RPC function should be fairly easy here from what I see in the docs. (nzbGet uses RPC as well, if you want a template)
 - Use the proper fetch function, we use "fetchWithTrustedCertificatesAsync" because of certificates.

hotrungnhan:
i'll tag your guy whenever i finished.

packages/definitions/src/integration.ts
hotrungnhan: Aria2 don't have any offical icons. We can use the one below as it icon, or if you find any better, please help add to the repo.
aria2

packages/definitions/src/integration.ts
hotrungnhan: updated to use "miscellaneous"

hotrungnhan:
@SeDemal @manuel-rw @Meierschlumpf finished. can you help review again ?

packages/definitions/src/integration.ts
Meierschlumpf: aria2

I would suggest we use the above one. It is an svg

Meierschlumpf:
> @SeDemal @manuel-rw @Meierschlumpf finished. can you help review again ?
> 
> is there existing way to disable the action button base on download item state ?

What do you mean with disable the action button base?

packages/integrations/src/download-client/aria2/aria2-integration.ts
```diff
@@ -0,0 +1,181 @@
...
+      "files",
+      "status",
+      "gid",
+    ];
+    const [actives, waitings, stoppeds, globalstats] = await Promise.all([
+      client.tellActive(),
+      client.tellWaiting(0, 1000, keys),
+      client.tellStopped(0, 1000, keys),
+      client.getGlobalStat(),
+    ]);
```
Meierschlumpf, line 29: I would suggest that we don't load 1000 entries as max here and instead something like 50 / 100 to not fill the memory to much (all data fetched is cached in redis)

packages/integrations/src/download-client/aria2/aria2-integration.ts
```diff
@@ -0,0 +1,181 @@
...
+          };
+        },
+      },
+    ) as Aria2GetClient;
+  }
+
+  static getState(aria2Status: Aria2Download["status"], isTorrent: boolean): DownloadClientItem["state"] {
+    return isTorrent
+      ? Aria2Integration.getNonTorrentState(aria2Status)
+      : Aria2Integration.getNonTorrentState(aria2Status);
```
Meierschlumpf, line 141: I think the first one should call getTorrentState instead

packages/translation/src/lang/vi.json
Meierschlumpf, line 1: Do not change translations here. On github we only specify the english translations. If you want to translate it later, you can do so on crowdin:
https://crowdin.com/project/homarr_labs

packages/widgets/src/downloads/component.tsx
```diff
@@ -726,15 +730,22 @@ const ItemInfoModal = ({ items, currentIndex, opened, onClose }: ItemInfoModalPr
           <NormalizedLine itemKey="index" values={item.index} />
           <NormalizedLine itemKey="type" values={item.type} />
           <NormalizedLine itemKey="state" values={t(item.state)} />
-          <NormalizedLine
-            itemKey="upSpeed"
-            values={item.upSpeed === undefined ? undefined : humanFileSize(item.upSpeed, "/s")}
-          />
+          {item.type === "torrent" && (
```
Meierschlumpf, line 733: I think this should probably instead be `!== "miscellaneous"`, because otherwise other integrations will also no longer show this (same for sent & ratio)

packages/widgets/src/downloads/component.tsx
```diff
@@ -745,9 +756,11 @@ const ItemInfoModal = ({ items, currentIndex, opened, onClose }: ItemInfoModalPr
               unitDisplay: "narrow",
             }).format(item.progress)}
           />
-          <NormalizedLine itemKey="ratio" values={item.ratio} />
+          {item.type === "torrent" && <NormalizedLine itemKey="ratio" values={item.ratio} />}
           <NormalizedLine itemKey="added" values={item.added === undefined ? "unknown" : dayjs(item.added).format()} />
-          <NormalizedLine itemKey="time" values={item.time !== 0 ? dayjs().add(item.time).format() : "∞"} />
+          {item.added && (
```
Meierschlumpf, line 761: Why is this rendered conditionally?

Meierschlumpf (changes requested):

manuel-rw:
Any updates on this @hotrungnhan ? Please have a look at the open discussions and implement them if you agree. Thank you!

manuel-rw:
To avoid stale PRs, I'm taking over this...

packages/integrations/src/download-client/aria2/aria2-integration.ts
manuel-rw: Can't we use a parameter for how many to fetch?
Also, unless we load binaries into the memory, we should be fine even when loading thousands of entries, no?
As a comparison, 100 paragraphs of the lorem ipsum text is only 0.04MB - which should be easily to handle by Redis.
See https://redis.io/docs/latest/operate/rs/databases/memory-performance/memory-limit/

Meierschlumpf: Sounds like a good idea 👍🏼

manuel-rw:
Tests are not passing, not sure why... Can you have a look, @Meierschlumpf ? Did they already fail in the first iteration?

Meierschlumpf:
Not sure, but it seems like the test failed already previously

manuel-rw:
IMO we can merge as is and users can create bug reports, if the integration doesn't work as expected.
Aria2 seems rather niche. We could skip the tests for now.

packages/definitions/src/integration.ts
```diff
@@ -49,6 +49,12 @@ export const integrationDefs = {
     iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons@master/svg/qbittorrent.svg",
     category: ["downloadClient", "torrent"],
   },
+  aria2: {
+    name: "Aria2",
+    secretKinds: [[], ["apiKey"]],
```
Meierschlumpf, line 54:
```suggestion
    secretKinds: [["apiKey"], []],
```

I suggest we put the api key option first as we want to encourage people to use auth

Meierschlumpf (changes requested):
Also fix the useMemo lint issue in the component, after that it seems OK to me

packages/definitions/src/integration.ts
Meierschlumpf: As discussed we leave it like that for now

manuel-rw: As discussed, by default there is no API key and I couldn't figure out where to obtain it. Let's keep it like this.

PR #2228, opened 2025-02-03
feat(board): allow to set icon color of widgets

Aandree5:
- Added a new color option to the board to allow picking an icon color
- Added new css theme variable for the icon color
- Updated app, bookmarks and dns-hole>controls widgets to take this new color into account when displaying icons 
- Leaving the color field empty, will use the respective  icon colors

Done this by leveraging the css mantine theme library, by adding a new `iconColour` color option plus an `hasIconColor` in the `other` section, so it can default to the default app icon. Not sure if this would be the best way to do it or you would have a better way to achieve this, I can adjust it if needed.

<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


packages/db/schema/mysql.ts
```diff
@@ -272,6 +272,7 @@ export const boards = mysqlTable("board", {
   opacity: int().default(100).notNull(),
   customCss: text(),
   columnCount: int().default(10).notNull(),
+  iconColor: text().default("").notNull(),
```
Meierschlumpf, line 275: I would suggest to make it nullable instead and replace a trimmed empty string with null when updating the database field in the router

packages/widgets/src/app/component.tsx
```diff
@@ -65,7 +66,7 @@ export default function AppWidget({ options, isEditMode }: WidgetComponentProps<
         styles={{ tooltip: { maxWidth: 300 } }}
       >
         <Flex
-          className={combineClasses("app-flex-wrapper", app.name, app.id)}
+          className={combineClasses("app-flex-wrapper", app.name, app.id, app.href && classes.appWrapper)}
```
Meierschlumpf, line 69: This change seems incorrect to me what is the goal here?

packages/widgets/src/app/component.tsx
```diff
@@ -78,7 +79,27 @@ export default function AppWidget({ options, isEditMode }: WidgetComponentProps<
...
+                WebkitMaskSize: "contain",
+                WebkitMaskRepeat: "no-repeat",
+                WebkitMaskPosition: "center",
+                WebkitMaskImage: `url(${app.iconUrl})`,
+                maskSize: "contain",
+                maskRepeat: "no-repeat",
+                maskPosition: "center",
+                maskImage: `url(${app.iconUrl})`,
+              }}
+            />
```
Meierschlumpf, line 99: I would suggest to make a shared component in the ui package for this image / image with backdrop or at least a method to reuse the style properties so it has the logic centralized. I actually did not know that css can do this. It's great

packages/widgets/src/bookmarks/component.tsx
```diff
@@ -50,6 +64,7 @@ export default function BookmarksWidget({ options, width, height, itemId }: Widg
           hideIcon={options.hideIcon}
           hideHostname={options.hideHostname}
           openNewTab={options.openNewTab}
+          hasIconColor={theme.other.hasIconColor as boolean}
```
Meierschlumpf, line 67: Maybe we can define the type of this through mantine

Meierschlumpf (changes requested):

packages/db/schema/mysql.ts
Aandree5: Ahh I think I see waht you're saying, I actually started with null but the color picker complained does not accept nulls and I got confused, will update the code

packages/widgets/src/app/component.tsx
Aandree5: This was just so the hover effects, on the icon, only happen if there is an url defined, I guess I could always have the class there fixed, and create another class on the icon that would trigger only if `href` is defined

packages/widgets/src/app/component.tsx
Aandree5: That's a great point, any idea where would be the best place for such component to live?

packages/widgets/src/bookmarks/component.tsx
Aandree5: I was going to have another look at this, wasn't happy with it either, will find a way

Aandree5: Cany you give me some pointer with this, I followed Matine docs and there is a way to define the type of the `other` properties. It works fine in the `project` where I defined the `theme.other.d.ts` file, it gets imported correctly by `tsconfig.json`, but on the actual component taht calls `theme.other.hasIconColor` it's still showing as `any`. If I create the same file in that `src` folder or edit the `tsconfig.json` file to point that the previous file `apps/nextjs/src/theme/theme.other.d.ts` it also works ok. 

My quesiton is without having to create/maintain potentially several files (only two at the moment), or editing all `tsconfig.json` files, how can I set it up? Is there a common locaiton where all `tsconfig.json` files will be able to import the `theme.other.d.ts` file?

Aandree5: Found a way by using 
```
    "typeRoots": ["../../types"],
    "types": ["theme.other"]
```

on the base file for `tsconfig.json` and it works everywhere, now the test was I created a folder on the root called `types`, but I'm sure there's a better place to put the file. Just not sure if you already have somewhere where it's best to place it

Aandree5: I've tried an approach, by creating a new @homarr/types package to old the common `*.d.ts` types and installed it on the `nextjs` and `widgets` projects and addded the new type to `tsconfig.json`. This seems to work nicely, let me know your thoughts on it.

The `tsconfig.json` seems to help with intelisense (which wasn't working without it), the `pnpm lint` and `pnpm typecheck` commands don't seems to require it

packages/widgets/src/app/component.tsx
Meierschlumpf: Put it into `@homarr/ui` so in the `packages/ui` where you create a shared component for it

packages/db/schema/mysql.ts
Aandree5: Fixed with https://github.com/homarr-labs/homarr/pull/2228/commits/e9aca0d642262d5af74f2db10a2b863aec0148de

packages/widgets/src/app/component.tsx
Aandree5: Clarified with https://github.com/homarr-labs/homarr/pull/2228/commits/00a6be719a010be5e85a460e3320c597a2cbfc49

This allows the over effect (of the icon) to show when the user hovers over the app widget, before it would only happen if the user was hovering the icon itself

packages/widgets/src/app/component.tsx
Aandree5: Done https://github.com/homarr-labs/homarr/pull/2228/commits/cb2f5743f125e042e3efe45708421c811810993f

packages/widgets/src/bookmarks/component.tsx
Aandree5: Not sure if this is the best approach, I'm here to adjust if needed. Done with https://github.com/homarr-labs/homarr/pull/2228/commits/4f682b7d792a38addb211a279b39bb9506632c69 and https://github.com/homarr-labs/homarr/pull/2228/commits/1556dce5633bf3bca85b8042b83aaed205c34f0a

packages/ui/src/components/image-colored.tsx
```diff
@@ -0,0 +1,29 @@
+import { getThemeColor, MantineColor, useMantineTheme } from "@mantine/core";
+
+interface ImageColoredProps {
+  src: string;
+  color: MantineColor;
+  alt: string;
+  styles?: React.CSSProperties;
```
Meierschlumpf, line 7: Rename to `style`

packages/ui/src/components/image-colored.tsx
```diff
@@ -0,0 +1,29 @@
...
+  className?: string;
+}
+
+export const ImageColored = ({ src, color, alt, styles, className }: ImageColoredProps) => {
```
Meierschlumpf, line 11: Rename to `MaskedImage`

packages/validation/src/board.ts
```diff
@@ -13,6 +13,12 @@ import { commonItemSchema, createSectionSchema } from "./shared";
 
 const hexColorSchema = z.string().regex(/^#[0-9A-Fa-f]{6}$/);
 
+const hexColorNullableSchema = z
+  .string()
+  .regex(/^(#[0-9A-Fa-f]{6}|)$/)
```
Meierschlumpf, line 18:
```suggestion
const hexColorNullableSchema = hexColorSchema
```

packages/widgets/src/dns-hole/controls/component.tsx
```diff
@@ -322,14 +330,29 @@ const ControlsCard: React.FC<ControlsCardProps> = ({
...
+            className="dns-hole-controls-item-icon"
+            src={integrationDefs[data.integration.kind].iconUrl}
+            w="20cqmin"
+            h="20cqmin"
+            fit="contain"
+            style={{
+              filter: !isConnected ? "grayscale(100%)" : undefined,
+            }}
+          />
+        )}
```
Meierschlumpf, line 355: It would be nice as well if you could put this in a component in the up package. Like 

```tsx
<MaskedOrNormalImage
	color="iconColor"
	src={iconUrl}
 	alt={data.integration.name}
	className="dns-hole-controls-item-icon"
	fit="contain" // Add new property to MaskedImage to pass to maskSize
        style={{
		height: "20cqmin",
		width: "20cqmin",
              	filter: !isConnected ? "grayscale(100%)" : undefined,
	}} />
```

Meierschlumpf (changes requested):
Some other things to do:

1. Drop db migrations 23 / 24 (in packages/db run `pnpm migration:sqlite:drop` (and mysql) and select the migrations to drop
2. Move the theme override to @homarr/ui and remove the types package
3. I'll also take a look regarding the types if there is a better way to automatically make it work in the whole monorepository

packages/validation/src/board.ts
Aandree5: The regex I was using if different than the `hexColorSchema`, the `hexColorNullableSchema` would also allow empty strings.
`hexColorSchema`              = `/^#[0-9A-Fa-f]{6}$/`
`hexColorNullableSchema` = `/^(#[0-9A-Fa-f]{6}|)$/`   =  `/^(hexColorSchema|)$/`
Wrapped it in a group and added a pipe to nothing, in a sense saying "hexColorSchema or nothing".
Could also do something like `hexColorSchema.or(z.string().lenght(0))`

packages/ui/src/components/image-colored.tsx
Aandree5: Done https://github.com/homarr-labs/homarr/pull/2228/commits/5ceca15b453b752d4e821ff748b4fffdaf600575

packages/ui/src/components/image-colored.tsx
Aandree5: Done https://github.com/homarr-labs/homarr/pull/2228/commits/f992c5f1a562c77d5851dc955add9a5d84633be1

Aandree5:
> Some other things to do:
> 
> 1. Drop db migrations 23 / 24 (in packages/db run `pnpm migration:sqlite:drop` (and mysql) and select the migrations to drop
> 2. Move the theme override to @homarr/ui and remove the types package
> 3. I'll also take a look regarding the types if there is a better way to automatically make it work in the whole monorepository

There only one 23 showing when I try to drop both sqlite or mysql. Do I just remove the files manually?

Meierschlumpf:
No removing the files manually will probably break as it also modifies the `_journal.json`
Yes you only have one 23 showing for sqlite and one showing for mysql. After you've removed it you can then also merge the changes from dev (where some db migrations were made)

packages/validation/src/board.ts
Meierschlumpf: Oh haven't seen that, maybe instead use `hexColorSchema.or(z.literal(''))`

Aandree5:
This should all be sorted now, just a question:
- I've noticed someone used `const board = useRequiredBoard()` on the app widget, would it be worth to change the `theme.other.hasIconColor` on the theme (which we are trying to find a way to cascade the type around all the projects) to use this variable and go straight to the board properties to see if a color has been defined?

Also, please note since I've dropped the migrations I've not generated a new one with the changes to the board scheme, I'm guessing that is done when merging to `dev` but let me know if you want me run them.

Aandree5 (commented):
This should all be sorted now, just a question:

I've noticed someone used const board = useRequiredBoard() on the app widget, would it be worth to change the theme.other.hasIconColor on the theme (which we are trying to find a way to cascade the type around all the projects) to use this variable and go straight to the board properties to see if a color has been defined?
Also, please note since I've dropped the migrations I've not generated a new one with the changes to the board scheme, I'm guessing that is done when merging to dev but let me know if you want me run them.

Meierschlumpf:
Yeah maybe it is a good idea to not have this type hacks in the project and just use the `useRequiredBoard` hook 👍🏼

Aandree5:
Tear when i são that as was thinking the same. I will try to make those changes this weekend, they shouldn't take long.
It's fine don't worry! I can see that the project is busy, which is good!

Aandree5:
It should all be done!

Aandree5 (commented):
It should all be done!

Meierschlumpf:
I'll take a look at it tomorrow evening and will also fix the merge conflicts if you want 👍🏼 Thank you for your work ❤️

Aandree5:
> I'll take a look at it tomorrow evening and will also fix the merge conflicts if you want 👍🏼 Thank you for your work ❤️

Just happy to help. I've aaceepted all incoming changes on the migrations and re-generated the required ones, should be sorted but feel free to let me know if not :)

Meierschlumpf (approved):
Thank you for this nice feature! I already look forward to the upcoming show-offs

PR #2248, opened 2025-02-04
feat: quick add app modal

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

## Step 1
image
## Step 2
image
## Step 3
image


apps/nextjs/src/app/[locale]/manage/apps/edit/[id]/_app-edit-form.tsx
```diff
@@ -55,6 +54,7 @@ export const AppEditForm = ({ app }: AppEditFormProps) => {
       buttonLabels={{
         submit: t("common.action.save"),
       }}
+      showBackToOverview={false}
```
Comment (Meierschlumpf) [inline apps/nextjs/src/app/[locale]/manage/apps/edit/[id]/_app-edit-form.tsx:57]: Why do we hide this button here?

packages/api/src/router/app.ts
```diff
@@ -111,16 +111,19 @@ export const appRouter = createTRPCRouter({
   create: permissionRequiredProcedure
     .requiresPermission("app-create")
     .input(validation.app.manage)
-    .output(z.void())
+    .output(z.object({ appId: z.string() }))
```
Meierschlumpf, line 114: Does it make sense to name it appId? I would say this is unnecessary

packages/widgets/src/_inputs/widget-app-input.tsx
```diff
@@ -18,42 +20,63 @@ export const WidgetAppInput = ({ property, kind }: CommonWidgetInputProps<"app">
...
+            // eslint-disable-next-line no-restricted-syntax
+            async onClose(createdAppId) {
+              await refetch();
+              form.setFieldValue(`options.${property}`, createdAppId);
+            },
+          })
+        }
+      >
+        {t("widget.common.app.quickCreate")}
+      </Button>
```
Meierschlumpf, line 78: To also look good on mobile devices I would suggest an `ActionIcon.Group` with a `ActionIcon.GroupSection` for the input

Meierschlumpf (changes requested):

apps/nextjs/src/app/[locale]/manage/apps/edit/[id]/_app-edit-form.tsx
Comment (manuel-rw) [reply apps/nextjs/src/app/[locale]/manage/apps/edit/[id]/_app-edit-form.tsx:57]: Because it will redirect to ``/manage/apps`` which is not what we want in the quick add

packages/api/src/router/app.ts
manuel-rw: In my opinion yes, because it is not clear from the ``create`` verb what this ID is used for

packages/widgets/src/_inputs/widget-app-input.tsx
manuel-rw: Done:
image
image

The ActionIcon didn't look good and didn't work - hence I used this solution.

packages/api/src/router/app.ts
Meierschlumpf: IMO it is already known in the contexts that it is used, but let's leave it like that

apps/nextjs/src/app/[locale]/manage/apps/edit/[id]/_app-edit-form.tsx
Comment (Meierschlumpf) [reply apps/nextjs/src/app/[locale]/manage/apps/edit/[id]/_app-edit-form.tsx:57]: But the _app-edit-form is only used on the `apps/edit/{id}` page where it would make sense to show that button, no?

packages/widgets/src/_inputs/widget-app-input.tsx
Meierschlumpf: I don't like that it takes so much attention to this button, but yeah let's leave it like that for now

Meierschlumpf (commented):
Please resolve the thing with the _app-edit-form, after that it's good for me

Meierschlumpf (approved):

PR #2268, opened 2025-02-07
fix(calendar): info popover overlap

hillaliy:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

### DIFF
--- a/packages/widgets/src/calendar/calender-day.tsx
+++ b/packages/widgets/src/calendar/calender-day.tsx
@@ -39,12 +39,12 @@ export const CalendarDay = ({ date, events, disabled }: CalendarDayProps) => {
           style={{
             alignContent: "center",
             borderRadius: "3.5cqmin",
-            cursor: events.length === 0 || disabled ? "default" : "pointer",
+            cursor: disabled ? "default" : "pointer",
           }}
           onClick={() => {
-            if (events.length > 0 && !disabled) {
-              setOpend((prev) => !prev);
-            }
+            if (disabled) return;
+
+            setOpend((prev) => !prev);
           }}
         >
           <div
--- a/packages/widgets/src/calendar/calender-day.tsx
+++ b/packages/widgets/src/calendar/calender-day.tsx
@@ -25,7 +25,7 @@ export const CalendarDay = ({ date, events, disabled }: CalendarDayProps) => {
       transitionProps={{
         transition: "pop",
       }}
-      onChange={() => setOpend(false)}
+      onChange={setOpend}
       opened={opened}
       disabled={disabled}
     >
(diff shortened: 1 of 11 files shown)


hillaliy:
Sorry about Semantic Commits, my mistake

Meierschlumpf:
No problem 😉

Meierschlumpf:
Can you check why the popover does not close? Mantine states it should close by default:
https://mantine.dev/core/popover/#click-outside

I would guess that the problem is that we use onClose. Can you try out with onChange instead and instead of useDisclosure use useState? Then also don't add onClick in the container as it's already done through the target.

In the component.tsx you will also need to change the line:

```
return <CalendarDay date={tileDate} events={eventsForDate} disabled={isEditMode} />;
```

to

```
return <CalendarDay date={tileDate} events={eventsForDate} disabled={isEditMode || events.length === 0} />;
```

as we want to disable the popover when no event exists for the day

hillaliy:
> Can you check why the popover does not close? Mantine states it should close by default: https://mantine.dev/core/popover/#click-outside
> 
> I would guess that the problem is that we use onClose. Can you try out with onChange instead and instead of useDisclosure use useState? Then also don't add onClick in the container as it's already done through the target.
> 
> In the component.tsx you will also need to change the line:
> 
> ```
> return <CalendarDay date={tileDate} events={eventsForDate} disabled={isEditMode} />;
> ```
> 
> to
> 
> ```
> return <CalendarDay date={tileDate} events={eventsForDate} disabled={isEditMode || events.length === 0} />;
> ```
> 
> as we want to disable the popover when no event exists for the day

Removing onClick from container

packages/widgets/src/calendar/calender-day.tsx
```diff
@@ -25,13 +25,12 @@ export const CalendarDay = ({ date, events, disabled }: CalendarDayProps) => {
       transitionProps={{
         transition: "pop",
       }}
-      onClose={close}
+      onChange={() => setOpend(false)}
```
Meierschlumpf, line 28:
```suggestion
      onChange={setOpened}
```

packages/widgets/src/calendar/calender-day.tsx
```diff
@@ -42,6 +41,11 @@ export const CalendarDay = ({ date, events, disabled }: CalendarDayProps) => {
             borderRadius: "3.5cqmin",
             cursor: events.length === 0 || disabled ? "default" : "pointer",
           }}
+          onClick={() => {
+            if (events.length > 0 && !disabled) {
+              setOpend((prev) => !prev);
+            }
+          }}
```
Meierschlumpf, line 48:
```suggestion
```

Meierschlumpf (changes requested):

packages/widgets/src/calendar/calender-day.tsx
hillaliy: I tryed to remove it but the popover don't pop

Meierschlumpf: Okay I just tried it locally and you are right. I made some minor adjustments regarding the disabled check

Meierschlumpf (approved):

Meierschlumpf:
Thank you!

PR #2288, opened 2025-02-09
feat: docker widget

hillaliy:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

### DIFF
--- a/packages/api/src/router/docker/docker-router.ts
+++ b/packages/api/src/router/docker/docker-router.ts
@@ -3,18 +3,21 @@ import { z } from "zod";
 
 import { db, like, or } from "@homarr/db";
 import { icons } from "@homarr/db/schema";
-import { DockerSingleton } from "@homarr/docker";
 import type { Container, ContainerInfo, ContainerState, Docker, Port } from "@homarr/docker";
+import { DockerSingleton } from "@homarr/docker";
 import { logger } from "@homarr/log";
 import { createCacheChannel } from "@homarr/redis";
 
 import { createTRPCRouter, permissionRequiredProcedure } from "../../trpc";
 
 const dockerCache = createCacheChannel<{
-  containers: (ContainerInfo & { instance: string; iconUrl: string | null })[];
+  containers: (ContainerInfo & { instance: string; iconUrl: string | null; cpuUsage: number; memoryUsage: number })[];
 }>("docker-containers", 5 * 60 * 1000);
 
+const previousStats = new Map<string, { totalUsage: number; systemUsage: number }>();
+
 export const dockerRouter = createTRPCRouter({
+  // The first time getContainers runs, there’s no previous CPU usage data, so the percentage will be 0
   getContainers: permissionRequiredProcedure.requiresPermission("admin").query(async () => {
     const result = await dockerCache
       .consumeAsync(async () => {
@@ -23,10 +26,33 @@ export const dockerRouter = createTRPCRouter({
           // Return all the containers of all the instances into only one item
           dockerInstances.map(({ instance, host: key }) =>
             instance.listContainers({ all: true }).then((containers) =>
-              containers.map((container) => ({
-                ...container,
-                instance: key,
-              })),
+              Promise.all(
+                containers.map(async (container) => {
+                  const stats = await instance.getContainer(container.Id).stats({ stream: false });
+                  const prevStats = previousStats.get(container.Id);
+                  const totalUsage = stats.cpu_stats.cpu_usage.total_usage;
+                  const systemUsage = stats.cpu_stats.system_cpu_usage;
+                  const cpuCount = stats.cpu_stats.online_cpus;
+
+                  let cpuPercent = 0;
+                  if (prevStats) {
+                    const totalDiff = totalUsage - prevStats.totalUsage;
+                    const systemDiff = systemUsage - prevStats.systemUsage;
+                    if (systemDiff > 0) {
+                      cpuPercent = (totalDiff / systemDiff) * cpuCount * 100;
+                    }
+                  }
+
+                  previousStats.set(container.Id, { totalUsage, systemUsage });
+
+                  return {
+                    ...container,
+                    instance: key,
+                    cpuUsage: Math.round(cpuPercent * 100) / 100,
+                    memoryUsage: parseFloat((stats.memory_stats.usage / (1024 * 1024)).toFixed(2)),
+                  };
+                }),
+              ),
             ),
           ),
         ).then((containers) => containers.flat());
@@ -169,10 +195,12 @@ interface DockerContainer {
   image: string;
   ports: Port[];
   iconUrl: string | null;
+  cpuUsage?: number;
+  memoryUsage?: number;
 }
 
 function sanitizeContainers(
-  containers: (ContainerInfo & { instance: string; iconUrl: string | null })[],
+  containers: (ContainerInfo & { instance: string; iconUrl: string | null; cpuUsage: number; memoryUsage: number })[],
 ): DockerContainer[] {
   return containers.map((container) => {
     return {
@@ -183,6 +211,8 @@ function sanitizeContainers(
       image: container.Image,
       ports: container.Ports,
       iconUrl: container.iconUrl,
+      cpuUsage: container.cpuUsage,
+      memoryUsage: container.memoryUsage,
     };
   });
 }
--- a/packages/api/src/router/widgets/index.ts
+++ b/packages/api/src/router/widgets/index.ts
@@ -1,4 +1,5 @@
 import { createTRPCRouter } from "../../trpc";
+import { dockerRouter } from "../docker/docker-router";
 import { appRouter } from "./app";
 import { calendarRouter } from "./calendar";
 import { dnsHoleRouter } from "./dns-hole";
@@ -30,5 +31,6 @@ export const widgetRouter = createTRPCRouter({
   healthMonitoring: healthMonitoringRouter,
   mediaTranscoding: mediaTranscodingRouter,
   minecraft: minecraftRouter,
+  docker: dockerRouter,
   options: optionsRouter,
 });
--- a/packages/definitions/src/widget.ts
+++ b/packages/definitions/src/widget.ts
@@ -20,5 +20,6 @@ export const widgetKinds = [
   "bookmarks",
   "indexerManager",
   "healthMonitoring",
+  "docker",
 ] as const;
 export type WidgetKind = (typeof widgetKinds)[number];
--- a/packages/widgets/src/docker/component.tsx
+++ b/packages/widgets/src/docker/component.tsx
@@ -1,0 +1,317 @@
+"use client";
+
+import type { MantineColor, MantineStyleProp } from "@mantine/core";
+import { ActionIcon, Avatar, Badge, Group, Progress, Stack, Text, Tooltip } from "@mantine/core";
+import type { IconProps } from "@tabler/icons-react";
+import { IconBrandDocker, IconPlayerPlay, IconPlayerStop, IconRotateClockwise } from "@tabler/icons-react";
+import type { MRT_ColumnDef } from "mantine-react-table";
+import { MantineReactTable } from "mantine-react-table";
+
+import type { RouterOutputs } from "@homarr/api";
+import { clientApi } from "@homarr/api/client";
+import { useTimeAgo } from "@homarr/common";
+import { showErrorNotification, showSuccessNotification } from "@homarr/notifications";
+import type { TranslationFunction } from "@homarr/translation";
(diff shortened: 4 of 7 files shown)


hillaliy:
I will remove unused variables

hillaliy:
Screenshot 2025-02-10 at 21 03 07

Screenshot 2025-02-10 at 21 03 15

manuel-rw:
Hi @hillaliy , thanks for the contribution. Please see the lint and DeepSource first. After that, we'll happily review your PR. Thanks!

hillaliy:
Lint and deepSource errors is because unused variables. I'm still thinking if need to add options so if you can review my PR and if you have an idea for widget options. Thanks

packages/api/src/router/docker/docker-router.ts
```diff
@@ -23,10 +26,33 @@ export const dockerRouter = createTRPCRouter({
           // Return all the containers of all the instances into only one item
           dockerInstances.map(({ instance, host: key }) =>
             instance.listContainers({ all: true }).then((containers) =>
-              containers.map((container) => ({
-                ...container,
-                instance: key,
-              })),
+              Promise.all(
+                containers.map(async (container) => {
+                  const stats = await instance.getContainer(container.Id).stats({ stream: false });
```
Meierschlumpf, line 31: I would suggest to store the stats in a seperate cache / redis channel so we don't fetch unnecessary data when only the page is used. Also please find a better way than reading the previous stats

packages/api/src/router/widgets/index.ts
```diff
@@ -30,5 +31,6 @@ export const widgetRouter = createTRPCRouter({
   healthMonitoring: healthMonitoringRouter,
   mediaTranscoding: mediaTranscodingRouter,
   minecraft: minecraftRouter,
+  docker: dockerRouter,
```
Meierschlumpf, line 34: Docker router already exists in the root router

packages/definitions/src/widget.ts
```diff
@@ -20,5 +20,6 @@ export const widgetKinds = [
   "bookmarks",
   "indexerManager",
   "healthMonitoring",
+  "docker",
```
Meierschlumpf, line 23: Rename widget to `docker-container-table` or something similar

Meierschlumpf (changes requested):

packages/api/src/router/docker/docker-router.ts
hillaliy: I need some help with this. I will upload my solution it's not good because CPU prevStats (get 0% all time) also I need help with docker refresh rate. Thanks

hillaliy:
@Meierschlumpf I need your help with this widget. Refresh rate is not good enough.

Meierschlumpf:
1. What do you mean with refresh rate?
2. How can I help you?

Meierschlumpf:
Okay yeah currently it is cached for 5 minutes:

```ts
const dockerCache = createCacheChannel<{
  containers: (ContainerInfo & { instance: string; iconUrl: string | null; cpuUsage: number; memoryUsage: number })[];
}>("docker-containers", 5 * 60 * 1000);
```

I would suggest that we replace this with the request-handlers we use for other integrations as well. Then we can also add a subscription functionallity

ajnart:
@hillaliy i think for this use case we'd want the info to be refreshed when the page loads as well as after every action (start/stop/restart) 
Refreshing it in the background is not be needed since it's not shown on dashboards

Meierschlumpf:
It's the widget and not a modal or something, so we want to refresh the data every few seconds

packages/api/src/router/docker/docker-router.ts
```diff
@@ -75,6 +77,14 @@ export const dockerRouter = createTRPCRouter({
       timestamp,
     };
   }),
+  getContainersWidget: permissionRequiredProcedure.requiresPermission("admin").query(async () => {
+    const innerHandler = dockerContainersRequestHandler.handler({
+      widjetOptions: {},
```
Meierschlumpf, line 82: Rename

packages/request-handler/src/docker.ts
```diff
@@ -0,0 +1,114 @@
...
+      status: z.string(),
+      iconUrl: z.string().nullable(),
+      cpuUsage: z.number(),
+      memoryUsage: z.number(),
+    }),
+  ),
+});
+
+export type DockerContainersStatus = z.infer<typeof responseSchema>;
+const docker = new Docker();
```
Meierschlumpf, line 45: Use docker singleton or add some way the user can choose which of the configured docker instances should be used

packages/request-handler/src/docker.ts
```diff
@@ -0,0 +1,114 @@
...
+
+async function getContainersWithStats() {
+  const containers = await docker.listContainers({ all: true });
+
+  // Extract image name from container
+  const extractImage = (container: ContainerInfo) => container.Image.split("/").at(-1)?.split(":").at(0) ?? "";
```
Meierschlumpf, line 51: Put outside function

packages/widgets/src/docker/component.tsx
```diff
@@ -0,0 +1,311 @@
...
+  const t = useScopedI18n("docker");
+  const containerStates = {
+    created: "cyan",
+    running: "green",
+    paused: "yellow",
+    restarting: "orange",
+    exited: "red",
+    removing: "pink",
+    dead: "dark",
+  } satisfies Record<ContainerState, MantineColor>;
```
Meierschlumpf, line 29: Move outside function

packages/widgets/src/docker/component.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+  const ContainerStateBadge = ({ state }: { state: ContainerState }) => {
+    return (
+      <Badge size="10cqmin" radius="sm" variant="transparent" color={containerStates[state]}>
+        <Text size="6cqmin">{t(`field.state.option.${state}`)}</Text>
+      </Badge>
+    );
+  };
```
Meierschlumpf, line 37: Move outside function

packages/widgets/src/docker/component.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+  const badgeColor = (number: number, state: string) => {
+    if (number === 0 && state !== "running") return "red";
+    if (number < 40) return "green";
+    if (number < 60) return "yellow";
+    if (number < 90) return "orange";
+    return "red";
+  };
```
Meierschlumpf, line 45: Move outside function

packages/widgets/src/docker/component.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+  const safeValue = (value?: number, fallback = 0) => value ?? fallback;
```
Meierschlumpf, line 47: Move outside function

packages/widgets/src/docker/component.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+  const actionIconIconStyle: IconProps["style"] = {
+    height: "var(--ai-icon-size)",
+    width: "var(--ai-icon-size)",
+  };
```
Meierschlumpf, line 52: Move outside function

packages/widgets/src/docker/component.tsx
```diff
@@ -0,0 +1,311 @@
...
+              },
+              onError() {
+                showErrorNotification({
+                  title: t(`action.${action}.notification.error.title`),
+                  message: t(`action.${action}.notification.error.message`),
+                });
+              },
+            },
+          );
+        };
```
Meierschlumpf, line 177: Move outside function and pass translation string as parameter (where needed)

packages/widgets/src/docker/component.tsx
```diff
@@ -0,0 +1,311 @@
...
+  );
+
+  const baseStyle: MantineStyleProp = {
+    "--total-width": "calc(100cqw / var(--total-width))",
+    "--ratio-width": "calc(100cqw / var(--total-width))",
+    "--space-size": "calc(var(--ratio-width) * 0.1)", //Standard gap and spacing value
+    "--text-fz": "calc(var(--ratio-width) * 0.45)", //General Font Size
+    "--icon-size": "calc(var(--ratio-width) * 2 / 3)", //Normal icon size
+    "--mrt-base-background-color": "transparent",
+  };
```
Meierschlumpf, line 238: Move outside function

packages/widgets/src/docker/component.tsx
```diff
@@ -0,0 +1,311 @@
...
+          borderTop: "0.0625rem solid var(--border-color)",
+        }}
+      >
+        <Group gap={1}>
+          <IconBrandDocker size="8cqmin" />
+          <Text size="6cqmin">{t("table.footer", { count: totalContainers })}</Text>
+        </Group>
+        <Group gap="2cqmin">
+          {containerStateCounts &&
+            Object.entries(containerStateCounts).map(([state, count]) =>
```
Meierschlumpf, line 297: Use typesafe `objectEntries` from `@homarr/common` to no longer need the `as keyof typeof containerStates`

packages/widgets/src/docker/component.tsx
```diff
@@ -0,0 +1,311 @@
...
+                <IconRotateClockwise style={actionIconIconStyle} />
+              </ActionIcon>
+            </Tooltip>
+          </Group>
+        );
+      },
+    },
+  ];
+
+  const { data } = clientApi.docker.getContainersWidget.useQuery();
```
Meierschlumpf, line 211: Please use `useSuspenseQuery` to automatically show a loading spinner until it has loaded

packages/widgets/src/docker/component.tsx
Meierschlumpf, line 1: When resizing the widget on the board it is quite laggy for me. Can you check if this still happens after the improvements (with moving functions outside the widget component function)

packages/widgets/src/index.tsx
```diff
@@ -14,6 +14,7 @@ import * as clock from "./clock";
 import type { WidgetComponentProps } from "./definition";
 import * as dnsHoleControls from "./dns-hole/controls";
 import * as dnsHoleSummary from "./dns-hole/summary";
+import * as docker from "./docker";
```
Meierschlumpf, line 17: Rename import to `dockerContainers`

packages/widgets/src/docker/component.tsx
```diff
@@ -0,0 +1,311 @@
...
+        const { mutateAsync: restartContainer } = clientApi.docker.restartAll.useMutation();
+
+        const handleActionAsync = async (action: "start" | "stop" | "restart") => {
+          const mutation = action === "start" ? startContainer : action === "stop" ? stopContainer : restartContainer;
+
+          await mutation(
+            { ids: [row.original.id] },
+            {
+              async onSettled() {
+                await utils.docker.getContainers.invalidate();
```
Meierschlumpf, line 161: Please invalidate the correct query

packages/request-handler/src/docker.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export const dockerContainersRequestHandler = createCachedWidgetRequestHandler({
+  queryKey: "dockerContainersResult",
+  widgetKind: "dockerContainers",
+  async requestAsync() {
+    const containers = await getContainersWithStats();
+
+    return responseSchema.parse({ containers });
+  },
+  cacheDuration: dayjs.duration(5, "seconds"),
```
Meierschlumpf, line 27: Change to something like `20 seconds`

Meierschlumpf (changes requested):
Also there should be a background job that periodically updates the data (like every minute) and we need to decide who should be able to add this widget to a board. The interactions should be restricted to only admins I would guess.

packages/widgets/src/docker/component.tsx
hillaliy: I think it's better now

Meierschlumpf:
Sadly we had to revert my restriction logic because of an infinite dependency import loop which also caused the dev server to crash / not run properly. I'll need to find another way to implement this. Sorry for it taking so long

hillaliy:
> Sadly we had to revert my restriction logic because of an infinite dependency import loop which also caused the dev server to crash / not run properly. I'll need to find another way to implement this. Sorry for it taking so long

It's OK if I can help with something just tell.

manuel-rw:
What's the status on this PR? Can we continue?

Meierschlumpf:
Okay @manuel-rw had a good idea that we just use the trpc response and if it is a tRPC-Error with Forbidden we could always just show the restricted widget text (so using suspense that will throw the error where we also have the NoIntegrationSelectedError)

Meierschlumpf:
@manuel-rw can you take a look and review this PR? It should work as intended

manuel-rw (approved):
Very nice, thank you!

PR #2335, opened 2025-02-15
feat: Scaling unification and betterment PoC

SeDemal:
https://github.com/user-attachments/assets/3b33de71-5fef-44d5-a28f-d742ce6790dc

Currently only works on the Clock widget to test the waters.

### DIFF
--- a/apps/nextjs/src/components/board/items/item-content.tsx
+++ b/apps/nextjs/src/components/board/items/item-content.tsx
@@ -35,8 +35,9 @@ export const BoardItemContent = ({ item }: BoardItemContentProps) => {
       styles={{
         root: {
           "--opacity": board.opacity / 100,
-          containerType: "size",
           overflow: item.kind === "iframe" ? "hidden" : undefined,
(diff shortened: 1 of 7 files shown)


Meierschlumpf:
When I understand you concept correctly, your plan is to remove any cqmin that we used, by default use the mantine sizes and scale them with this advanced options?

SeDemal:
> When I understand you concept correctly, your plan is to remove any cqmin that we used, by default use the mantine sizes and scale them with this advanced options?

The plan is to make the cqmin the base of the mantine sizes, so anyone can use it, and the "complex" equation should make it easy to modify in CSS so users have full control.

It's also the plan to be able to toggle cqmin/rem using this template.

The clock widget shows how easy you can implement cqmin, without ever typing the word, and only using premade "xs | sm | md | lg | xl".

Then the scaling in advanced settings helps modifying the size without ever typing it in CSS even.

I truly believe that most people's complaints are about the font being too big and not being able to control *how* it scales, not that it does scale up.

PR #2380, opened 2025-02-20
feat: add app ping url

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

image

### DIFF
--- a/packages/old-import/src/import-apps.ts
+++ b/packages/old-import/src/import-apps.ts
@@ -120,6 +120,7 @@ const addMappingFor = <TApp extends OldmarrApp | BookmarkApp>(
         href: existing.href,
         iconUrl: existing.iconUrl,
         description: existing.description,
+        pingUrl: existing.pingUrl,
         exists: true,
       });
       continue;
@@ -144,6 +145,7 @@ const convertApp = (app: OldmarrApp): DbAppWithoutId => ({
   href: app.behaviour.externalUrl === "" ? app.url : app.behaviour.externalUrl,
   iconUrl: app.appearance.iconUrl,
   description: app.behaviour.tooltipDescription ?? null,
+  pingUrl: app.behaviour.externalUrl.length > 0 ? app.behaviour.externalUrl : null,
 });
 
 /**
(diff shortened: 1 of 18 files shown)


packages/old-import/src/import-apps.ts
```diff
@@ -144,6 +145,7 @@ const convertApp = (app: OldmarrApp): DbAppWithoutId => ({
   href: app.behaviour.externalUrl === "" ? app.url : app.behaviour.externalUrl,
   iconUrl: app.appearance.iconUrl,
   description: app.behaviour.tooltipDescription ?? null,
+  pingUrl: app.behaviour.externalUrl.length > 0 ? app.behaviour.externalUrl : null,
```
Meierschlumpf, line 148: What about using `url` here? So the internal one

packages/old-import/src/mappers/map-app.ts
```diff
@@ -11,6 +11,7 @@ export const mapOldmarrApp = (app: OldmarrApp): InferSelectModel<typeof apps> =>
     iconUrl: app.appearance.iconUrl,
     description: app.behaviour.tooltipDescription ?? null,
     href: app.behaviour.externalUrl || app.url,
+    pingUrl: app.behaviour.externalUrl.length > 0 ? app.behaviour.externalUrl : null,
```
Meierschlumpf, line 14: What about using url here? So the internal one

packages/translation/src/lang/en.json
```diff
@@ -599,7 +599,17 @@
         "label": "Description"
       },
       "url": {
-        "label": "Url"
+        "label": "Url",
+        "placeholder": "https://...."
```
Meierschlumpf, line 603: I wouldn't put this in translations, because it will look exactly the same for all languages (Same for the ping input)

Meierschlumpf (changes requested):

Meierschlumpf:
If you agree with my points and are able to fix them until todays release, you can merge this PR if you want. If you don't agree then we can discuss further after it and release it next week

PR #2448, opened 2025-02-28
feat(pihole): add support for v6

### LINKED ISSUE #2367: feat: upgrade pi-hole integration to v6

Pi-hole released v6 and introduced breaking changes. https://pi-hole.net/blog/2025/02/18/introducing-pi-hole-v6/#page-content
They now have a real rest api, which also means they broke the integration that now needs to be implemented with the new api.

Path for api changed from `/admin/api.php` to `/api`

### Additional information

Also update the pi-hole test to latest

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Closes #2367


packages/integrations/src/base/creator.ts
```diff
@@ -35,15 +35,21 @@ export const integrationCreator = <TKind extends keyof typeof integrationCreator
     );
   }
 
-  return new integrationCreators[integration.kind](integration) as InstanceType<(typeof integrationCreators)[TKind]>;
+  const creator = integrationCreators[integration.kind];
+
+  if (Array.isArray(creator)) {
+    return (await creator[0](integration)) as IntegrationInstanceOfKind<TKind>;
+  }
```
manuel-rw, line 42: Document why ``creators`` can be an array and why we do this

packages/integrations/src/base/creator.ts
```diff
@@ -35,15 +35,21 @@ export const integrationCreator = <TKind extends keyof typeof integrationCreator
     );
   }
 
-  return new integrationCreators[integration.kind](integration) as InstanceType<(typeof integrationCreators)[TKind]>;
+  const creator = integrationCreators[integration.kind];
+
+  if (Array.isArray(creator)) {
+    return (await creator[0](integration)) as IntegrationInstanceOfKind<TKind>;
```
manuel-rw, line 41: Do we always have exactly one ``func``? Is the array construct simply a way of checking whether it's a function that can be executed?

packages/integrations/src/base/error.ts
```diff
@@ -0,0 +1,47 @@
...
+  constructor(response: Response | UndiciResponse, content: unknown) {
+    super("Response failed");
+
+    this.statusCode = response.status;
+    this.url = response.url;
+
+    try {
+      this.content = JSON.stringify(content);
+    } catch {
+      // ignore
```
manuel-rw, line 31: Why ignore? Consider logging as debug then?

packages/integrations/src/pi-hole/pi-hole-integration-v6.ts
```diff
@@ -0,0 +1,230 @@
...
+const statsSummaryGetSchema = z.object({
+  queries: z.object({
+    total: z.number(),
+    blocked: z.number(),
+    percent_blocked: z.number(),
+  }),
+  gravity: z.object({
+    domains_being_blocked: z.number(),
+  }),
+});
```
manuel-rw, line 44: Can't we move v5 and v6 into directories instead and create separate files for their corresponding schemas?

packages/integrations/src/pi-hole/pi-hole-integration-v6.ts
```diff
@@ -0,0 +1,230 @@
...
+    });
+
+    if (!response.ok) {
+      throw new IntegrationResponseError(this.integration, response, await response.json());
+    }
+
+    const result = dnsBlockingGetSchema.safeParse(await response.json());
+
+    if (!result.success) {
+      throw new ParseError("dns blocking status", result.error, await response.json());
```
manuel-rw, line 70:
```suggestion
      throw new ParseError("DNS blocking status", result.error, await response.json());
```

manuel-rw (changes requested):

packages/integrations/src/base/creator.ts
Meierschlumpf: Yes we have exactly one function, it is already described in a comment above the `integrationCreators` variable that: 
`factories are an array, to differentiate in js between class constructors and functions`

I moved this comment up as well

packages/integrations/src/base/error.ts
Meierschlumpf: I decided to just make it a string then

packages/integrations/src/base/creator.ts
manuel-rw: And [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#description) (e.g. ``typeof a == 'function'``) doesn't work? Please do not resolve this yet, I do not agree on this implementation

Meierschlumpf: no, a constructor is also  a function

manuel-rw: Then can't we infer this somehow differently?
I really dislike this approach because...
- it is not guaranteed that the array contains exactly one element
- you are abusing types to perform different business logic. We should infer differently
- ``class`` and ``function`` overlap poorly in my opinion and should not be used in one type here.
- The name ``creator`` became misleading
- ``creator`` implies synchronous workloads. Making it both synchronous (e.g. ``class``) and asynchronous (``Promise<T>``) goes against the naming
- the above explanation is not readable from the code. It is confusing to see index 0 being executed as a function

I do not agree on this solution for the above mentioned reasons. Please let me know if you disagree and explain your reasons. We can discuss this in the dev team if we cannot come to a conclusion ourselves.

Meierschlumpf: Okay I agree that the solution is quite hacky on the implementation side, but regarding the definition site it is very easy to use. If you have a better idea on how to do it, let me know. Here are the responses for your arguments:

1. is guaranteed because of typescript, see `satisfies Record<IntegrationKind, IntegrationInstance | [(input: IntegrationInput) => Promise<Integration>]>`
2. Both of the times it has to create the required integration. So I would consider this not different Business logic. It's true that the current solution is hacky, but I did not find a better simple solution.
3. What would you suggest otherwise? I think in general it is quite a simple to understand definition. For sure the implementation to run the correct logic from a class / function is complicated and could need some more comments.
4. Why do you think this? It still creates the integration class based on the integration-input, so IMO it is still a creator.
5. Do you mean `creator` in this context based on the file name? I would think the name makes sense. Also the function has a decent name.
6. I agree with this, I could add some more variables to make it more understandable / add comments

manuel-rw: We decided that we refactor this in a later revision to not block this fix

manuel-rw (approved):

packages/integrations/src/base/creator.ts
Meierschlumpf: https://github.com/homarr-labs/homarr/issues/2500

PR #2575, opened 2025-03-11
feat: Update Validation for Apps

frankvHoof93:
Allow other protocols (such as ftp://, steam://, rdp://) whilst still blocking javascript.

<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

- [X] Builds without warnings or errors (``pnpm build``, autofix with ``pnpm format:fix``)
- [X] Pull request targets ``dev`` branch
- [X] No shorthand variable names are used (eg. ``x``, ``y``, ``i`` or any abbrevation)

### DIFF
--- a/packages/validation/src/app.ts
+++ b/packages/validation/src/app.ts
@@ -13,7 +13,7 @@ const manageAppSchema = z.object({
     .string()
     .trim()
     .url()
-    .regex(/^https?:\/\//) // Only allow http and https for security reasons (javascript: is not allowed)
+    .regex(/^(?!javascript)[a-zA-Z]*:\/\//) // javascript: is not allowed
     .or(z.literal(""))
     .transform((value) => (value.length === 0 ? null : value))
     .nullable(),
@@ -21,7 +21,7 @@ const manageAppSchema = z.object({
     .string()
     .trim()
     .url()
-    .regex(/^https?:\/\//) // Only allow http and https for security reasons (javascript: is not allowed)
+    .regex(/^(?!javascript)[a-zA-Z]*:\/\//) // javascript: is not allowed
     .or(z.literal(""))
     .transform((value) => (value.length === 0 ? null : value))
     .nullable(),
--- a/packages/validation/package.json
+++ b/packages/validation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@homarr/validation",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "private": true,
   "license": "Apache-2.0",
   "type": "module",


packages/validation/src/app.ts
```diff
@@ -13,15 +13,15 @@ const manageAppSchema = z.object({
     .string()
     .trim()
     .url()
-    .regex(/^https?:\/\//) // Only allow http and https for security reasons (javascript: is not allowed)
+    .regex(/^(?!javascript)[a-zA-Z]*:\/\//) // javascript: is not allowed
```
Meierschlumpf, line 16: I'm not that happy with this because I think there will be different ways to run javascript beside this. So I would like to have a whitelist instead of a blacklist, also when this means, that some protocols will not be supported until added

Meierschlumpf: For example the following would also run js:
`data://text/html,%3Cscript%3Ealert('hi');%3C/script%3E`

frankvHoof93: Whitelist completely invalidates the usecase, unless the user can edit that whitelist.
For instance, I've personally set up 'kodi://' to start Kodi from my browser.
That's a 'protocol' that's not generally known or accepted.

I don't really understand the point of blocking all of this in the first place, since it's user-added and basically runs an href.
If the user didn't want to run JS, why would they directly enter in JS there in the first place?
If it's an attack vector, then the page opened by the http(s)-link is likely far more subject to being used than the link itself.

frankvHoof93: @manuel-rw 
>Would you be willing to implement this, @frankvHoof93 ?
Ehm. I'm terrible at TypeScript, and I'm mostly a Unity/.NET-dev 😅 
So I'm not sure if I'm best suited to add this UI. 👼

manuel-rw: No worries, I'm a .NET developer too.
We currently are very busy fixing other high priority tickets in the backlog and I don't this this will have a big value to most of our users.
Therefore, you can try and give it a shot if you're interested for a challenge or you can wait for someone else to contribute and pick this up.
Thanks for your understanding

StancuFlorin: @manuel-rw - I don't think you need a settings page for this. An environment variable should be enough. If the environment variable is missing, the default can be `/^https?:\/\//`

manuel-rw: I agree, that'd be enough. Is someone willing to pick up this change? I don't have any capacity at the moment

StancuFlorin: I will open a PR now

ajnart:
We already saw that users will always find new ways to use homarrr, having this restriction to only HTTP in the first place is restrictive, we could just define proper CSP rules so that it shows "VM1705:1 Refused to run the JavaScript URL because it violates the following Content Security Policy" error.

Either way we are **not** responsible for anyone's use of homarr, i'd remove the validation and expect my users to be smart enough to understand that a link is meant to be used there

manuel-rw:
I made some research and believe that CSP should be enough to protect the user. I will update the PR...

Meierschlumpf (approved):

manuel-rw:
@Meierschlumpf can you confirm CSP didn't break your setup?

Meierschlumpf:
I haven't tried with iframes / similar as I don't really have a setup myself. But I've tried to get around our javascript restrictions and fixed the relevant things. If you want I can also try out with iframes / images

PR #2579, opened 2025-03-12
feat: stock price widget

### LINKED ISSUE #2528: feat: Adding stock price integration

Not necessary the real-time stock price. But somewhat updated stock price would be nice/useful.
User can input their own api key for different services. 

### Additional information

_No response_

jackwellerreal:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

image

Closes #2528


packages/api/src/router/widgets/stocks.ts
```diff
@@ -0,0 +1,22 @@
+import { z } from "zod";
+
+import { fetchStockPriceHandler } from "@homarr/request-handler/stock-price";
+
+import { createTRPCRouter, publicProcedure } from "../../trpc";
+
+const stockPriceInputSchema = z.object({
+  stock: z.string().nonempty(),
+  timeRange: z.string().nonempty(),
+  timeInterval: z.string().nonempty(),
```
manuel-rw, line 10: Since we only offer pre-defined values for these properties, we could likely create a type for this, no?
Example:
```ts
z.enum(["7d", "2w", "1M", "6M", "1y"])
```

packages/request-handler/src/stock-price.ts
```diff
@@ -0,0 +1,20 @@
...
+import { fetchWithTimeout } from "@homarr/common";
+
+import { createCachedWidgetRequestHandler } from "./lib/cached-widget-request-handler";
+
+export const fetchStockPriceHandler = createCachedWidgetRequestHandler({
+  queryKey: "fetchStockPriceResult",
+  widgetKind: "stockPrice",
+  async requestAsync(input: { stock: string; timeRange: string; timeInterval: string }) {
+    const response = await fetchWithTimeout(
+      `https://query1.finance.yahoo.com/v8/finance/chart/${input.stock}?range=${input.timeRange}&interval=${input.timeInterval}`,
```
manuel-rw, line 13: Do we know what the rate limit on that API is?

packages/request-handler/src/stock-price.ts
```diff
@@ -0,0 +1,20 @@
...
+    );
+    return responseSchema.parse(await response.json()) as object;
```
manuel-rw, line 15: Why is the ``as object`` nessesary here?

packages/widgets/src/stocks/component.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { useScopedI18n } from "@homarr/translation/client";
+
+import type { WidgetComponentProps } from "../definition";
+
+export default function StockPriceWidget({ options, width, height }: WidgetComponentProps<"stockPrice">) {
+  const t = useScopedI18n("widget.stockPrice");
+  const theme = useMantineTheme();
+  const [{ data }] = clientApi.widget.stockPrice.getPrices.useSuspenseQuery(options);
+
+  const stockData = data.chart.result[0];
```
manuel-rw, line 17: Will there always be **exactly** one result?

packages/widgets/src/stocks/component.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const stockValues = Object.values(stockData.indicators.quote[0].close);
+  const stockValuesChange = (
+    100 *
+    ((stockValues[stockValues.length - 1] - stockValues[0]) / stockValues[stockValues.length - 1])
+  ).toFixed(2) as int;
```
manuel-rw, line 23: This is not true. It returns a string: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed
Do not cast to ``int`` or ``number``

packages/widgets/src/stocks/component.tsx
```diff
@@ -0,0 +1,75 @@
...
+        {stockValues[stockValues.length - 1].toFixed(2)}
+      </Title>
+
+      <Stack pos="absolute" top={10} left={10}>
+        <Text size="xl" fw={700} lh="0.715">
+          {stockValuesChange > 0 ? (
+            <IconTrendingUp size="1.5rem" color={theme.colors.green[7]} />
+          ) : (
+            <IconTrendingDown size="1.5rem" color={theme.colors.red[7]} />
+          )}{" "}
```
manuel-rw, line 64:
```suggestion
          )}
```

packages/widgets/src/stocks/index.ts
```diff
@@ -0,0 +1,44 @@
...
+        options: [
+          { value: "1d", label: "1 day" },
+          { value: "5d", label: "5 days" },
+          { value: "1mo", label: "1 month" },
+          { value: "6mo", label: "6 months" },
+          { value: "ytd", label: "Year to date" },
+          { value: "1y", label: "1 year" },
+          { value: "5y", label: "5 years" },
+          { value: "10y", label: "10 years" },
+          { value: "max", label: "Max" },
```
manuel-rw, line 24: You already added translations for this, no?
Can't check myself but don't we reference those translations in the other widgets?

manuel-rw (changes requested):

packages/widgets/src/stocks/component.tsx
jackwellerreal: From all the examples I've made, it has only responded with one result.

packages/request-handler/src/stock-price.ts
jackwellerreal: I'm not sure, you would have to ask @Meierschlumpf since he recommended Yahoo API to be used

packages/widgets/src/stocks/index.ts
jackwellerreal: Sorry, I do not understand the question. Do you mean that the translations don't need to be included in `en.json` because they are written in the labels?

packages/request-handler/src/stock-price.ts
Meierschlumpf: I saw one post where they said 5000 requests a day

packages/widgets/src/stocks/component.tsx
manuel-rw: Then please update the trpc procedure and only return the object

packages/widgets/src/stocks/index.ts
manuel-rw: Use the ``t`` translation function here instead.

Meierschlumpf:
I really like the look of this widget, good work!

packages/widgets/src/stocks/index.ts
```diff
@@ -0,0 +1,32 @@
+import { IconBuildingBank } from "@tabler/icons-react";
+
+import { createWidgetDefinition } from "../definition";
+import { optionsBuilder } from "../options";
+
+const timeRangeOptions = ["1d", "5d", "1mo", "6mo", "ytd", "1y", "5y", "10y", "max"] as const;
+const timeIntervalOptions = ["1m", "5m", "15m", "30m", "1h", "6h", "1d", "5d", "1wk", "1mo"] as const;
```
manuel-rw, line 7: Nitpick: Perhaps you could infer this from the zod schema?

manuel-rw (approved):
Please wait for Meier's approval

packages/widgets/src/stocks/index.ts
jackwellerreal: After importing the zod schema, I tried doing something like this but it fails to run and build
```ts
import { stockPriceInputSchema } from "../../../api/src/router/widgets/stocks";

const timeRangeOptions = stockPriceInputSchema.shape.timeRange.Enum;
const timeIntervalOptions = stockPriceInputSchema.shape.timeInterval.Enum;
```

manuel-rw (approved):

packages/api/src/router/widgets/stocks.ts
```diff
@@ -0,0 +1,23 @@
+import { z } from "zod";
+
+import { fetchStockPriceHandler } from "@homarr/request-handler/stock-price";
+
+import { stockPriceTimeFrames } from "../../../../widgets/src/stocks";
+import { createTRPCRouter, publicProcedure } from "../../trpc";
+
+const stockPriceInputSchema = z.object({
+  stock: z.string().nonempty(),
+  timeRange: z.enum([...stockPriceTimeFrames.range]),
```
Meierschlumpf, line 10: Does it not work with `z.enum(stockPriceTimeFrames.range)`?
If it works, fix interval as well

packages/api/src/router/widgets/stocks.ts
```diff
@@ -0,0 +1,23 @@
...
+});
+
+export const stockPriceRouter = createTRPCRouter({
+  getPrices: publicProcedure.input(stockPriceInputSchema).query(async ({ input }) => {
+    const innerHandler = fetchStockPriceHandler.handler({
+      stock: input.stock,
+      timeRange: input.timeRange,
+      timeInterval: input.timeInterval,
+    });
+    return await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: true });
```
Meierschlumpf, line 21: We should not force update here as this will result in a new call, even if there is a cached non stale value

packages/request-handler/src/stock-price.ts
```diff
@@ -0,0 +1,23 @@
...
+    const response = await fetchWithTimeout(
+      `https://query1.finance.yahoo.com/v8/finance/chart/${input.stock}?range=${input.timeRange}&interval=${input.timeInterval}`,
+    );
+    const data = await response.json();
+
+    if (data.error) {
+      throw new Error(data.error.description);
+    }
+
+    return data.chart.result[0];
```
Meierschlumpf, line 20: Isn't data any here? If so please use a zod schema to valdidate

packages/api/src/router/widgets/stocks.ts
```diff
@@ -0,0 +1,23 @@
...
+import { createTRPCRouter, publicProcedure } from "../../trpc";
+
+const stockPriceInputSchema = z.object({
+  stock: z.string().nonempty(),
+  timeRange: z.enum([...stockPriceTimeFrames.range]),
+  timeInterval: z.enum([...stockPriceTimeFrames.interval]),
+});
+
+export const stockPriceRouter = createTRPCRouter({
+  getPrices: publicProcedure.input(stockPriceInputSchema).query(async ({ input }) => {
```
Meierschlumpf, line 15: I first did not understand why you used plural. Maybe we could rename it to `getPriceHistory`

Meierschlumpf (changes requested):

packages/request-handler/src/stock-price.ts
```diff
@@ -11,13 +12,27 @@ export const fetchStockPriceHandler = createCachedWidgetRequestHandler({
...
     );
-    const data = await response.json();
+    const data = await response.json() as object;
 
     if (data.error) {
       throw new Error(data.error.description);
     }
 
-    return data.chart.result[0];
+    return responseSchema.parse(data.chart.result[0]);
```
Meierschlumpf: I would suggest to parse `data` the `await response.json()` directly.
So you would probably need something in the direction of

```
z.object({
	chart: z.object({
		result: z.array(responseSchema)
	})
}).or(z.object({
	error: z.object({
		description: z.string()
	})
}))
```

Meierschlumpf (changes requested):

packages/request-handler/src/stock-price.ts
```diff
@@ -0,0 +1,52 @@
...
+    const response = await fetchWithTimeout(
+      `https://query1.finance.yahoo.com/v8/finance/chart/${input.stock}?range=${input.timeRange}&interval=${input.timeInterval}`,
+    );
+    const data = dataSchema.parse(await response.json());
+
+    if ("error" in data) {
+      throw new Error(data.error.description);
+    }
+
+    return responseSchema.parse(data.chart.result[0]);
```
Meierschlumpf, line 21: This parse is unnecessary, but you'll also have to make the array check to have a length of more than 1

packages/widgets/package.json
```diff
@@ -68,6 +69,7 @@
     "next": "15.1.7",
     "react": "19.0.0",
     "react-dom": "19.0.0",
+    "recharts": "2",
```
Meierschlumpf, line 72: I guess this should be `^2.15.1`

Meierschlumpf (changes requested):

manuel-rw (dismissed):

Meierschlumpf (dismissed):

jackwellerreal:
Whoops, sorry

manuel-rw (approved):

Meierschlumpf (approved):

Meierschlumpf:
Can you fix the typescript issues? Once done we can merge IMO

manuel-rw (dismissed):

Meierschlumpf (approved):

ajnart:
Congratulations @jackwellerreal !

PR #2663, opened 2025-03-21
feat: add default search engines seeding / set homarr docs as global server search engine

ajnart:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Added the following search engines as default during the seeding process. Feel free to directly edit the code if you'd want to use another one than Google / add more search engines. 
I also set Homarr docs as the default search engine 

image

### DIFF
--- a/packages/db/migrations/seed.ts
+++ b/packages/db/migrations/seed.ts
@@ -101,20 +101,11 @@ const seedDefaultSearchEnginesAsync = async (db: Database) => {
   // Set Homarr docs as the default search engine in server settings
   const searchSettings = await getServerSettingByKeyAsync(db, "search");
 
-  if (searchSettings) {
-    if (!searchSettings.defaultSearchEngineId) {
-      await updateServerSettingByKeyAsync(db, "search", {
-        ...searchSettings,
-        defaultSearchEngineId: homarrId,
-      });
-      console.log("Set Homarr docs as the default search engine");
-    }
-  } else {
-    await updateServerSettingByKeyAsync(db, "search", {
-      defaultSearchEngineId: homarrId,
-    });
-    console.log("Created search settings with Homarr docs as the default search engine");
-  }
+  await updateServerSettingByKeyAsync(db, "search", {
+    ...searchSettings,
+    defaultSearchEngineId: homarrId,
+  });
+  console.log("Set Homarr docs as the default search engine");
 };
 
 const seedServerSettingsAsync = async (db: Database) => {
--- a/packages/db/migrations/seed.ts
+++ b/packages/db/migrations/seed.ts
@@ -1,5 +1,5 @@
 import { objectKeys } from "@homarr/common";
-import { everyoneGroup } from "@homarr/definitions";
+import { createDocumentationLink, everyoneGroup } from "@homarr/definitions";
 import { defaultServerSettings, defaultServerSettingsKeys } from "@homarr/server-settings";
 
 import type { Database } from "..";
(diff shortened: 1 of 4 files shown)


packages/db/migrations/seed.ts
```diff
@@ -48,6 +49,77 @@ const seedOnboardingAsync = async (db: Database) => {
...
+      type: "generic" as const,
+      integrationId: null,
+    },
+  ];
+
+  await db.insert(searchEngines).values(defaultSearchEngines);
+  console.log(`Created ${defaultSearchEngines.length} default search engines through seeding process`);
+  
+  // Set Homarr docs as the default search engine in server settings
+  const searchSettings = await db.query.serverSettings.findFirst({
```
Meierschlumpf, line 98: There are methods under queries in the db package that already handle superjson parsing and have full typesafety

Meierschlumpf (changes requested):

packages/db/migrations/seed.ts
```diff
@@ -48,21 +48,82 @@ const seedOnboardingAsync = async (db: Database) => {
   console.log("Created onboarding step through seed");
 };
 
+const seedDefaultSearchEnginesAsync = async (db: Database) => {
+  const existingSearchEngines = await db.query.searchEngines.findMany();
```
Meierschlumpf, line 52: Use `db.$count(searchEngines)` instead

packages/db/migrations/seed.ts
```diff
@@ -48,21 +48,82 @@ const seedOnboardingAsync = async (db: Database) => {
...
   const serverSettingsData = await db.query.serverSettings.findMany();
 
   for (const settingsKey of defaultServerSettingsKeys) {
     const currentDbEntry = serverSettingsData.find((setting) => setting.settingKey === settingsKey);
     if (!currentDbEntry) {
-      await db.insert(serverSettings).values({
-        settingKey: settingsKey,
-        value: SuperJSON.stringify(defaultServerSettings[settingsKey]),
-      });
+      await updateServerSettingByKeyAsync(db, settingsKey, defaultServerSettings[settingsKey]);
```
Meierschlumpf, line 121: This does not work as it has to be inserted instead of updated

Meierschlumpf (changes requested):

packages/db/migrations/seed.ts
```diff
@@ -48,21 +52,82 @@ const seedOnboardingAsync = async (db: Database) => {
...
+        defaultSearchEngineId: homarrId,
+      });
+      console.log("Set Homarr docs as the default search engine");
+    }
+  } else {
+    await updateServerSettingByKeyAsync(db, "search", {
+      defaultSearchEngineId: homarrId,
+    });
+    console.log("Created search settings with Homarr docs as the default search engine");
+  }
```
Meierschlumpf, line 116: Remove this as the server-settings are always initialized before it and so this is not necessary (and would be wrong if we would like to create it)

Meierschlumpf (dismissed):
In general fix the failing CI and my comment. Then it is good to go

packages/db/migrations/seed.ts
Meierschlumpf: What I mean is, that the seeding of the server-settings happens first already, so we will never get into this else clause.
AND
if we would have needed this else clause, then the server settings should not be updated and rather inserted

ajnart: Thanks ! Much clearer now :D I've committed d289fe33c2cba3cabe092b10d8f07ee9e09be461 to address this

packages/db/migrations/seed.ts
```diff
@@ -48,21 +52,74 @@ const seedOnboardingAsync = async (db: Database) => {
...
+      integrationId: null,
+    },
+    {
+      id: homarrId,
+      name: "Homarr Docs",
+      iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons/svg/homarr.svg",
+      short: "docs",
+      description: "Search the Homarr documentation",
+      // eslint-disable-next-line no-restricted-syntax
+      urlTemplate: "https://homarr.dev/search?q=%s",
```
Meierschlumpf, line 92: What about using the recommended method and updating it to accept query params 😉

ajnart: You're right! I did that and added tests as well 👍🏼

Meierschlumpf (approved):
Thanks for the tests ❤️

PR #2664, opened 2025-03-21
feat: auto sign-in when creating the initial user

ajnart:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

What the title says, avoids a double sign-in by calling `await signIn` after successfully creating the first user


manuel-rw (approved):
lgtm, wait for @Meierschlumpf's review
Do not bypass the merge requirements, fix the format first.

Meierschlumpf (approved):
I made a quick research, because there are some concerns that stand against this convenience improvement. But all of them don't really apply for onboarding on a self hosted app. So IMO this is okay 👍🏼

manuel-rw (approved):

PR #2665, opened 2025-03-21
feat: Prefer primary color

ajnart:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Simple code mod to search and replace in the codebase in order to to remove `color` prop when it is set to `teal` so that it uses the primary color instead (from the user preference)

### DIFF
--- a/apps/nextjs/src/app/[locale]/boards/[name]/settings/_appereance.tsx
+++ b/apps/nextjs/src/app/[locale]/boards/[name]/settings/_appereance.tsx
@@ -9,11 +9,11 @@ import {
   Grid,
   Group,
   InputWrapper,
-  isLightColor,
   Select,
   Slider,
   Stack,
   Text,
+  isLightColor,
   useMantineTheme,
 } from "@mantine/core";
 import { useDisclosure } from "@mantine/hooks";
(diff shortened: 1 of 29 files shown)


manuel-rw:
Your PR is still not ready for review, the format is failing.

ajnart:
> Your PR is still not ready for review, the format is failing.

My bad, I didn't see it was the case. I just re-ran the formatting script and pushed another commit

Meierschlumpf:
IMO it would only make sense on the board pages, but not on the management pages, no? I mean there is no option to configure styles outside the board page

ajnart:
> IMO it would only make sense on the board pages, but not on the management pages, no? I mean there is no option to configure styles outside the board page

It just removes the hard-coded colors, the teal buttons with black text clash with the overall theme/consistency of the UI everywhere where it's used. Perhaps at some point we can add "server colors" similar to the user settings to set that color globally

manuel-rw:
The reason why we used teal is, because they are save / confirm buttons.
As a suggestion: Why not change the color to a more pleasing one instead of using ``primary``?

ajnart:
> The reason why we used teal is, because they are save / confirm buttons.
> 
> As a suggestion: Why not change the color to a more pleasing one instead of using ``primary``?

For now this PR is only to remove the inconsistencies (hence the name)
I believe there is a Mantine-wide config especially for the confirm/cancel buttons (primary/destructive?) 

I could do a follow up PR introducing new consistency changes and checking that every modal stays consistent with the defined theme

manuel-rw:
Please don't forget to squash. And next time, you should use the ``feat`` prefix, since this is not a refactor

PR #2679, opened 2025-03-23
feat: implement auto-select icon

### LINKED ISSUE #2678: feat: Automatic icon selection based on name

Like in oldmarr, auto-suggest / select an icon if its name matches 

### Additional information

_No response_

ajnart:
Uses the debounced icon name in order to auto-populate the icon in the "Add app" form. 

Doesn't re-apply if there is already an icon selected. 

Closes #2678

### DIFF
--- a/packages/forms-collection/src/new-app/_form.tsx
+++ b/packages/forms-collection/src/new-app/_form.tsx
@@ -13,6 +13,7 @@ import { useI18n } from "@homarr/translation/client";
 import { validation } from "@homarr/validation";
 
 import { IconPicker } from "../icon-picker/icon-picker";
+import { findBestIconMatch } from "./icon-matcher";
 
 type FormType = z.infer<typeof validation.app.manage>;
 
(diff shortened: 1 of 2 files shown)


packages/forms-collection/src/new-app/_form.tsx
```diff
@@ -68,6 +72,65 @@ export const AppForm = ({
...
+      if (exactSvgMatch) {
+        form.setFieldValue("iconUrl", exactSvgMatch);
+      } else if (exactNonSvgMatch) {
+        form.setFieldValue("iconUrl", exactNonSvgMatch);
+      } else if (partialSvgMatch) {
+        form.setFieldValue("iconUrl", partialSvgMatch);
+      } else if (partialNonSvgMatch) {
+        form.setFieldValue("iconUrl", partialNonSvgMatch);
+      }
+    }
```
manuel-rw, line 131: Extract this code

manuel-rw, line 131: I have a few issues with this code:
- The code is repetitive (e.g. ``setFieldValue``)
- The prioritization could be implemented with less code duplication
- We iterate over **all** icons, which will lead to poor performance. We should interrupt the loop when we find a match

manuel-rw (dismissed):

packages/forms-collection/src/new-app/_form.tsx
ajnart: You're right, I added an early exit when there is an exact match and removed all of the variables

ajnart: Did that to `icon-matcher.ts`. I'm not sure if it should be located somewhere else but I figured that it's not necessary to move it because it won't be re-used. 
I used the `inferRouterOutputs` to get the correct type too 😄

manuel-rw (approved):
Such a big improvement from iteration 1! Great work, @ajnart

PR #2681, opened 2025-03-23
feat: [proposal] quick add app to board

ajnart:
## Feature
Adds a "Add app" into the edit mode header directly, as a quick way to add an app to the dashboard. 
```diff
- Less generic than the "Widget" way of adding apps
- Might be confusing to add more items onto the header
- Introduces redundancy with the "wiget" way of adding an app to a dash
+ Easier/Quicker to use 
+ More visual way to quickly add an app
```

## Demo 

https://github.com/user-attachments/assets/eed1ca16-8309-48ac-8a64-7baf8e2b32e4

### DIFF
--- a/apps/nextjs/src/app/[locale]/boards/(content)/_header-actions.tsx
+++ b/apps/nextjs/src/app/[locale]/boards/(content)/_header-actions.tsx
@@ -7,7 +7,6 @@ import { useRouter } from "next/navigation";
 import { Group, Menu, ScrollArea } from "@mantine/core";
 import { useHotkeys } from "@mantine/hooks";
 import {
-  IconApps,
   IconBox,
   IconBoxAlignTop,
   IconChevronDown,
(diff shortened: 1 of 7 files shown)


packages/modals-collection/src/apps/app-select-modal.tsx
```diff
@@ -0,0 +1,122 @@
...
+        data-autofocus
+        onKeyDown={(event) => {
+          if (event.key === "Enter" && filteredApps.length === 1 && filteredApps[0]) {
+            handleSelect(filteredApps[0].id);
+          }
+        }}
+      />
+
+      <Grid>
+        {/* New App option */}
```
manuel-rw, line 64: Useless, remove comment

manuel-rw (approved):
Nice idea, I like the change. Please wait for at least one other approval

apps/nextjs/src/app/[locale]/boards/(content)/_header-actions.tsx
```diff
@@ -105,6 +121,10 @@ const AddMenu = () => {
           {t("item.action.create")}
         </Menu.Item>
 
+        <Menu.Item leftSection={<IconApps size={20} />} onClick={handleSelectApp}>
```
Comment (Meierschlumpf) [inline apps/nextjs/src/app/[locale]/boards/(content)/_header-actions.tsx:124]: I would suggest that we change the icon for items and use the same icon for apps as we have in management. There we use `IconBox` for apps. Maybe we can use `IconResize` or something similar for items

packages/modals-collection/src/apps/quick-add-app/quick-add-app-modal.tsx
```diff
@@ -8,7 +8,7 @@ import { useI18n, useScopedI18n } from "@homarr/translation/client";
 import type { validation } from "@homarr/validation";
 
 interface QuickAddAppModalProps {
-  onClose: (createdAppId: string) => Promise<void>;
+  onClose: (createdAppId: string) => void;
```
Meierschlumpf, line 11: I would suggest that we still allow a `Promise` here. You could use the type `MaybePromise` which allows both promise and no promise values

Meierschlumpf (changes requested):
I like this feature a lot, sorry for it taking so long until I reviewed it

Meierschlumpf (approved):

PR #2682, opened 2025-03-23
feat(middleware): implement onboarding step cache

ajnart:
This solution is very simplistic and does not rely on redis on purpose.
This is so that it saves another "round-trip" at every single request

This choice is because the middleware should be the fastest possible im my opinion 

## Decision taken 
- [ ] Keep using `let` to define a variable that lives in the middleware (kept alive by nextjs since we are not deploying our API to functions) 
- [ ] Switch to using redis to cache this value (increases round trip time but it should still be faster than trpc) 

## Tests 
Reduced the time between each page switch by about 45ms on my machine

### DIFF
--- a/apps/nextjs/src/middleware.ts
+++ b/apps/nextjs/src/middleware.ts
@@ -1,34 +1,48 @@
+import { createTRPCClient, httpLink } from "@trpc/client";
 import type { NextRequest } from "next/server";
 import { NextResponse } from "next/server";
-import { createTRPCClient, httpLink } from "@trpc/client";
 import SuperJSON from "superjson";
 
 import type { AppRouter } from "@homarr/api";
 import { createHeadersCallbackForSource, getTrpcUrl } from "@homarr/api/shared";
+import { SupportedLanguage } from "@homarr/translation";
 import { createI18nMiddleware } from "@homarr/translation/middleware";
 
 // Simple in-memory cache for onboarding status (makes it only request once and keep result in memory)
-let onboardingStatusCache: string | null = null;
+const onboardingStatusCache = new Map<string, string>();
+const cultureCache = new Map<string, { defaultLocale: string }>();
 
 export async function middleware(request: NextRequest) {
+  const csrfToken = request.cookies.get('authjs.csrf-token')?.value;
+  
   // fetch api does not work because window is not defined and we need to construct the url from the headers
   // In next 15 we will be able to use node apis and such the db directly
-  const culture = await serverFetchApi.serverSettings.getCulture.query();
+  let culture;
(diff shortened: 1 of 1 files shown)


manuel-rw (approved):
lgtm, please wait for @Meierschlumpf since he has more experience with Next.js.

>  Switch to using redis to cache this value (increases round trip time but it should still be faster than trpc

I think the roundtrip time should be pretty similar, no?

ajnart:
> >  Switch to using redis to cache this value (increases round trip time but it should still be faster than trpc
> 
> 
> 
> I think the roundtrip time should be pretty similar, no?

I believe that redis cache interactions are near instantaneous since the connection is kept alive whereas trpc queries the DB again

ajnart:
@manuel-rw [b7ba3c6](https://github.com/homarr-labs/homarr/pull/2682/commits/b7ba3c6e1f94b9841af662316de42dfe0fd5ea16) is an optional commit that tries to also cache the "Culture" attribute for a user. When making these changes in dev mode I saw a very significant change in the speed taken to switch pages. From ~450ms to < 50ms with the cached values. I believe this will also enhance prod significantly

manuel-rw:
lgtm, please wait for Meier

Meierschlumpf:
I would suggest that we use the new experimental feature for node-middleware once it is fully available. Because then we can directly query the db and don't need to use fetch for it.

Meierschlumpf:
Then we'll probably also don't need a cache

ajnart:
> I would suggest that we use the new experimental feature for node-middleware. Because then we can directly query the db and don't need to use fetch for it.
> 
> For this you'll need 15.2+ which is available once #2701 is merged
> 
> [nextjs.org/docs/app/building-your-application/routing/middleware#runtime](https://nextjs.org/docs/app/building-your-application/routing/middleware#runtime)

I don't understand how that will fix this.. Currently our middleware is running on node, since we deploy as a standalone package we are never using an edge server. Querying the DB will still have a similar latency ± 100ms. We could however use an existing redis singleton connection to reach into redis directly 
The main point is that this value doesn't need to be stored but instead cached

Meierschlumpf:
No middleware did not support node packages, so it will have quite an increase. Of course we can discuss if we want to cache it further once we tried the node middleware

Meierschlumpf:
Node middleware is currently only available in canary

Meierschlumpf:
Storing it with the csrf token will result in the stale cache still existing as long as you don't login / logout again. So not really a good way to cache it. I would consider waiting for nodeMiddleware and using `unstable_cache` to cache it with a tag which we can use to invalidate the cache. Because otherwise it will just return the wrong data. We could however add a custom env variable or something to cache it forever locally so we have better performance

Issue #2680, opened 2025-03-23
feat(UX): auto set home board when creating a first board


ajnart:
### Describe the feature you'd like to request

If there are no server-wide home board, the user's home board should be defined as the newly created board 

- We could send a toast to tell the user about it (or not, I think this behavior makes sense)

PR #2683, opened 2025-03-23
feat: set home board when creating first board

ajnart:
Closes https://github.com/homarr-labs/homarr/issues/2680 

Auto-assigns a home board to a user when they create their first board

### DIFF
--- a/apps/nextjs/src/app/[locale]/manage/boards/_components/create-board-button.tsx
+++ b/apps/nextjs/src/app/[locale]/manage/boards/_components/create-board-button.tsx
@@ -3,18 +3,39 @@
 import { Affix, Button, Menu } from "@mantine/core";
 import { IconCategoryPlus, IconChevronDown, IconFileImport } from "@tabler/icons-react";
 
+import { clientApi } from "@homarr/api/client";
+import { revalidatePathActionAsync } from "@homarr/common/client";
 import { useModalAction } from "@homarr/modals";
 import { AddBoardModal, ImportBoardModal } from "@homarr/modals-collection";
 import { useI18n } from "@homarr/translation/client";
 
 export const CreateBoardButton = () => {
   const t = useI18n();
+  const { data: boards } = clientApi.board.getAllBoards.useQuery();
+  const { mutate: setHomeBoard } = clientApi.board.setHomeBoard.useMutation({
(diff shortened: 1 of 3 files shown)


apps/nextjs/src/app/[locale]/manage/boards/_components/create-board-button.tsx
Comment (manuel-rw) [inline apps/nextjs/src/app/[locale]/manage/boards/_components/create-board-button.tsx:1]: Code isn't wrong, but I would prefer doing this on the server side, to reduce round trips and point of failures. The second request could easily fail and we could run into race conditions here.
Please change the implementation accordingly

manuel-rw (dismissed):

apps/nextjs/src/app/[locale]/manage/boards/_components/create-board-button.tsx
Comment (ajnart) [reply apps/nextjs/src/app/[locale]/manage/boards/_components/create-board-button.tsx:1]: You're right, I've just made the change in 922528a

packages/api/src/router/board.ts
```diff
@@ -1081,8 +1094,7 @@ export const boardRouter = createTRPCRouter({
                 options: section.kind === "dynamic" ? superjson.stringify(section.options) : emptySuperJSON,
                 name: prev?.kind === "category" && "name" in section ? section.name : null,
               })
-              .where(eq(sections.id, section.id))
-              .run();
+              .where(eq(sections.id, section.id));
```
Meierschlumpf, line 1097: .run has to stay. it is relevant for the transaction

packages/api/src/router/board.ts
```diff
@@ -1098,8 +1110,7 @@ export const boardRouter = createTRPCRouter({
                 })
                 .where(
                   and(eq(sectionLayouts.sectionId, section.id), eq(sectionLayouts.layoutId, sectionLayout.layoutId)),
-                )
-                .run();
+                );
```
Meierschlumpf, line 1113: .run has to stay. it is relevant for the transaction

Meierschlumpf (changes requested):

packages/api/src/router/board.ts
ajnart: I just added it back. I'm not sure how this got to be removed in the first place

Meierschlumpf (approved):

PR #2686, opened 2025-03-23
feat: add widget feedback feature

ajnart:
Add a "give feedback" on every widget when in edit mode. 

Also exposes a ummai client that allows to send umami data through the client


packages/modals-collection/src/feedback/widget-feedback-modal.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+export const WidgetFeedbackModal = createModal<WidgetFeedbackModalProps>(({ actions, innerProps }) => {
+  const t = useI18n();
+  const tFeedback = useScopedI18n("feedback");
+  const [usabilityRating, setUsabilityRating] = useState(0);
+  const [featuresRating, setFeaturesRating] = useState(0);
+  const [overallRating, setOverallRating] = useState(0);
+  const [discordUsername, setDiscordUsername] = useState("");
+  const [feedbackText, setFeedbackText] = useState("");
+  const [isSubmitting, setIsSubmitting] = useState(false);
```
manuel-rw, line 23: Use https://mantine.dev/hooks/use-disclosure/?

apps/nextjs/src/components/board/items/item-menu.tsx
```diff
@@ -115,6 +131,9 @@ export const BoardItemMenu = ({
         <Menu.Item leftSection={<IconCopy size={16} />} onClick={() => duplicateItem({ itemId: item.id })}>
           {tItem("action.duplicate")}
         </Menu.Item>
+        <Menu.Item leftSection={<IconMessageCircle2 size={16} />} onClick={openWidgetFeedbackModal}>
+          {tItem("action.feedback")}
+        </Menu.Item>
```
manuel-rw, line 136: Render this conditionally, only when analytics are enabled

manuel-rw (dismissed):

apps/nextjs/src/components/board/items/item-menu.tsx
ajnart: Technically this is not analytics. We are just leveraging our umami endpoint to send feedback. It won't send anything else other than that form

manuel-rw: Hmm... in theory you're right, but we disable umami when you disabled analytics.
For the sake of transparency and integrity, I think we should keep doing it that way, hence disabling the feedback menu item.

packages/modals-collection/src/feedback/widget-feedback-modal.tsx
```diff
@@ -0,0 +1,109 @@
...
+  widgetKind: string;
+}
+
+export const WidgetFeedbackModal = createModal<WidgetFeedbackModalProps>(({ actions, innerProps }) => {
+  const t = useI18n();
+  const tFeedback = useScopedI18n("feedback");
+  const [usabilityRating, setUsabilityRating] = useState(0);
+  const [featuresRating, setFeaturesRating] = useState(0);
+  const [overallRating, setOverallRating] = useState(0);
+  const [discordUsername, setDiscordUsername] = useState("");
```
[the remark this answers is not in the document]
ajnart: discordUsername is if we want to follow up with the person to ask them some questions or keep them updated if we add their suggestions

apps/nextjs/src/components/board/items/item-menu.tsx
ajnart: @Meierschlumpf @SeDemal what's your opinion on this ?

packages/modals-collection/src/feedback/widget-feedback-modal.tsx
```diff
@@ -0,0 +1,110 @@
...
+      }
+
+      showSuccessNotification({
+        title: t("notification.success.title"),
+        message: t("notification.success.message"),
+      });
+
+      actions.closeModal();
+    } catch (error) {
+      console.error("Error sending feedback:", error);
```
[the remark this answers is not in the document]
ajnart: Actually not a bad idea

PR #2700, opened 2025-03-26
chore: add development package and update workspace

ajnart:
The aim of this PR is to make it easier to run the dev environment 

- Introduced a new development package with a Docker Compose script.
- Updated pnpm workspace to include the development directory.
- Changed Turbo UI setting from 'stream' to 'tui'.

## Preview
image

### DIFF
--- a/development/package.json
+++ b/development/package.json
@@ -1,0 +1,10 @@
+{
+  "name": "@homarr/dev-db",
+  "version": "0.1.0",
+  "private": true,
+  "license": "Apache-2.0",
+  "type": "module",
+  "scripts": {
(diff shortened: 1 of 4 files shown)


development/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@homarr/dev-db",
```
manuel-rw, line 2: Name is misleading, the compose file *could* contain things besides a DB

manuel-rw (changes requested):

development/package.json
ajnart: Fixed it ;)

Meierschlumpf (approved):

PR #2836, opened 2025-04-09
fix(widget): format stock values and changes for better readability

m0s4ik:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


.env.example
manuel-rw, line 1: Please revert this change

packages/widgets/src/stocks/component.tsx
```diff
@@ -57,7 +57,7 @@ export default function StockPriceWidget({ options, width, height }: WidgetCompo
           ) : (
             <IconTrendingDown size="1.5rem" color={theme.colors.red[7]} />
           )}
-          {data.meta.symbol}
+           {data.meta.symbol}
```
manuel-rw, line 60: Revert, not required

manuel-rw (changes requested):
Can you post a screenshot of how this formats it differently?

m0s4ik:
> Can you post a screenshot of how this formats it differently?

Stock
I have added number formatting with dots and commas. In this case, use the browser's default format.The larger one is the new format.

manuel-rw:
Thanks. My previous comment stills stands, please run the auto prettier format

manuel-rw (approved):

packages/widgets/src/stocks/component.tsx
```diff
@@ -67,12 +67,13 @@ export default function StockPriceWidget({ options, width, height }: WidgetCompo
...
       <Title pos="absolute" bottom={10} right={10} order={width > 280 ? 1 : 2} fw={700}>
-        {round(stockValues[stockValues.length - 1] ?? 0)}
+        {new Intl.NumberFormat().format(round(stockValues[stockValues.length - 1] ?? 0))}
       </Title>
 
       {width > 280 && (
         <Text pos="absolute" top={10} right={10} size="xl" fw={700}>
-          {Math.abs(stockValuesChange)} ({Math.abs(stockValuesChangePercentage)}%)
+          {new Intl.NumberFormat().format(stockValuesChange)} ({stockValuesChange > 0 ? "+" : ""}
+          {new Intl.NumberFormat().format(stockValuesChangePercentage)}%)
```
jackwellerreal, line 76: These should still be `Math.abs(stockValueChange)` as it is already shown in the title above if the stock price has gone down or up. This is not needed to be changed since it should always be shown as a positive percentage

Issue #2141, opened 2025-01-27
bug: system health widget using dash. misrepresents available space


sharko789:
### Describe the bug

When using dash. as datasource for the system health widget, the available space is not calculated. While the visual percentage seems correct, the tooltip just shows the drive's entire size as being available.

1TB drive, which is about a quarter used:
![Image](https://github.com/user-attachments/assets/6992e0b2-b815-4c99-b5af-590193f27ce0)

### Steps to reproduce

1. Install homarr and dash.
2. Setup a board with a health monitoring widget using dash. as datasource

### Impact

wrong information, minor annoyance

### Version

1.2.0

### Installation method

Portainer

### Browser

Firefox

Issue #2140, opened 2025-01-27
bug: system health widget using dash. misrepresents unmounted drives


sharko789:
### Describe the bug

When a drive is not mounted, dash. is able to detect and list it under the /info endpoint. But under the /load/storage endpoint it reports a load of -1 Byte. This edge case isn't handled by the system health widget, instead it displays a small percentage as being used.

![Image](https://github.com/user-attachments/assets/a8eb5cc2-e8d0-452d-9350-34c847dc1e6f)

### Steps to reproduce

1. Have a machine with at least two drives
2. Boot it so that one drive stays unmounted
3. Install homarr and dash.
4. Setup a board with a health monitoring widget using dash. as datasource

### Impact

minor visual annoyance

### Version

1.2.0

### Installation method

Portainer

### Browser

Firefox

PR #2885, opened 2025-04-16
fix: calculate available space correctly and filter unmounted drives

rezstje:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

I've tested this with an unmounted drive and they are being filtered out now. However, the info you get back from the dashdot API is very limited, so the whole function of mapping indexes from the /load/storage endpoint to the /info endpoint might not always be correct. But this was already there before, I only applied a filter to remove the unmounted drives and fix the available storage to display the correct info

fixes: https://github.com/homarr-labs/homarr/issues/2141
fixes: https://github.com/homarr-labs/homarr/issues/2140
image
image


Meierschlumpf (approved):
Thank you for the contribution!

manuel-rw (approved):

PR #2497, opened 2025-03-04
feat: add releases widget

Aandree5:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

**Thank you for your contribution. Please ensure that your pull request meets the following pull request:**

- [x] Builds without warnings or errors (``pnpm build``, autofix with ``pnpm format:fix``)
- [x] Pull request targets ``dev`` branch
- [x] Commits follow the [conventional commits guideline](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] No shorthand variable names are used (eg. ``x``, ``y``, ``i`` or any abbrevation)

Requires: https://github.com/homarr-labs/homarr/pull/2496

The goal for this is to have a widget where we can load a series of repositories that we want to follow and show their new releases. Sometimes getting a notification is not the most useful thing because we may not be able to act on it right there and then, so with this we can at any point open the dashboard and see what new releases are available and which ones we want to act on.

https://github.com/user-attachments/assets/2c264c57-ef9a-4e04-a93a-0c2a0be0e013

    
image
**Left:**
    - Show Only Highlighted: **off**
    - Show Details: **on**
    
**Middle:**
    - Show Only Highlighted: **on**
    - Show Details: **off**
    
**Right:**
    - Show Only Highlighted: **on**
    - Show Details: **on**
    - 
image

**Options page:**
image
image
The `Version Regex` box is an autocomplete field, so we can have standard regex codes but still allow the user to make their own, if they want to track a specific release type.

### Edit 09 Apr 2025
**Version Filtering Changed**
`versionRegex` field has been changed to `versionFilter`, to help reduce the risk of malicious RegEx patterns

https://github.com/user-attachments/assets/a1a36dcf-0ae7-4af3-aa16-72985fe7b49e

PR #2891, opened 2025-04-17
fix(releases-widget): error display, decouple database repository object from responses and batch widget queries

### LINKED ISSUE #2952: Release Widget - Homarr crash
Every time I have more than 10 entries in the new release widget I get the following error message:

Image

Additionally, the language changed upon restart. I normally use German, but now I had Japanese and Polish...

### LINKED ISSUE #2954: Release Widget - Docker Hub rollback
So, I've added both Github and Docker Hub entries. For example, if I click on Homarr and then open release, I land on Github. But if I click on open project on Docker Hub, a new Homarr page simply opens. Is that normal?

Aandree5:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Dpends on: https://github.com/homarr-labs/homarr/pull/2497

This improves a couple points:
- **User can now see the errors**. Error releases will always display (regardless of the "Show Only Highlighted" swicth). By expanding the release, the user will be able to see the error message. See examples below.
- **Decoupled the database object from the response object**. When in board editing mode, all the data received from the provider Api was being saved into the database as part of the object. This would in some cases cause issues, if a new release did not have a description, the description from the old release would be shown. The same would happpen with the error message.
- **Batch widget queries** (fixes #2952). When adding enough repositories the `useSuspenseQuery` was throwing a `input too big` error. Changed this to batching the repositories in chinks of 5, and calling `useSuspenseQueries` so all the calls happen in one hook. The max number of repositories that go on a query call will depend on the size of each individual field on each repository
- fixes #2954 

image
image


Aandree5:
This is fixed with 4b52612de4608a673021393acc482a2feddca909
@Meierschlumpf Should I merge this PR with the previous one since we missed the release on Fraiday or keep both separate PRs?

Aandree5:
In the [packages/request-handler/src/releases.ts](https://github.com/homarr-labs/homarr/pull/2891/files/06dee62b675e6707a0cb4c06f789071a42545f4e#diff-fa1fa5a5f6e2c1e6e602ae89318ce04cf7babdb15a16e2d91ef0cda9dc7066fd), I have the follwoing check to see if there are any version with the given version filter
```
 const latest =
        releases.length === 0
          ? formatErrorRelease(input.identifier, input.providerKey, "Could not find any releases for version filter")
```

I was thinking on having the message localised

Meierschlumpf:
Okay I see, IMO most sense would make to return the error as key like `noMatchingVersion` or similar and sent it that way to the frontend where it then could be translated

Aandree5:
Fixed the error localisation by adding a code field with `noMatchingVersion` as suggested 8d3831dc664ce88a612f8b7eef7df4180ecb6bd6.
Fixed a new issue, where adding too many repositories was making `useSuspenseQuery` throw an error of `input too big`. replaced with call to `useSuspenseQueries` with batches of 5 repositories at a time 9a2596f90aea2addeac8f1e2c9e0f2b277ff56d8

Aandree5:
If you need me to split this PR so it doesn't show all the previous commits let me know

Aandree5:
This also fixes #2954, by setting missing data from the api calls to `undefined` instead of using defaults, helps to better check for missing information

packages/request-handler/src/releases-providers.ts
```diff
@@ -67,12 +63,7 @@ export const Providers: ProvidersProps = {
           ),
         })
         .transform((resp) => {
-          return resp.results.map((release) => ({
-            ...release,
-            releaseUrl: "",
-            releaseDescription: "",
-            isPreRelease: false,
-          }));
+          return resp.results.map((release) => release);
```
manuel-rw, line 66: Isn't this redundant?
```suggestion
          return resp.results;
```

packages/widgets/src/releases/component.tsx
```diff
@@ -55,61 +56,73 @@ function isDateWithin(date: Date, relativeDate: string): boolean {
   }
 }
 
+function chunkArray<T>(array: T[], size: number): T[][] {
+  const chunks: T[][] = [];
+  for (let i = 0; i < array.length; i += size) {
+    chunks.push(array.slice(i, i + size));
+  }
+  return chunks;
+}
```
manuel-rw, line 65: - Do we not have an existing function for this?
- Please extract from this class, it has no business here and is mostly unrelated to the code. Perhaps add test coverage

manuel-rw (changes requested):

packages/widgets/src/releases/component.tsx
Aandree5: I do agree that doesn't belong there, I couldn't find one, will have another look. But if you know where it's at that would be great.

Aandree5: fixed with 3aa31b442699da7c6c09fb3b1d816156663f2579
- I created a new `splitToChunksWithNItems` (can change the naem if you want). Mostly because the `splitToNChunks` changes the input array, which was clearing the repositories variable, it also works on the base of creating a set number of chunks.
- Took the time to also move the `isDateWithin` function into the `common` package, to help clean the `components.tsx`
- Added test coverage to `array.ts` and the new `date.ts` in the `common` package

Aandree5: fixed with 3aa31b442699da7c6c09fb3b1d816156663f2579
- I created a new `splitToChunksWithNItems` (can change the naem if you want). Mostly because the `splitToNChunks` changes the input array, which was clearing the repositories variable, it also works on the base of creating a set number of chunks.
- Took the time to also move the `isDateWithin` function into the `common` package, to help clean the `components.tsx`
- Added test coverage to `array.ts` and the new `date.ts` in the `common` package

packages/common/src/date.ts
```diff
@@ -0,0 +1,37 @@
+export const isDateWithin = (date: Date, relativeDate: string): boolean => {
```
manuel-rw, line 1: Why not use https://day.js.org/docs/en/query/is-between ?

packages/common/src/test/array.spec.ts
manuel-rw, line 1: ❤️

manuel-rw (changes requested):

packages/common/src/date.ts
Aandree5: Honestly I just wasn't aware of it. I'm still learning about all the packages I can use. Day to day work is more .NET but I do like and want to learn more TS.
Although that `isBetween` does not have the exact same functionality, I've adapted the code to use it, at least the calculations will be handled by `dayjs`, which should be best. Fixed with 57768174fd6e498328726e838a7b11eadee665f5

I also took the liberty to adjust the units to match `dayjs`, basically just changed it to use `M` for month instead of `m`. I'm not sure about this in terms of breaking changes. Let me know what you think, do I just map `m` to `M` in the funtion or do we go with the breaking change and just use `M`? 
In `dayjs`, `m` is used for minutes which I don't think we will be using in the widget, although you never know.

Meierschlumpf: I would suggest that we map the parameters so that the casing does not matter, so users can also insert `D` for days if they want. I think that would be the easiest for this

Aandree5: Check 9a3c304041907a55c1341c0ef2d02c286251d229
I was thinking at doing it has the user types it int he options, but couldn't find a way without editing the "text" input component. So tried doing it in the `component.tsx` should only happen once per widget if I'm not mistaken, so it shouldn't be too bad, let me know you thoughts.

manuel-rw: Don't worry, .NET has great date-types that are more easy and flexible to use. Sometimes dates can be quite annoying in Node.js. The solution looks good, thanks!

manuel-rw (approved):
Looks good to me, maybe wait for a review from @Meierschlumpf

packages/widgets/src/releases/component.tsx
```diff
@@ -10,106 +10,95 @@ import {
...
-
-    case "y":
-      return diffHours / (24 * 365) < amount;
-
-    default:
-      throw new Error("Invalid unit");
-  }
-}
+const formatRelativeDate = (value: string): string => {
+  const isMinutes = /\d+m/g.test(value);
```
Meierschlumpf, line 31: Isn't this isMonths?

Meierschlumpf (changes requested):

packages/common/src/array.ts
```diff
@@ -5,3 +5,11 @@ export const splitToNChunks = <T>(array: T[], chunks: number): T[][] => {
   }
   return result;
 };
+
+export const splitToChunksWithNItems = <T>(array: T[], items: number): T[][] => {
```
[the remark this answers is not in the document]
Aandree5: Done 9a0721568a8fa0216cc3914d28491af6af693dd8

packages/request-handler/src/releases.ts
```diff
@@ -8,22 +8,49 @@ import { createCachedWidgetRequestHandler } from "./lib/cached-widget-request-ha
 import { Providers } from "./releases-providers";
 import type { DetailsResponse } from "./releases-providers";
 
+const _errorSchema = z.object({
```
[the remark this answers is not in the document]
Aandree5: .NET habit, done e49a5207c6cf79015774877e0d408e38b4d762b3

packages/widgets/src/releases/component.tsx
Aandree5: Don't even have words, sorry. Done a5769e41d3acde03d0ec7d3527747708c8b22935

manuel-rw:
@Meierschlumpf please re-review

Meierschlumpf (changes requested):
Okay nvm there are some lint issues, please fix them first

Aandree5:
Not sure how I missed that, fixed 3461c797fa4185ca455a116714ec8c0082f0b628

Meierschlumpf (approved):

PR #2900, opened 2025-04-18
feat: add ntfy integration

itzTheMeow:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

**WORK IN PROGRESS**

Adds widget to show [ntfy](https://ntfy.sh/) notifications.

- Adds widget to show notifications from notification-based integrations.
- Adds new integration category `notifications`.
- Adds ntfy integration for notifications.
- Adds new "every 30 seconds" cron timer.

Preview:
image

I attempted to make the `notifications` widget accessible to other notification services such as [gotify](https://gotify.net/), however not all services use topics to define the notification locations so it may need reworking if another service is to be supported.

The ntfy integration supports authentication or no authentication. To test the URL can be set to `https://ntfy.sh` and access token left unset. Some example topics to subscribe to: `mytopic`, `test`
To publish messages to these topics, use https://ntfy.sh/mytopic

Additional notes:
I set the cron timer for these to 30 seconds to avoid spamming the API with requests every 5 seconds (was getting ratelimited).

The integrations selector allows for multiple ntfy integrations to be selected for the same widget. The selected topics are shared between both instances so most likely one or the other will not have a specified topic.


itzTheMeow:
Note:
The icon here (new integration dropdown) is cut off, due to the way its shaped. Is there a way to fix this or should I just leave it?
image

ajnart:
> Note: The icon here (new integration dropdown) is cut off, due to the way its shaped. Is there a way to fix this or should I just leave it? image

I'd say don't bother with it for now, we'll review the code and once every's good we can look into that

ajnart:
Don't forget to mark the PR as ready for review when you're done ;)

itzTheMeow:
My config for the screenshot (left to right, top to bottom):

Integrations:
`ntfy`: `https://ntfy.sh` - No Token
`main`: `https://ntfy.xela.codes` - Auth Token Provided (my personal ntfy instance)

Widgets:
`main` Selected
`meow-backups` Topics

`ntfy` Selected
`test`, `mytopic` Topics

`ntfy` Selected
No Topics

`ntfy` Selected
`mytopic` Topics

packages/integrations/src/ntfy/ntfy-schema.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from "zod";
+
+// There are more properties, see: https://docs.ntfy.sh/subscribe/api/#json-message-format
```
manuel-rw, line 3: Thanks for linking, this is really useful

packages/request-handler/src/notifications.ts
```diff
@@ -0,0 +1,20 @@
...
+export const notificationsRequestHandler = createCachedIntegrationRequestHandler<
+  Notification[],
+  IntegrationKindByCategory<"notifications">,
+  { topics: string[] }
+>({
+  async requestAsync(integration, input) {
+    const integrationInstance = await createIntegrationAsync(integration);
+    return await integrationInstance.getNotificationsAsync(input.topics);
+  },
+  cacheDuration: dayjs.duration(5, "seconds"),
```
manuel-rw, line 18: You set the interval to 30 seconds but the cache to 5 seconds? Must it really expire so quickly? I'd say the cache could expire after 1 minute or 30 seconds

manuel-rw (changes requested):
Thanks a lot for this contribution. Let me know if you have any questions

packages/definitions/src/integration.ts
```diff
@@ -169,6 +169,12 @@ export const integrationDefs = {
     iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons@master/png/unifi.png",
     category: ["networkController"],
   },
+  ntfy: {
+    name: "ntfy",
+    secretKinds: [[], ["apiKey"]],
```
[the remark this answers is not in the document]
itzTheMeow: Topics can be password-protected which requires authentication, but this is not required. Some topics are public and can be accessed without a token.

packages/integrations/src/interfaces/notifications/notifications-integration.ts
```diff
@@ -0,0 +1,7 @@
+import { Integration } from "../../base/integration";
+import type { Notification } from "./notification";
+
+export abstract class NotificationsIntegration extends Integration {
+  /** Get current notification list. */
+  public abstract getNotificationsAsync(topics: string[]): Promise<Notification[]>;
```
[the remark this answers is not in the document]
manuel-rw: Yes.

manuel-rw:
Hi @itzTheMeow , do you need further input from us or what is the status of this?

itzTheMeow:
Hey, currently on vacation so I haven't had time to work on this. It's still in my plans.

manuel-rw:
No worries, thanks for the update. Enjoy the vacation

manuel-rw:
Hi @itzTheMeow , any updates? :)

Meierschlumpf:
No problem, we only wanted to now if you are still interested in working on this. Otherwise we would either have dismissed it or implemented it ourself.

itzTheMeow:
This actually works quite nicely, you can combine different URLs and topics in the same notification widget.
image
You can still provide multiple topics (for ntfy specifically) by comma-separating them.

packages/request-handler/src/notifications.ts
```diff
@@ -0,0 +1,20 @@
...
+export const notificationsRequestHandler = createCachedIntegrationRequestHandler<
+  Notification[],
+  IntegrationKindByCategory<"notifications">,
+  Record<string, never>
+>({
+  async requestAsync(integration) {
+    const integrationInstance = await createIntegrationAsync(integration);
+    return await integrationInstance.getNotificationsAsync();
+  },
+  cacheDuration: dayjs.duration(30, "seconds"),
```
manuel-rw: It's this a bit too aggressive? In my opinion, 5 minutes should be plenty!
Remember, Homarr will fetch even if there is no user "active", meaning that we need to be careful with setting too low intervals.

manuel-rw (changes requested):

manuel-rw (approved):
LGTM, @homarr-labs/maintainers please add a second review.

packages/integrations/src/ntfy/ntfy-integration.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  public async getNotificationsAsync() {
+    const url = this.getTopicURL();
+    const notifications = await Promise.all(
+      (
+        await fetchWithTrustedCertificatesAsync(url, { headers: this.getHeaders() })
+          .then((response) => {
+            if (!response.ok) {
+              throw new Error(response.statusText);
```
Meierschlumpf: This should use the `ResponseError` instead

packages/integrations/src/ntfy/ntfy-integration.ts
```diff
@@ -0,0 +1,67 @@
...
+            }
+            return response.text();
+          })
+          .catch((error) => {
+            if (error instanceof Error) {
+              throw new Error(error.message);
```
Meierschlumpf: I would suggest to just throw the existing error, so just `throw error` instead of `throw new Error(error.message)`

packages/integrations/src/ntfy/ntfy-schema.ts
```diff
@@ -0,0 +1,12 @@
...
+// Not all properties are required for this use case.
+export const ntfyNotificationSchema = z.object({
+  id: z.string(),
+  time: z.number(),
+  event: z.literal("message"), // we only care about messages
```
Meierschlumpf: Do I understand correctly, that we *let the parsing fail* if the event is different than message? IMO it would make more sense to just have event a string then and check in the integration that the event is message. Then we could also always throw a `ParseError` if there is an issue with parsing (just using `parseAsync` would be enough because of the existing event handler in the base integration for zod)

packages/integrations/src/ntfy/ntfy-schema.ts
```diff
@@ -0,0 +1,12 @@
...
+  topic: z.string(),
+  title: z.optional(z.string()),
```
Meierschlumpf, line 10: In the docs it says that it already defaults to `nfty.sh/topic` so can this be undefined?

packages/definitions/src/integration.ts
```diff
@@ -169,6 +170,12 @@ export const integrationDefs = {
     iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons@master/png/unifi.png",
     category: ["networkController"],
   },
+  ntfy: {
+    name: "ntfy",
+    secretKinds: [["topic"], ["topic", "apiKey"]],
```
Meierschlumpf, line 175: In the future we'll be able to replace the topic with something that can be defined per widget (on base of the integration). But for now this is ok 👍🏼

packages/widgets/src/notifications/component.tsx
```diff
@@ -0,0 +1,102 @@
...
+    </ScrollArea>
+  );
+}
+
+const InfoDisplay = ({ date }: { date: string }) => (
+  <Group gap={5} align={"center"}>
+    <IconClock size={"1rem"} color={"var(--mantine-color-dimmed)"} />
+    <Text size="sm" c="dimmed">
+      {date}
+    </Text>
```
Meierschlumpf: Maybe you could use the `useTimeAgo` hook that automatically updates the timestamp. Maybe you can add a second argument that specifies the update frequency and make it a longer time between updates here

Meierschlumpf (changes requested):
Thank you for the contribution ❤️
Looks good beside a few small things

packages/integrations/src/ntfy/ntfy-schema.ts
itzTheMeow: Looking at `Notification message (minimal)` under the examples shows a message without a title.
Polling the endpoint can also possibly return messages without a title:
```json
{"id":"Sl3vCmZ3V3VD","time":1749318367,"expires":1749361567,"event":"message","topic":"mytopic","message":"test"}
```

Meierschlumpf: Okay then we'll leave it like that, thanks for clarification

packages/integrations/src/ntfy/ntfy-integration.ts
```diff
@@ -0,0 +1,65 @@
...
+            else {
+              throw new Error("Error communicating with ntfy");
+            }
+          })
+      )
+        // response is provided as individual lines of JSON
+        .split("\n")
+        .map(async (line) => {
+          // ignore empty lines
+          if (line.length == 0) return null;
```
Meierschlumpf:
```suggestion
          if (line.length === 0) return null;
```

packages/integrations/src/ntfy/ntfy-integration.ts
```diff
@@ -0,0 +1,65 @@
...
+
+          const json = JSON.parse(line) as unknown;
+          const parsed = await ntfyNotificationSchema.parseAsync(json);
+          if (parsed.event == "message") return parsed;
```
Meierschlumpf:
```suggestion
          if (parsed.event === "message") return parsed;
```

Meierschlumpf (changes requested):

Meierschlumpf (dismissed):

PR #3082, opened 2025-05-11
fix(releases-widget): delete the empty repository when canceling  `Add Reposotory`

### LINKED ISSUE #3067: feat: Release Widget: No new repositories if a new one was added but nothing was changed and it is canceled

What I noticed is that if you're editing the Release widget and want to add a new repository, you click Add, but then decide against it and haven't changed anything and click Cancel, a new one is added at the bottom, even though you clicked Cancel and haven't changed anything. It would be better if it really wasn't created. Is that possible?

Image
Image

### Additional information

_No response_

Aandree5:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

fixes #3067

### DIFF
--- a/packages/widgets/src/_inputs/widget-multiReleasesRepositories-input.tsx
+++ b/packages/widgets/src/_inputs/widget-multiReleasesRepositories-input.tsx
@@ -83,11 +83,12 @@ export const WidgetMultiReleasesRepositoriesInput = ({
       fieldPath: `options.${property}.${index}`,
       repository: item,
       onRepositorySave: (saved) => onRepositorySave(saved, index),
+      onRepositoryCancel: () => onRepositoryRemove(index),
       versionFilterPrecisionOptions,
     });
   };
 
-  const onReleaseRemove = (index: number) => {
+  const onRepositoryRemove = (index: number) => {
     form.setValues((previous) => {
       const previousValues = previous.options?.[property] as ReleasesRepository[];
       return {
@@ -99,6 +100,7 @@ export const WidgetMultiReleasesRepositoriesInput = ({
       };
     });
   };
+
   return (
     <Fieldset legend={t("label")}>
       <Stack gap="5">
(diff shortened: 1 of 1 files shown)


packages/widgets/src/_inputs/widget-multiReleasesRepositories-input.tsx
```diff
@@ -210,6 +213,14 @@ const ReleaseEditModal = createModal<ReleaseEditProps>(({ innerProps, actions })
     setLoading(false);
   }, [innerProps, tempRepository, actions]);
 
+  const handleCancel = useCallback(() => {
+    if (innerProps.onRepositoryCancel) {
+      innerProps.onRepositoryCancel();
+    }
+
+    actions.closeModal();
+  }, [innerProps, actions]);
```
Meierschlumpf, line 222: Cancel should not always remove the repository and only if it was newly created

Meierschlumpf (changes requested):

packages/widgets/src/_inputs/widget-multiReleasesRepositories-input.tsx
Aandree5: @Meierschlumpf yes it will only remove if it's a new repository and the identifier is empty for a couple resons:
- There are only two calls to open the modal, one by editing a repository (line `140`) and one to add a new repository (line `82`), and the `onRepositoryCancel` is currently only passed in by the `openModal` action of the `addNewItem`. The `openModal` from the `edit` button does not pass any method to it. Like the model is just notifying the parent of the cancel action, if needed.
- Also that method (which is only passed in by the `addNew...`) will only delete if the identifier is empty, something that you can't do by editing a repository because the from won't validate and won't let the user close the model with `confirm`. The only way the `identifier` is empty is by trying to add a new repository and cancel it.

Meierschlumpf: Oh sorry, haven't seen that there are two different open modal statements where one has the removal and the other does not. That makes total sense then

Meierschlumpf (approved):

PR #3121, opened 2025-05-16
Logo Customization from custom CSS

tech62:
Add a "logo" class to customize the logo with CSS

### DIFF
--- a/apps/nextjs/src/components/layout/logo/logo.tsx
+++ b/apps/nextjs/src/components/layout/logo/logo.tsx
@@ -9,13 +9,13 @@ interface LogoProps {
   shouldUseNextImage?: boolean;
 }
 
-export const Logo = ({ size = 60, shouldUseNextImage = false, src, alt }: LogoProps) =>
+export const Logo = ({ size = 60, shouldUseNextImage = false, src, alt, class }: LogoProps) =>
   shouldUseNextImage ? (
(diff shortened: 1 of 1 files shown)


Meierschlumpf:
Hey @tech62 
Thank you for your contribution. I can fix the pr title if you want, beside that, instead of `class=` it would need to be `className=` because JSX is part of javascript and class is a reserved keyword. Also the  LogoProps interface would either need that property or you can remove it as we currently don't need to pass a className into our logo component

Meierschlumpf:
The conventions for commits can be ignored as it will be squashed anyway

tech62:
> Hey @tech62 Thank you for your contribution. ~I can fix the pr title if you want~, beside that, instead of `class=` it would need to be `className=` because JSX is part of javascript and class is a reserved keyword. Also the LogoProps interface would either need that property or you can remove it as we currently don't need to pass a className into our logo component

I've made the change, if you want to check.
I'm not a javascript developper, i do my best :)

Meierschlumpf:
All good, I'll fix the format issue really quick locally

Meierschlumpf:
Once it is merged it will be part of the next release (we normally release every friday night and sometimes if needed / wanted by us. And the dev image tag will also be released about 5 minutes after the pr is merged

Meierschlumpf (approved):

PR #3212, opened 2025-05-24
fix(visual): check for creator board permission is wrong

### LINKED ISSUE #2671: bug: Edit and Settings button visible to non logged in users`
### Describe the bug

im not logged in and i can see the edit and settings button for the board, it errors out when i try to make an edit or go to the settings pages but it would be nice if they were hidden

### Steps to reproduce

Just don't be logged in

### Impact

Causes confusion to users and I would rather them not even be able to mess with the board even if it doesn't save

### Additional information

_No response_

### Version

1.12.0

### Installation method

Docker Compose

### Browser

Chrome

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Closes #2671

This could have been pretty bad! If we for example included the creator in this query, it would have been possible to bypass the view check for non public boards without an owner (so most likely imported ones) when opening them without being logged in:

```
const board = await db.query.boards.findFirst({
    where: boardWhere,
    columns: {
      id: true,
      creatorId: true,
      isPublic: true,
    },
    with: {
      userPermissions: {
        where: eq(boardUserPermissions.userId, session?.user.id ?? ""),
      },
      groupPermissions: {
        where: inArray(boardGroupPermissions.groupId, groupsOfCurrentUser.map((group) => group.groupId).concat("")),
      },
    },
  });
```

I've checked all paths where the createBoardPermissions method was used and luckily none of them results in an unauthorized access. The only thing is, that some visual things are wrong.

Usages of throwIfActionForbiddenAsync would never fall into this because of the creator not included / all except two where used anyway with protected routers (so session is never undefined and rather null `Thank you ===`)

Previously:
image

Now:
image

### DIFF
--- a/packages/auth/permissions/board-permissions.ts
+++ b/packages/auth/permissions/board-permissions.ts
@@ -23,21 +23,22 @@ export type BoardPermissionsProps = (
 
 export const constructBoardPermissions = (board: BoardPermissionsProps, session: Session | null) => {
   const creatorId = "creator" in board ? board.creator?.id : board.creatorId;
+  const isCreator = session !== null && session.user.id === creatorId;
 
   return {
     hasFullAccess:
-      session?.user.id === creatorId ||
+      isCreator ||
       board.userPermissions.some(({ permission }) => permission === "full") ||
       board.groupPermissions.some(({ permission }) => permission === "full") ||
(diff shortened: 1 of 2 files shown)


manuel-rw (approved):
Good catch, thanks for fixing. No vuln report nessesary

Issue #1141, opened 2024-09-17
bug: dns requests are not cached


SeDemal:
### Provide environment information

any

### Describe the bug

Some queries are not cached but should. This results in an insane amount of queries (This is a 7h span).
This is because the media information for each element needs to be queried separately, resulting in 20 calls every 5 seconds.
![image](https://github.com/user-attachments/assets/470abf5a-0313-48b9-a0b5-e3b26a6cbbb7)

### To reproduce

Mostly visible using the media request widgets.
Let the page opened for a while then go check your queries in pihole.

manuel-rw:
I think the requests are behaving as designed - we don't know whether the remote data has changed or not - therefore we must fetch all the time.
However, 128K DNS requests do seem extreme - I wonder whether we don't properly cache DNS resolve requests.
@SeDemal can you confirm, that the problem is the **amount of DNS requests** and not the **amount & frequency of HTTP requests**? As far as I saw, we cannot reduce the amount of HTTP requests (except the frequency) because the data we need is distributed over multiple API endpoints.

bonovski:
I had to stop the container, because tasks.cjs burns CPU cycles and spams pihole constantly.

manuel-rw:
I'm analyzing this now. I will dump some requests using Wireshark first...

manuel-rw:
Here are my findings so far:
- Homarr uses ``undici`` to fetch data from integrations, which does not have any DNS caching by itself. It always lookups all hostnames with a DNS request, which is visible in Wireshark: <br/> ![Image](https://github.com/user-attachments/assets/45569876-e4b0-4df2-9e2c-73c97db7dea7)
These packets are being sent every time Homarr is fetching data from any integration or any website (e.g. GitHub or jsdelivr).
- The above is slightly odd, because [Docker will inherit the DNS settings from your host, unless you are using a custom network](https://docs.docker.com/engine/network/#dns-services). The upstream DNS server should handle DNS caching, not your application (e.g. Homarr).
- After inspecting the DNS requests with Wireshark, I noticed that PiHole is always returning a TTL of ``0`` when replying to the DNS resolve requests of Homarr: <br/>
![Image](https://github.com/user-attachments/assets/52e6b97c-eccf-4ce7-96cf-9b8963de0162)
A TTL of ``0`` means that it **expires immediately** and will (and should!) not be cached. This is normally something that your DNS will control and should not be overridden in your client, unless you know what you're doing
[...]

PR #3235, opened 2025-05-26
feat: reduce cron job intervals

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Reduces intervals for #1141


Meierschlumpf:
Does it make sense to also increase the update interval (so that it's updated less) for the other jobs? Should we do this as part of this PR? Other than that, approved

manuel-rw:
These are the most "expensive" jobs, hence we can start with these and do more later.

Issue #3294, opened 2025-06-01
bug: OpenMediaVault integration fails


lthomsen:
### Describe the bug

Whenever I add my openmediavault to the system health monitoring, the widget fails:

![Image](https://github.com/user-attachments/assets/71192fd1-dd49-44c7-b3e1-accfc8cfcc6f)

The log simply says:
2025-06-01T17:24:47.131Z error Failed to run integration job integration=lyhfs55q4wn2q1ky318mwo61 inputHash='W3t9XQ=='

### Steps to reproduce

Create integration to OpenMediaVault. Add to system health monitoring widget.

### Impact

Makes the widget unuseable

### Additional information

From what I can tell, it might be because my OMV returns invalid values from the RPC calls, most likely the system / getinformation call:
{
    "response": {
        "ts": 1748795990,
        "time": "Sun 01 Jun 2025 06:39:50 PM CEST",
        "hostname": "openmediavault",
        "version": "7.7.7-1 (Sandworm)",
        "cpuModelName": null,
        "cpuUtilization": 0.375,
        "cpuCores": null,
        "cpuMhz": null,
        "memTotal": "8303374336",
        "memFree": "271605760",
        "memUsed": "3273465856",
        "memAvailable": "5029908480",
        "memUtilization": "0.39423",
        "kernel": "Linux 6.1.57",
        "uptime": 10186329.17,
        "loadAverage": {
            "1min": 0.01,
            "5min": 0.09,
            "15min": 0.25
        },
        "configDirty": false,
        "dirtyModules": [],
        "rebootRequired": false,
        "availablePkgUpdates": 1,
        "displayWelcomeMessage": false
    },
    "error": null
}

cpuModelName is null, which seems to be invalid when compared to packages/integrations/src/openmediavault/openmediavault-types.ts

Homarr is installed on an LXC in Proxmox using the Promox VE helper scripts.

### Version

1.22.0

### Installation method

Other (describe in "additional information")

### Browser

Edge (Chromium)

lthomsen:
Just quick add:
On my own installation, I've done the following:
in openmediavault-types.ts
changed line 7 from
cpuModelName: z.string(), 
to 
cpuModelName: z.string().nullable(),
and in openmediavault-integration.ts
changed line 69 from
cpuModelName: systemResult.data.response.cpuModelName,
to
cpuModelName: systemResult.data.response.cpuModelName && "Unknown CPU",

[...]

PR #3310, opened 2025-06-03
fix: Allow nullable on cpuModelName and default to "Unknown CPU" if null in OMV Integration

lthomsen:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

fix for #3294

### DIFF
--- a/packages/integrations/src/openmediavault/openmediavault-integration.ts
+++ b/packages/integrations/src/openmediavault/openmediavault-integration.ts
@@ -66,7 +66,7 @@ export class OpenMediaVaultIntegration extends Integration {
 
     return {
       version: systemResult.data.response.version,
-      cpuModelName: systemResult.data.response.cpuModelName && "Unknown CPU",
+      cpuModelName: systemResult.data.response.cpuModelName ?? "Unknown CPU",
       cpuUtilization: systemResult.data.response.cpuUtilization,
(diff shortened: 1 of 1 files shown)


manuel-rw (changes requested):
Please fix the builds and check the TODO list. Code change looks good

lthomsen:
@manuel-rw the issues should be fixed and the todo has been checked.

packages/integrations/src/openmediavault/openmediavault-integration.ts
```diff
@@ -66,7 +66,7 @@ export class OpenMediaVaultIntegration extends Integration {
 
     return {
       version: systemResult.data.response.version,
-      cpuModelName: systemResult.data.response.cpuModelName,
+      cpuModelName: systemResult.data.response.cpuModelName && "Unknown CPU",
```
Meierschlumpf: Is the goal here to show the text `Unknown CPU`? If so this will not work and should be replaced with the following:
```suggestion
      cpuModelName: systemResult.data.response.cpuModelName ?? "Unknown CPU",
```

Meierschlumpf (changes requested):

Meierschlumpf:
Perfect, we can ignore the semantic commits workflow, it's okay that way as the changes will be squashed anyway (so into one commit that follows the conventions)

PR #1959, opened 2025-01-15
feat: Health Monitoring widget unification with graphs

SeDemal:
…n Progress)

<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

**Thank you for your contribution. Please ensure that your pull request meets the following pull request:**

- [ ] Builds without warnings or errors (``pnpm buid``, autofix with ``pnpm format:fix``)
- [ ] Pull request targets ``dev`` branch
- [ ] Commits follow the [conventional commits guideline](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] No shorthand variable names are used (eg. ``x``, ``y``, ``i`` or any abbrevation)

Relates to #2923

PR #3538, opened 2025-07-03
feat: system resources widget

### LINKED ISSUE #2923: Enhance (bring back) Dashdot style
Like in the old Homarr (that I still run)
Image

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Superseeds #1959 
Closes #2923

- [x] Support light mode
- [x] Define a margin for the network traffic graphs
- [x] Fix tooltip position in small widgets
- [x] Add support for opacity


Meierschlumpf:
Is the plan here to add the graphs widget (like dashdot was in 0.15.10?)
Just to know what this is about 😉

manuel-rw:
Yes, correct. The old branch is messed up and I'm not satisifed with it, hence starting over. Sorry.
I also decided to make a new widget instead of replacing the old widget. My idea is this:
- We do not replace the widgets automatically
- The system resources support more integrations, e.g. Unraid, TrueNAS, ....
- The system health shows more S.M.A.R.T or system related data, not CPU, RAM, ...
- We will improve the design of the system health later and remove CPU, RAM in a second iteration

Meierschlumpf:
No problem, I agree that it is a mess 👍🏼 
I think it's a lot easier to start over and make a simple widget

manuel-rw:
@Meierschlumpf can you have an early view?
I think we can get this out for this week's release.

packages/widgets/src/system-resources/component.tsx
```diff
@@ -0,0 +1,94 @@
...
+      cpu: itemData.healthInfo.cpuUtilization,
+      memory: itemData.healthInfo.memUsedInBytes,
+      network: itemData.healthInfo.network,
+    }));
+    queueHandlers.setState((queue) => [...queue, ...items].slice(0, MAX_QUEUE_SIZE));
+
+    if (data[0]) {
+      setMemoryCapacityInBytes(data[0].healthInfo.memAvailableInBytes + data[0].healthInfo.memUsedInBytes);
+    }
+  }, [data, queueHandlers]);
```
Meierschlumpf: I currently get the following error when it tries to update the charts:
image

I think it would make sense to instead just define the initial value of the `useListState` after the `useQuery` and then update it only through the `useSubscription`

Meierschlumpf (commented):
Also the size of the charts is to big, especially for small ones, except of course this is intended

packages/widgets/src/system-resources/component.tsx
manuel-rw: Sorry, I messed something up while fixing some code smells. Will fix.

packages/widgets/src/system-resources/index.ts
```diff
@@ -0,0 +1,12 @@
+import { IconGraphFilled } from "@tabler/icons-react";
+
+import { createWidgetDefinition } from "../definition";
+import { optionsBuilder } from "../options";
+
+export const { definition, componentLoader } = createWidgetDefinition("systemResources", {
+  icon: IconGraphFilled,
+  supportedIntegrations: ["dashDot", "openmediavault", "proxmox"],
```
Aandree5:
```suggestion
  supportedIntegrations: ["dashDot", "openmediavault"],
```

The router does not accept `Proxmox`, so shouldn't be able to add it

Aandree5: A question on this, can't `Proxmox` be used on this widget?

packages/integrations/src/dashdot/dashdot-integration.ts
```diff
@@ -113,8 +115,13 @@ export class DashDotIntegration extends Integration implements ISystemHealthMoni
     };
   }
 
+  private async getCurrentNetworkLoadAsync() {
+    const memoryLoad = await fetchWithTrustedCertificatesAsync(this.url("/load/network"));
```
Aandree5: Should this be `networkLoad`?

packages/widgets/src/system-resources/chart/common-chart.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const scheme = useMantineColorScheme();
+  const board = useRequiredBoard();
+
+  tooltipProps = tooltipProps ?? {};
+  tooltipProps.allowEscapeViewBox = { x: true, y: true };
+  tooltipProps.wrapperStyle = { zIndex: 20 };
+
+  const opacity = board.opacity / 100;
+  const backgroundColor =
+    scheme.colorScheme == "dark" ? `rgb(67, 67, 67, ${opacity})` : `rgb(230, 230, 230, ${opacity})`;
```
Aandree5:
```suggestion
    scheme.colorScheme === "dark" ? `rgba(67, 67, 67, ${opacity})` : `rgba(230, 230, 230, ${opacity})`;
```

Opacity wasn't being taken into account.

Aandree5: Also think the backgrounds are a bit too strong (though this is just my personal opinion, so won't say it blocks the PR), was thinking more along the lines of:
- Dark: 57, 57, 57
- Light: 246, 247, 248 (followes the System Health Monitor grey color a bit more closely)

packages/widgets/src/system-resources/component.tsx
```diff
@@ -0,0 +1,94 @@
...
+          memory: data.healthInfo.memUsedInBytes,
+          network: data.healthInfo.network,
+        };
+        queueHandlers.setState((queue) => [...queue, obj].slice(0, MAX_QUEUE_SIZE));
+      },
+    },
+  );
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const showNetwork = queue.length === 0 ? true : queue[queue.length - 1]!.network != null;
```
Aandree5:
```suggestion
  const showNetwork = queue.length === 0 || queue[queue.length - 1]?.network !== null;
```

May I suggest this? No need for the `eslint` comment or the forced assertion

packages/widgets/src/system-resources/component.tsx
```diff
@@ -0,0 +1,94 @@
...
+      cpu: itemData.healthInfo.cpuUtilization,
+      memory: itemData.healthInfo.memUsedInBytes,
+      network: itemData.healthInfo.network,
+    }));
+    queueHandlers.setState((queue) => [...queue, ...items].slice(0, MAX_QUEUE_SIZE));
+
+    if (data[0]) {
+      setMemoryCapacityInBytes(data[0].healthInfo.memAvailableInBytes + data[0].healthInfo.memUsedInBytes);
+    }
+  }, [data]);
```
Aandree5: I'm sure you already seen this, but `queueHandlers` missing

packages/widgets/src/system-resources/component.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <SystemResourceMemoryChart
+          memoryUsageOverTime={queue.map((item) => item.memory)}
+          totalCapacityInBytes={memoryCapacityInBytes}
+        />
+      </div>
+      {showNetwork &&
+        (width > 200 ? (
+          <>
+            {/* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */}
+            <NetworkTrafficChart usageOverTime={queue.map((item) => item.network!.down)} isUp={false} />
```
Aandree5: I'm not sure if this could cause an issue or not, might be me just not liking the non-null forced assersions 😄, but the queue network is only checked for non null on the last item in the list (line 47), in theory, we could have it happen where the last item is not null but the previous were (maybe some issue with the integration) this would cause a crash.

We could do something similar to:
```ts
{showNetwork && (
        <NetworkCharts
          usageOverTime={queue.map((item) => item.network).filter((network) => network !== null)}
          isSingleChart={width > 200}
        />
      )}

[... removed for brevity]

const NetworkCharts = ({
  usageOverTime,
  isSingleChart,
}: {
  usageOverTime: {
    up: number;
    down: number;
  }[];
  isSingleChart: boolean;
}) =>
  !isSingleChart ? (
    <>
      <NetworkTrafficChart usageOverTime={usageOverTime.map((item) => item.down)} isUp={false} />

      <NetworkTrafficChart usageOverTime={usageOverTime.map((item) => item.up)} isUp />
    </>
  ) : (
    <div className={classes.colSpanWide}>
      <CombinedNetworkTrafficChart usageOverTime={usageOverTime} />
    </div>
  );
```

The worst that could happen here, is some network points missing in the graph.

packages/widgets/src/system-resources/component.tsx
manuel-rw: Yes, but then render cycles breaks completely. Could just be me missing something 🤔 
And ``queueHandlers`` is a const, it should never change, no?

Aandree5: makes sense, maybe just needs ignoring?

Meierschlumpf: It's object reference is changing every render, you'll probably need to use the actual method instead (so for example `queueHandlers.setState`. Then it should work

packages/widgets/src/system-resources/chart/combined-network-traffic.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <CommonChart
+      data={chartData}
+      dataKey={"index"}
+      series={[
+        { name: "up", color: "orange.5" },
+        { name: "down", color: "yellow.5" },
+      ]}
+      title={"NET"}
```
Meierschlumpf: Add missing translations

manuel-rw: Isn't "NET" similar to "CPU", it being a name?
IMO fine to leave untranslated

Meierschlumpf: I mean we can also leave it for now, I would just have said that we can translate all strings

Meierschlumpf (approved):
Please add the missing documentation

manuel-rw:
> Please add the missing documentation

Done: https://github.com/homarr-labs/documentation/pull/352

Issue #3350, opened 2025-06-07
feat: Release Widget: Docker import - third column for hidden entries


FasteryXD:
### Describe the feature you'd like to request

The problem here is that I simply took all the entries from Docker and pasted them in. But, as mentioned in #3344, not all of the identifiers are correct. I can edit them without any problems, and most of them work. But when I change the identifier, these entries appear again under "List of found images," which isn't actually correct because I added them. Therefore, my suggestion is to create a third row/column or something similar to hide such entries so that they no longer fall into this category.
Or are there better options?

![Image](https://github.com/user-attachments/assets/3819d84e-8dc9-4b85-9eae-8d6b6eceb64e)

Aandree5:
I think this would fix "itself" with the addition of #3344, as all those would be in the "already imported" section. When lscr.io is recognised

FasteryXD:
Unfortunately, the bug described in #3626 was not fixed by the latest update, so I'm hoping for an implementation like the one described in this post.

Aandree5:
> Unfortunately, the bug described in #3626 was not fixed by the latest update, so I'm hoping for an implementation like the one described in this post.

@FasteryXD I think this must be automatically done and an extra manual step isn't the solution. Has the other issue fixed part of the problem? Are all the remaining ones `linuxserver.io` or any other currently unrecognised provider?
If `linuxserver.io`, then they will be parsed properly with the referenced PR. 

If they are not, can you send a new screenshot of the repositories list (on the widget options) and the import screen?

FasteryXD:
Okay, just like before. I had already imported all the Docker images except for the .io files, which were still there. But after a few updates, everything except Redis and DB is back under importable.
Here's an example from Homarr:

<img width="777" height="496" alt="Image" src="https://github.com/user-attachments/assets/77f8d05f-43f1-42f3-8a67-4835a3bce17d" />

This is the configuration that was created during the first import. I haven't actually changed anything.

<img width="777" height="232" alt="Image" src="https://github.com/user-attachments/assets/d73b0587-e6c8-4a7c-8b7f-7bae54d38e06" />

Nevertheless, as I said, it is now listed as importable...

<img width="772" height="479" alt="Image" src="https://github.com/user-attachments/assets/f70355d0-da84-49b5-8609-ef3a191d63c1" />

And this is what the configuration would look like if I were to import it again now.

Aandree5:
[...]

Issue #3344, opened 2025-06-06
feat(release-widget): support lscr.io docker repository


FasteryXD:
If you now add all the entries using the new Release Widget feature, the default identifier for Docker Hub is "lscr.io/linuxserver/radarr," which generally leads to an error. At least for me. The correct default setting should be without "lscr.io," so for radarr: "linuxserver/radarr."

Meierschlumpf:
It seems like lscr.io is just a proxy for ghcr.io so we can probably just use the ghcr.io api later as well for this (So convert it similar to the existing ghcr.io mapping)
https://github.com/getwud/wud/issues/143#issuecomment-1233192413

Aandree5:
There's still some work to be done on #3253, but I guess it should be easy enough to add, specially if it's just a proxy for ghcr.io

Meierschlumpf:
Closed through #3607

Issue #3162, opened 2025-05-20
feat(releases-widget): add GitHub Packages Container provider


Aandree5:
> I think it would make the most sense if we change the release providers to integrations.
> 
> Having a single source of truth would be really appreciated indeed!
> 
> Quick suggestion on the "GitHub" Provider currently available, is it possible to consider making a distinction (or support it in some way) between a GitHub repo releases kind and a GitHub Container Package kind from the container registry?
> 
> If we take [this repo](https://github.com/thrnz/docker-wireguard-pia) as an example, they only publish a container image with a new tag and without a proper release. Having a way to use GHCR tags instead of releases would be awesome! 

 _Originally posted by @Larsluph in [#3137](https://github.com/homarr-labs/homarr/issues/3137#issuecomment-2894348188)_

Aandree5:
@Larsluph Thought I would do this before the fix for #3137, but that would just the migration process to the providers in integrations more complex, so this will be available at the same time as  #3137

Aandree5:
Just had a look into this, don't think we will be able to implement it. GitHub Rest API Docs seem to point to all the packages endpoint nedd to be authenticated (which wouldn't be to bad in it self) but the auth token needs `read:packages` (or similar) access to the specific package.
I cannot find anything in the docs (or elsewhere) with a different way of achieving this.
Response I got from testing: `https://api.github.com/orgs/homarr-labs/packages/container/homarr/versions`
```
{
    "message": "You need at least read:packages scope to get a package's versions.",
    "documentation_url": "https://docs.github.com/rest/packages/packages#list-package-versions-for-a-package-owned-by-an-organization",
    "status": "403"
}
```

Aandree5:
@Meierschlumpf @Larsluph With this I was thinking to save the user having to configue 2 providers for GItHub, we could check the repository releases and if none is found, check the packages container registry (if an access token has been provided), what do you think?

Larsluph:
I'm worried about the number of requests generated for this, is it cached somehow?

I feel like if you have to apply fallback on all requests there might be a lot of wasted data.

Aandree5:
I don't think it would be too bad. This would only add 1 extra request to GitHub repositories that failed to find a repository release (which if it happens you either mistyped the version filter or you would want to check the packages register).

And it would only check if the user added a `personalAccessToken` which is required for the packages.

I'm open to either way, just trying to think what would be the pros and cons of each.

Meierschlumpf:
[...]

PR #3607, opened 2025-07-11
feat(releases-widget): add new providers, Github Packages, linuxserver.io and Quay

Aandree5:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

 ---
   
**Waiting on Quay icon [#1655](https://github.com/homarr-labs/dashboard-icons/issues/1655)**

- [x] Add Integrations
    - [x] GitHub Packages
    - [x] linuxserver.io (lscr.io)
    - [x] Quay
- [x] Import form docker
    - [x] linuxserver.io (lscr.io)
    - [x] Quay
- [x] Test providers
    - [x] GitHub Packages
    - [x] linuxserver.io (lscr.io)
    - [x] Quay

fixe #3162 #3344 #3350


packages/widgets/src/_inputs/widget-multiReleasesRepositories-input.tsx
```diff
@@ -613,20 +613,18 @@ const RepositoryImportModal = createModal<RepositoryImportProps>(({ innerProps,
...
-        const providerKey = source in containerImageToProviderKind ? containerImageToProviderKind[source] : "dockerHub";
-        const integrationId = Object.values(innerProps.integrations).find(
-          (integration) => integration.kind === providerKey,
-        )?.id;
-
-        const [identifier, version] = identifierImage.split(":");
-
-        if (!identifier || !integrationId) return acc;
+        const providerKeyList = containerImageToProviderKind[source] ?? ["dockerHub"];
+        const integrationId = findIntegrationId(innerProps.integrations, providerKeyList, identifier);
```
Aandree5: @Meierschlumpf Think this is almost done but I'm having a small issue with this part. I know this is wrong and won't work, just pushed to better show what I'm after.
Following on our previous https://github.com/homarr-labs/homarr/issues/3162#issuecomment-2955493887, of automatically checking if a release exists for Github and if it doesn't then use Github Packages, how could I best do this?

Trying to call the releases route will not work in this way, but do you see a way I could achieve this?

An example:
From the container image we have the source as ghrc.io then I use a map to the integration  "ghcr.io": ["github", "githubPackages"], but from here I would need to link the kind to an actual Id to test if I get a response if not, then move to the next, and if the last then no need to test.

This would be the idea, but calling a hook in a for loop or even inside another hook is a no-go, so I'm a bit lost on this. The only way I see this is to split it into a couple phases, parse each image source into all possible integration Ids (non-last options), then test all in a single call to the releases router, and from there link it all toghether, all that only have one option, then use that, when there's multiple test all non-last ones, if there's a response from those use that, if not use the last option.

Aandree5: @Meierschlumpf  Maybe I should just default to `Github` and in the rare event that no realeases exist the user can change to `Github Packages`?

Meierschlumpf: Hmm I see, yeah maybe just default to GitHub, I think that is a lot easier and still allows the user to use the other release

Aandree5: We can always look into it in the future, can't think of a way that would not involve a lot of changes and time, for a feature which I'm not sure would even be that much used. I will then leave it as `Github` which sould be most releases anyway, and the user can change to the packages.

Aandree5:
Think this is ready for review, but it's still needing the Quay icon added [#1655](https://github.com/homarr-labs/dashboard-icons/issues/1655)
I've added a placeholder link which should match, but I can test before merge

Aandree5:
@Meierschlumpf  Also, why is deepsource complaining about the `quay-integration.ts` and not `codeberg-integration-ts` which have the exact same code? Am I missing somehting?

Meierschlumpf:
The deepsource issue is not popping up for codeberg, because I've already ignored it. We don't do what it thinks and therefore don't need to have the suggested structure

manuel-rw (dismissed):

Meierschlumpf (approved):

PR #3643, opened 2025-07-17
feat: support postgresql database

### LINKED ISSUE #2100: feat: support postgresql database

Same as mysql we can add support for postgresql

### Additional information

_No response_

xoxyuxu:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Closes #2100


xoxyuxu:
This code will not pass the lint check.
For example, in the following code, db is a union type of MySql and Postgresql, so typecheck will determine it as MySql, resulting in a violation of the argument type.

FILE: packages/api/src/router/integration/integration-router.ts
```ts
      await handleTransactionsAsync(ctx.db, {
        async handleAsync(db, schema) {
          await db.transaction(async (transaction) => {
            await transaction
              .delete(schema.integrationGroupPermissions)
              .where(eq(schema.integrationGroupPermissions.integrationId, input.entityId));
```

By preparing a function that determines the type at runtime and branching the type even with the same code, it will be possible to pass lint. It is not possible to pass lint by determining the type based on environment variables.
What is the best way to deal with multiple Async DBs?

xoxyuxu:
failing check in DeepSource...

Why mysql is not detected ?
It looks like same. And same variable name is exported in "schema/postgresql.ts".

```ts
import * as postgresqlSchema from "../schema/postgresql";
import * as mysqlSchema from "../schema/mysql";

      const mysqlTable = mysqlSchema[tableName];
      const postgresqlTable = postgresqlSchema[tableName];
```

manuel-rw:
At the moment, most of us are on holiday, so we'll review this when we come back. Thanks for your understanding.

xoxyuxu:
Thank you for the holiday info. Have a great holiday!

xoxyuxu:
Hmm… I tried to avoid a `pnpm lint` warning.
Writing code to access DB asynchronously is too redundant.
The drizzle methods have the same name, so I think the code for MySQL and PostgreSQL will be the same as long as it passes type checking...
I'm newbies for TypeScript :|

xoxyuxu:
sorry, I found my mistake...
I'm going to fix it, and update with docker-component.yml for test.

packages/db/collection.ts
```diff
@@ -67,14 +43,17 @@ export const createDbInsertCollectionForTransaction = <TTableKey extends TableKe
...
-      if (isMysql()) {
-        await insertAllMysqlAsync(db as unknown as HomarrDatabaseMysql);
-      } else if (isPostgresql()) {
-        await insertAllPostgresqlAsync(db as unknown as HomarrDatabasePostgresql);
-      } else {
-        throw new Error(`Unsupported DB_DRIVER: ${env.DB_DRIVER}`);
-      }
+    // We allow any database that supports async passed here but then fallback to mysql to prevent typescript errors
+    insertAllAsync: async (db: HomarrDatabaseMysql | HomarrDatabasePostgresql) => {
+      const innerDb = db as HomarrDatabaseMysql;
```
xoxyuxu, line 48: Why can type assertions be used with HomarrDatabasePostgresql?

If there is no difference in the keys between HomarrDatabaseMysql and HomarrDatabasePostgresql, can we just use HomarrDatabaseMysql as the same type?

How do I think about type safety for typescript...

Meierschlumpf: In this case I just allow the passing of both db types so the consumer does not need to think about converting them. Then internally in this implementation I convert it to one of the two so I don't get into conflicts (because the type seems to complex for both to work)
And as both have the "same" api, I just convert it.

It can be used here, because it's already one of the types

xoxyuxu: Thank you for your reply.
It seems that type checking is a loose constraint, like casting in C language. I understood that I just need to make the check pass with a type assertion.
The basic operation seemed to be working at runtime, so I was having trouble with the type checking. Thank you very much.

Meierschlumpf: Yeah, most of the times you don't want to do casting, because you are probably doing something wrong, but in edge cases where you just "want it to work", you have to do some type overrides to make it happy 😉

manuel-rw (dismissed):

ajnart (dismissed):
I don't have any code-related feedback on this since @manuel-rw already gave the OK. 

Not sure why we'd keep the support of mysql since afaik postgres is better is almost every way.
We could benefit more from forcing people to move to pg and later down the line supercharging it with plugins. 

- I'm not sure the tradeoff of maintaining support for all db types is worth it. Might be shooting ourselves in the foot

packages/db/schema/mysql.ts
```diff
@@ -528,6 +528,7 @@ export const userRelations = relations(users, ({ one, many }) => ({
   groups: many(groupMembers),
   ownedGroups: many(groups),
   invites: many(invites),
+  medias: many(medias),
```
ajnart, line 531: was this forgotten before ? 😆

Meierschlumpf: Yes it was, but not really a problem as the types of sqlite are used, where those relations exist

Meierschlumpf:
> I don't have any code-related feedback on this since @manuel-rw already gave the OK.
> 
> Not sure why we'd keep the support of mysql since afaik postgres is better is almost every way. We could benefit more from forcing people to move to pg and later down the line supercharging it with plugins.
> 
> * I'm not sure the tradeoff of maintaining support for all db types is worth it. Might be shooting ourselves in the foot

I see your point regarding the effort to maintain all the different db types, however removing support for mysql
1. would result in a breaking change
2. would break the setup of existing users and not really leave them with a way to move to another db type (for example pg)
3. would result in less adoption (or people asking for mysql support)

Additionally I think the effort to keep it synchronized is not that big and at some point we might be able to replace it with a better abstraction, so we don't have to specify the special use cases for transactions etc.

So do you agree, that we move on without removing mysql? @ajnart

ajnart:
Yeah I think you're right @Meierschlumpf it would be much much more painful to make the switch than the keep maintaining both 

It would be nice to have some docs about how to use an external db if that's something that is supported

Meierschlumpf:
> Yeah I think you're right @Meierschlumpf it would be much much more painful to make the switch than the keep maintaining both
> 
> It would be nice to have some docs about how to use an external db if that's something that is supported

Right now we only have the available environment variables listed: https://homarr.dev/docs/next/advanced/environment-variables/#database
I've created an issue: https://github.com/homarr-labs/documentation/issues/372

Meierschlumpf (approved):

PR #3676, opened 2025-07-22
feat(releases-widget): add `Complete` action to mark releases as checked

### LINKED ISSUE #3324: Widget: Add option to mark watched repositories as "done"

Currently you can monitor Docker/Github/etc. Repositorys to get an Info when a new version arrives.
It would be create if I could mark this as "Done" so it wont show anymore after I updated the container/etc. 

Image

This would replace the need auf Diun or Watchtower for me 

### Additional information

_No response_

Aandree5:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

---
image

fixes #3324


Aandree5:
@Meierschlumpf Would using `localstorage` be ok here, or do you think some package would be better?

Meierschlumpf:
I think localstorage is a good first start, we might find other ways to store it in the future, however I don't know which yet. So we should just use localstorage for now yes.

manuel-rw (approved):
IMO local storage should only be temporary solution, better store it in the DB

Aandree5:
> IMO local storage should only be temporary solution, better store it in the DB

@manuel-rw I do agree with this, would like to be able to use this feature from any device, how do you think this would be best? Would it have it's own table? Or would/should we reuse the dashboard item JSON value for that given widget?

Meierschlumpf:
I think we would probably need "widget options per user", where we can store which versions were already seen by a user. So a new table with a reference to item and user and a json blob that somehow can be defined for each widget.
Anonymous users would then obviously still have it just in localstorage. I think for now localStorage is good enough

Aandree5:
> I would suggest that we change the "Complete" to something else that is more like "Mark as viewed" or similar. Because you do not really complete the release

@Meierschlumpf I agree with that name change, wasn't sure which to pick. I've also recently learnt of Mantine's `localsotrage` hook, so I've made a change to use it, feel like it works better (even syncs between tabs), let me know what you think.

Done with 6568789797551567c1b398fc5b00fd753996daee

Issue #3622, opened 2025-07-13
bug: app ping not loading after 1.28


steven-aj:
### Describe the bug

Status checks are no longer working as of the recent update and are stuck in a "Loading" state with a blue indicator.

### Steps to reproduce

- From a prior version, update to Homarr Version 1.28.1
   - `docker compose up -d --pull=always`
   - `docker compose restart`
- navigate to Dashboard with apps pinned
- activate "Enable status check" on an app

### Impact

Minor impact.

Not unusable, but Homarr helps me keep track of potential crashes from the dashboard. Frustrating, if anything.

### Additional information

<img width="313" height="151" alt="Image" src="https://github.com/user-attachments/assets/6cac1cc1-ce7c-42c2-933c-5dbc9906b8f0" />

### Version

1.28.1

### Installation method

Docker Compose

### Browser

Firefox

manuel-rw:
Hi, can you perform these troubleshooting steps?
https://homarr.dev/docs/1.18.0/management/integrations/#troubleshooting-connection

ydrob:
The same  problem on the latest (v1.28.1), on version v1.27.0 works well

steven-aj:
> Hi, can you perform these troubleshooting steps? https://homarr.dev/docs/1.18.0/management/integrations/#troubleshooting-connection

---

 > Check whether you followed the Homarr documentation and didn't skip any steps. This is the most often mistake that create many noisy support requests and following all steps will often resolve the problems.

Does not apply. This is not a new installation of Homarr. "Status Check" worked prior to recent update.

> Check Homarr's logs for more information. As we have no control over breaking changes in upstream APIs and third party software, they can break without notice in unexpected ways which can also lead to errors within Homarr. In such case, these occurrences must be reported and fixed within the source code of Homarr.

Does not apply. Status checks are irrelevant to upstream APIs.

> Check whether the host can reach the destination hostname / IP using ping and curl or wget afterward. Execute this command on the host machine, not inside containers, to check whether it can connect or not.

Homarr's host is the same host as other services. Status Check fails for both hostname _and_ local IP

> Check whether the container / pod can reach the destination hostname / IP. Execute the same commands in the shell inside the container / pod.

Does not apply. See above.

> Check your reverse proxy / authentication proxy logs for failures / hints

Does not apply to status checks

> Check your security log of tunnels and VPNs

Does not apply. Irrelevant.

> Check your router for security events

Does not apply. Irrelevant.

[...]

PR #3680, opened 2025-07-23
revert: "feat(ping): ignore certificate error and show request durati…

manuel-rw:
This reverts commit 1eb47311fa81184bd67830d0669d9f87d9922e70.

<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

#3622

image


Aandree5 (commented):
@manuel-rw as this is just reversing commit do we just ignore the lint issues, or will those get resolved?

packages/certificates/src/server.ts
```diff
@@ -5,8 +5,9 @@ import type { AgentOptions } from "node:https";
 import { Agent as HttpsAgent } from "node:https";
 import path from "node:path";
 import { checkServerIdentity, rootCertificates } from "node:tls";
+import { URL } from "url";
```
Aandree5: This is not used?

packages/certificates/src/server.ts
```diff
@@ -5,8 +5,9 @@ import type { AgentOptions } from "node:https";
...
 import axios from "axios";
-import { fetch } from "undici";
+import { fetch, RequestInfo, RequestInit, Response } from "undici";
```
Aandree5: Type only imports, should be set to type?

packages/ping/src/index.ts
```diff
@@ -21,14 +21,22 @@ export const sendPingRequestAsync = async (url: string) => {
  * https://stackoverflow.com/questions/46946380/fetch-api-request-timeout
  * @param param0 fetch arguments
  * @returns fetch response
+ * @param rejectUnauthorized whether untrusted certificates should be ignored
  */
-export const fetchWithTimeoutAndCertificates = (...[url, requestInit]: Parameters<typeof fetch>) => {
+export const fetchWithTimeoutAndCertificates = (
+  rejectUnauthorized = false,
```
Aandree5: Should both `fetch` methods default to true? Or is the goal to accept any certificate by default?

manuel-rw: No, only ping should bypass any checks

Aandree5: Yes of course, I interpreted it worng, read it as "reject unauthorized (untrusted) = no", just re-read the docs description. All good then.

Aandree5: @manuel-rw  Sorry to keep on this but I'm honestly just trying to understand it, 
So `rejectUnauthorized  = true` is to ignore, and `rejectUnauthorized = false` to not ignore, right?

Because if so, `sendPingRequestAsync()` is passing in `false`,

manuel-rw: ``rejectUnauthorized  = true`` means that certificates are checked and invalid certificates are rejected.
``rejectUnauthorized  = false`` means that certificates are ignored.
Hence ``sendPingRequestAsync()`` is passing ``false`` -> we want to ignore the certs

Aandree5: This is was I first thought, I think I know where my confusion was, this is in the `ping` package! So it will only relate to ping where we want to ignore certs.

Like I said, I will approve this in a bit I think it's quite an importante release to have ready for tomorrow (so I don't want to raise another comment for a tiny thing that may be nothing) but just want to mention, in the `certificates` package, even though the create agent function default the `rejectUnauthorized` to true, it's called from the fetch in the same package that defaults to false (if that makes sense 😂 ). I think this question was more thinking on case

Aandree5 (approved):

packages/ping/src/index.ts
manuel-rw: I think you catched a mistake, the default for ``fetchWithTrustedCertificatesAsync`` was wrong. Can you re-review carefully?

Aandree5: I will give it another look. I did check it carefully and all looked good. I was just refering to these two `fetch` statement bu probably didn't explain very well.

packages/ping/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "dependencies": {
+    "@homarr/certificates": "workspace:^0.1.0",
+    "@homarr/common": "workspace:^0.1.0",
+    "@homarr/log": "workspace:^0.1.0"
+  },
+  "devDependencies": {
+    "@homarr/eslint-config": "workspace:^0.2.0",
+    "@homarr/prettier-config": "workspace:^0.1.0",
+    "@homarr/tsconfig": "workspace:^0.1.0",
+    "eslint": "^9.30.1",
```
Aandree5:
```suggestion
    "eslint": "^9.32.0",
```

I'm sure you seen it, bump eslint to match other `package.json`

Meierschlumpf:
@manuel-rw can you check if my changes are ok? I've tested them and they worked (I only changed what we return and moved the timeout stuff / unauthorized stuff around) so nothing should that needs to be tested by the users I would say

manuel-rw:
lgtm, can we merge?

Aandree5 (approved):

PR #3792, opened 2025-08-05
Ports: Allow custom web and websocket ports

tebriel:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


Meierschlumpf:
Hey, welcome and thank you for your contribution. One thing that would also need to change is the nginx configuration which then would need to reuse the defined ports for forwarding. Is there a reason you want the internal ports customizable? It's generally recommended to use the nginx proxy when hosting Homarr.

manuel-rw:
I don't see the use-case for this. Can you explain @tebriel ?

tebriel:
@manuel-rw @Meierschlumpf I deploy on bare metal and already have things using 3000. Happy to update nginx to use something like envsubst to do so. I understand that since that's not a supported deployment methodology you might not accept this PR. Before I continue and update NGINX lmk if this works for you or not!

manuel-rw:
I understand. Bare metal isn't something we really want to support; it usually is more work than running a virtualized or containerized version and the upkeep is bigger and more complicated (e.g. resolve merge conflicts, pull frequently, install dependencies, handle upstream breaking changes, ...).
IMO we could accept this, but instead of ``WEBSOCKET_PORT`` we could rename to ``UNSAFE_WEBSOCKET_PORT`` or ``DANGEROUS_WEBSOCKET_PORT`` to signal that this isn't a common use case.

Please also update https://homarr.dev/docs/getting-started/installation/source and https://homarr.dev/docs/advanced/environment-variables/ and part of your PR: https://github.com/homarr-labs/documentation

tebriel:
@manuel-rw thanks for working with me here. Open to any feedback, this is a pretty complex project so I might've missed some nuance!

PR #2807, opened 2025-04-05
ci(helm-docs): deploy helm chart and update the documentation

oben01:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

**Thank you for your contribution. Please ensure that your pull request meets the following pull request:**

- [ ] Builds without warnings or errors (``pnpm build``, autofix with ``pnpm format:fix``)
- [ ] Pull request targets ``dev`` branch
- [ ] Commits follow the [conventional commits guideline](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] No shorthand variable names are used (eg. ``x``, ``y``, ``i`` or any abbrevation)

PR #3881, opened 2025-08-17
ci: add on-release workflow to trigger docs-release, update bug-report-template and update security-policy

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

See #3880 as reference for successful update of template, also https://github.com/homarr-labs/documentation/actions/runs/17021661509/job/48251545027 was triggered from local run on the trigger-docs-release job. Update of security-policy was not tested with whole workflow and instead only locally as it consists out of the same workflow (except commit messages etc.) as bug-report-template


SECURITY.md
```diff
@@ -1,18 +1,31 @@
...
+    </tr>
+    <tr>
+        <td>&gt;1.33.0</td>
+        <td>✅</td>
+    </tr>
+    <tr>
+        <td>&lt;1.33.0</td>
+        <td>❌</td>
+    </tr>
+</table>
```
ajnart: You could just remove this and say only the latest version is the one that is supported for security updates, we won't re-release old versions to patch security vulnerabilities. (Since I'm pretty sure almost no-one has their docker version to something fixed to something specific other than `latest`)

.github/workflows/on-release.yml
ajnart, line 1: nit: I saw in an other PR that `gh cli` tokens can have some kind of scope assigned to them. It might be useful to restrain that scope so that incoming PRs can't exploit the token we have in place to do anything outside of the allowed scope

I'm not 100% sure it can be exploited but I saw some articles about secrets leaking a while back

ajnart (dismissed):

SECURITY.md
Meierschlumpf: Hmm I agree, I think it would make sense if we just write that we support the **latest stable version**. Because when I think of for example beta releases or similar it would also result in updates (at least in the current configuration) which does not really make sense.

I'll adjust it

.github/workflows/on-release.yml
Meierschlumpf: Not sure if you can specify a scope for gh cli, but at least you can configure the permissions of the app. Theoretically it is also possible to restrict those permissions within the workflow, however I was not able to get it working.

Therefore I would suggest that we'll leave it that way for now and migrate to https://github.com/actions/create-github-app-token with an easier permission handling system in another issue (#3892)

ajnart: It's this comment by the GH security bot I was referring to https://github.com/homarr-labs/homarr/pull/2807#discussion_r2100160597

Not sure if it's exactly relevant in our case. If not I don't think my previous comment was valid, because surely GitHub would've flagged it as they did for the above comment (so no need to switch to something else)

ajnart (approved):

.github/workflows/on-release.yml
Meierschlumpf: Oh I see, yeah we have some of them still open. (see https://github.com/homarr-labs/homarr/security/code-scanning) It's just saying that we should consider specifying the permissions explicitly to only give the permissions to the normal github token that are required. As we are using app tokens for most of them the default there would be the permissions we gave to the github app. However as said it would be possible to restrict it there as well 👍🏼

Issue #2475, opened 2025-03-02
feat: add caldav / webdav calendar integration


Meierschlumpf:
### Describe the feature you'd like to request

Show events from external calendars in the calendar widget

### Additional information

Originally posted https://github.com/ajnart/homarr/issues/412

tylertylerday:
+1 Would love to be able to see all my meetings/cal events with the ability to join the zoom/gmeet meeting!

NicolasNewman:
@Meierschlumpf I'd love to see this functionality added. Is this something I could take on and implement myself? I was able to quickly get .ical links from Google Calendar loaded into the widget but I wanted to double check before I proceed any further.

Meierschlumpf:
Sure, feel free to create a pull-request and if you are stuck, to ping me with questions

NicolasNewman:
I have a [branch](https://github.com/homarr-labs/homarr/compare/dev...NicolasNewman:homarr:calendar-ical-support) with my changes so far. I'm currently implementing it to fetch an ical file from a url (testing it with Google Calendar at the moment). I personally don't use a caldav/webdav server so I have little experience with them. If they don't support getting the ical files through a get request, additional work will need to be done on a per-server basis.

I re-structured the CalendarEvent interface since its currently tightly coupled to showing events for media releases. If any of my current changes stand out, let me know and I can open a draft PR if you'd prefer to have the discussion there.

My main question is regarding the [NotificationIndicator](https://github.com/NicolasNewman/homarr/blob/52d18dc1327b277ea3c49f2b7f939cd827682f8b/packages/widgets/src/calendar/calender-day.tsx#L81) component. It only appears to render when `links` are defined in a CalendarEvent. `notificationColor` is the only field that is optional but I don't need any of the others if there is no need for a redirect button. Is there a reason why `notificationColor` was tightly coupled to links? It seems like those fields serve two different purposes. Not sure what the best approach to generalizing it would be.

(on a side note, [ical.js](https://github.com/kewisch/ical.js) is what I'm currently using to parse ical files but the auto-generated TS definitions aren't great. I'm planning on revisiting options once I have a functional prototype I'm happy with).

PR #3980, opened 2025-09-02
feat(integrations): add ICal

NicolasNewman:
# ICal integration for Calendar widget

## Design Decisions
* Integration loads ICal events from a fetch request to an endpoint supplying an ical file
    * i.e: `https://calendar.google.com/calendar/ical/my-secret-key/basic.ics`
* [ical.js](https://github.com/kewisch/ical.js) has been added to support parsing ical files
* To better generalize `CalendarEvent`, a metadata property has been added which can be differentiated by its `type` property
    * Fields formerly in  `mediaInformation` have been moved into `metadata` where type is `"audio" | "video" | "tv" | "movie"`
    * ICal event metadata is stored in `metadata` where type is `"event"`

## Additional Comments
* The typings for ical.js are auto generated and rely on many any/generic Object fallbacks. 5 lines needed to have the linter ignored.
* Documentation & an icon for the integration still need to be added. I'll open a PR into those repos once there isn't any major refactoring left.
* Only tested with Google Calendar. I don't personally use a caldav/webdav server so I can't test it myself. In theory as long as they expose an endpoint to a ical compliant file they will work.
* To be able to use `notificationColor` as is, I made `href` in the `links` property optional and only display the button when `href` is defined. I didn't want to change the underlying data structure to much without any input but ideally it should be made more dynamic.

## Demo
Screenshot from 2025-09-01 18-29-36

Addresses: #2475


Meierschlumpf:
Hey, thank you for the contribution ❤️ 

I see a few security concerns that we'll need to address.
1. the url is not encrypted in the db, so if somebody has access to the db (or the db file for sqlite) they can read the security-token in plaintext
2. dispatched requests are logging the url requested with LoggingAgent

Not quite sure how we should solve them just now

NicolasNewman:
I took a quick look at the LoggingAgent and I see what you mean. I'm assuming (1) isn't too tricky to address as part of/the whole url can be moved to a secret. For (2), the variability of URLs between services makes it a lot harder to redact in the dispatcher.

In theory, any server providing ical files should have the content-type set to `text/calendar` in accordance with [RFC5545 s-8.1](https://datatracker.ietf.org/doc/html/rfc5545#section-8.1). One possible solution for (2) may be to verify the provided URL has content-type=text/calendar, and then in the LoggingAgent, add a branch for that content type and remove everything in between the domain and file name.

Meierschlumpf:
1. Yeah I think we could just move the path to the secrets. That's actually not that bad of an idea 👍🏼 
2. I think we could just use something like if there are more or equal to 32 characters in the path part we redact it, otherwise we leave it as is.

Feel free to implement it that way 👍🏼

NicolasNewman:
I started the implementation of moving the url to secrets by creating a new `url` secretKind and modifying the integration-new-form component to hide the non-secret url form field when a url secret kind is present. Since the non-secret url field is required, it gets auto-populated with the hostname of the secret url (i.e. https://calendar.google.com/). Let me know if you have any issues with that approach. If not, I'll implement that functionality for edit as well.

apps/nextjs/src/app/[locale]/manage/integrations/new/_integration-new-form.tsx
```diff
@@ -55,12 +55,18 @@ const formSchema = integrationCreateSchema.omit({ kind: true }).and(
...
   const t = useI18n();
   const secretKinds = getAllSecretKindOptions(searchParams.kind);
+  const hasUrlSecret = secretKinds.some((kinds) => kinds.includes("url"));
   const router = useRouter();
   const [opened, setOpened] = useState(false);
+
+  let url = searchParams.url ?? getIntegrationDefaultUrl(searchParams.kind) ?? "";
+  if (hasUrlSecret) {
+    url = "http://localhost";
+  }
```
Comment (Meierschlumpf) [inline apps/nextjs/src/app/[locale]/manage/integrations/new/_integration-new-form.tsx]: Makes total sense how you did it 👍🏼 
I just think we should add some comments to why we for example set it to `http://localhost` when it has a url secret (which of course is because we still need to pass validation)

Thank you for the work. I like it 😄

Comment (NicolasNewman) [reply apps/nextjs/src/app/[locale]/manage/integrations/new/_integration-new-form.tsx]: Added the comment and also implemented the functionality for the edit form

Comment (Meierschlumpf) [reply apps/nextjs/src/app/[locale]/manage/integrations/new/_integration-new-form.tsx]: Okay nice, I'll give it a review later today and will try it out 👍🏼

packages/definitions/src/integration.ts
```diff
@@ -283,6 +284,14 @@ export const integrationDefs = {
     category: ["notifications"],
     documentationUrl: createDocumentationLink("/docs/integrations/ntfy"),
   },
+  ical: {
+    name: "iCal",
+    secretKinds: [["url"]],
+    // TODO(nn): valid icon
+    iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons@master/svg/ical.svg",
+    category: ["calendar"],
+    documentationUrl: createDocumentationLink("/docs/integrations/ntfy"),
```
Meierschlumpf: Has to be changed & documentation has to be added

packages/definitions/src/integration.ts
```diff
@@ -283,6 +284,14 @@ export const integrationDefs = {
     category: ["notifications"],
     documentationUrl: createDocumentationLink("/docs/integrations/ntfy"),
   },
+  ical: {
+    name: "iCal",
+    secretKinds: [["url"]],
+    // TODO(nn): valid icon
+    iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons@master/svg/ical.svg",
```
Meierschlumpf, line 290: Icon missing (see todo)

packages/translation/src/lang/en.json
```diff
@@ -945,6 +945,10 @@
           "label": "Topic",
           "newLabel": "New topic"
         },
+        "url": {
+          "label": "Url",
+          "newLabel": "New URL"
```
Meierschlumpf: Should use same casing for URL/Url

Meierschlumpf:
I took your pr as an opportunity to change the `CalendarEvent` object to a more generic structure, let me know if you like it that way. Feel free to suggest improvements / adjustments to my changes.

Meierschlumpf:
You can ignore the stability-days workflow. However it seems like we've broken the fetch agent tests, it seems like our current conversion is not quite right for the path

packages/definitions/src/integration.ts
NicolasNewman: Opened a PR to add the icon: [dashboard-icons#2035](https://github.com/homarr-labs/dashboard-icons/pull/2035)

Meierschlumpf: I've merged it, now it works 👍🏼

packages/definitions/src/integration.ts
NicolasNewman: I opened a PR to add documentation: [documentation#404](https://github.com/homarr-labs/documentation/pull/404)

NicolasNewman:
I fixed some of the tests. Having a hard time with the others since they pass when I run them individually through VSCode but not with the root test script. `Nzbget integration > Delete item should result in empty items` specifically I'm not sure why it would fail one way and not the other. 

Do you happen to use the Vitest extension through VSCode? I might have something misconfigured on my end.

Meierschlumpf:
Oh sorry regarding the nzbget test it is quite flaky and also fails sometimes in ci. So you can ignore it if you didn't make any changes to it 👍🏼

Meierschlumpf (approved):

Meierschlumpf:
Thank you again ❤️  for implementing this feature! I'll do some last manual tests with different calendar integrations (just to be sure it still works like it did previously) and then we'll merge this for the upcoming release on Friday.

Meierschlumpf:
Okay checked with ical (outlook), nextcloud and sonarr and all of them worked as expected 👍🏼

NicolasNewman:
Glad to hear everything looked good on your end! Looking forward to using this next release.

PR #3983, opened 2025-09-02
fix(lang): correct en-GB typos in application and integration labels and messages

ChristopherBull:


ChristopherBull:
This was only supposed to be a simple typo PR (homarr-labs/homarr@e935509e88f7be553baeb60eac7a99beeeda7e45)

ChristopherBull:
VS Code auto-created PR against `main` instead of `dev`. Correct PR is now https://github.com/homarr-labs/homarr/pull/3984

Issue #4091, opened 2025-09-16
bug: Notification "time ago" timestamps not updating


castielwaverly:
### Describe the bug

After setting up a notification widget with an active integration on a board, all notifications eventually say "a few seconds ago" until the page is refreshed.

There's an expected 30s refresh interval on the "time ago" timestamp:
https://github.com/homarr-labs/homarr/blob/b8f0d9a21f37f7d92a940149c7091d01e12ad72f/packages/widgets/src/notifications/component.tsx#L95-L106
https://github.com/homarr-labs/homarr/blob/4287d1743b2e90c7957b9c74dcd975ab3d494112/packages/common/src/hooks.ts#L13-L23

The logic is correct, since refreshing the page results in an accurate "time ago" timestamp, but the interval must be getting lost somehow, since it never actually updates every 30s on the page. 

### Steps to reproduce

1. Create a new board.
2. Create a notification integration (I used ntfy).
3. Add notifications widget to new board.
4. Generate some notifications or wait for some to be generated.
5. Wait at least 30s. Preferably longer to confirm behavior.
6. Observe the timestamps on the notifications.
7. Refresh the page.
8. Observe the timestamps on the notifications.

#### Expected
Timestamps observed in (6) are updated at an interval of 30s to reflect the correct amount of time since they came in.

#### Actual
Timestamps observed in (6) stay at their original value (i.e. "a few seconds ago" for new notifications) indefinitely.

### Impact

Minor aesthetic bug. Refreshing "fixes" timestamps.

### Version

1.37.0

### Installation method

Docker Compose

### Browser

Firefox

castielwaverly:
Tracked down the source of the bug. When you configure an aggressive (i.e. <30 seconds) Notification Updater frequency, the subscription hook below ends up causing a re-render before the interval gets triggered. 

https://github.com/homarr-labs/homarr/blob/312e084e2c731ce1bff9c9dc3e1d8f7a88df5429/packages/widgets/src/notifications/component.tsx#L29-L51

The reason the timestamp is "stuck" despite the constant re-renders, is that state only initializes on mounts and not on renders, so the initial value persists throughout the entire life of the component. 

I've updated the steps to reproduce accordingly, because I still think this is a bug. I think it's easy to "just set a less aggressive Notification Updater frequency", but now there are options in the interface that are invalid. Probably recalculating "time ago" on re-renders is the way to go.

castielwaverly:
PR is up: #4107

PR #4107, opened 2025-09-19
fix: Recalculate time ago on re-render

castielwaverly:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

### Checklist

### Description

Fixes:
- #4091 
- Unnecessary recalculations of "time ago"

When you set an aggressive enough Notification Updater frequency, your notification timestamps get stuck in their initial state. The parent `<NotificationsWidget />` component subscribes to notification updates, so it will necessarily re-render at least as frequently as your Notification Updater configuration. This means the `useTimeAgo` hook will always clear the update interval before it performs the update in these scenarios.

### DIFF
--- a/packages/common/src/hooks.ts
+++ b/packages/common/src/hooks.ts
@@ -11,13 +11,17 @@ const calculateTimeAgo = (timestamp: Date) => {
 };
 
 export const useTimeAgo = (timestamp: Date, updateFrequency = 1000) => {
-  const [timeAgo, setTimeAgo] = useState(calculateTimeAgo(timestamp));
+  const [timeAgo, setTimeAgo] = useState(() => calculateTimeAgo(timestamp));
+
+  useEffect(() => {
+    setTimeAgo(calculateTimeAgo(timestamp));
+  }, [timestamp]);
 
   useEffect(() => {
     const intervalId = setInterval(() => setTimeAgo(calculateTimeAgo(timestamp)), updateFrequency);
 
     return () => clearInterval(intervalId); // clear interval on hook unmount
-  }, [timestamp]);
+  }, [timestamp, updateFrequency]);
 
   return timeAgo;
 };


packages/common/src/hooks.ts
```diff
@@ -11,13 +11,17 @@ const calculateTimeAgo = (timestamp: Date) => {
 };
 
 export const useTimeAgo = (timestamp: Date, updateFrequency = 1000) => {
-  const [timeAgo, setTimeAgo] = useState(calculateTimeAgo(timestamp));
+  const [timeAgo, setTimeAgo] = useState(() => calculateTimeAgo(timestamp));
```
castielwaverly, line 14: While not necessary for this fix, this will reduce recalculations. Code immediately inside `useState(...)` gets evaluated every time, but the state is only initialized once per mount; this way, the calculation is only done on mount.

packages/common/src/hooks.ts
```diff
@@ -11,13 +11,17 @@ const calculateTimeAgo = (timestamp: Date) => {
...
+  useEffect(() => {
+    setTimeAgo(calculateTimeAgo(timestamp));
+  }, [timestamp]);
 
   useEffect(() => {
     const intervalId = setInterval(() => setTimeAgo(calculateTimeAgo(timestamp)), updateFrequency);
 
     return () => clearInterval(intervalId); // clear interval on hook unmount
-  }, [timestamp]);
+  }, [timestamp, updateFrequency]);
```
castielwaverly, line 24: Not part of the fix, but it's a good practice to list all dependencies here unless there's a specific reason not to.

Meierschlumpf:
I'll take a quick look in about 20 minutes so we can put it into todays release

Meierschlumpf:
Tried it out locally with 5 second update interval and worked like a charm (a few seconds ago was updated to one minute ago etc.)

castielwaverly:
Thanks for getting to this so quickly! Looking forward to using the officially deployed image again :sunglasses:.

PR #4158, opened 2025-09-27
fix: Search multiple docker hub pages for releases

castielwaverly:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

### Checklist

### Description

Primarily adds functionality to search multiple pages on docker hub. Pages will be fetched one by one until we either find a matching release or search through "enough" pages and give up. Five pages was chosen in the current implementation, but it can be easily changed (e.g. I think `3` is a fine value too).

Additionally, the related code was refactored to reduce coupling and use types more effectively. 

#### Decoupling
Regarding the reduction of coupling, the core call structure of fetching the latest release looks like this:
```
-> releaseRequestHandler
    -> instanceA.getLatestMatchingReleaseAsync
    -> instanceB.getLatestMatchingReleaseAsync
    ...
    -> instanceN.getLatestMatchingReleaseAsync
        -> getLatestRelease
```
Currently, every instance of `getLatestMatchingReleaseAsync` and  `getLatestRelease` are maintaining a return structure and juggling codes on behalf of `releaseRequestHandler`. This makes it harder to make changes to this flow and keep behavior consistent across integrations (e.g. when is an error [an error](https://github.com/homarr-labs/homarr/blob/2e56d9a6c760e90896bc543fe84287284a48474c/packages/integrations/src/github-container-registry/github-container-registry-integration.ts#L96-L99) and when is it ["noReleasesFound"](https://github.com/homarr-labs/homarr/blob/2e56d9a6c760e90896bc543fe84287284a48474c/packages/integrations/src/gitlab/gitlab-integration.ts#L86-L89)?)

The change here keeps the behavior the same, but it centralizes the construction of the final response and some of the repeated handling logic in the top-level `releaseRequestHandler`. It's simpler to align everything now (e.g. an error is always an error) if there is desire for that. 

#### Typing improvements
Regarding the typing, the main characters [here](https://github.com/homarr-labs/homarr/blob/dev/packages/integrations/src/interfaces/releases-providers/releases-providers-types.ts) have been changed to reflect how the app behaves. For example, in the current model, a `ReleasesResponse` of `{ id: "id"}` would not raise any type errors, but is clearly invalid. I believe this flexibility was created because when there is an error, none of the release fields are there, so they're "optional"; similarly, when there is a release, the error field is not there and so it is also "optional". However, you can capture this "either release details or error" behavior accurately with types.

### DIFF
--- a/packages/integrations/src/interfaces/releases-providers/releases-providers-types.ts
+++ b/packages/integrations/src/interfaces/releases-providers/releases-providers-types.ts
@@ -1,5 +1,16 @@
 import type { TranslationObject } from "@homarr/translation";
 
+export interface ParsedIdentifier {
+  owner: string;
+  name: string;
+}
+
+export interface ReleasesRepository {
+  id: string;
+  identifier: string;
+  versionRegex?: string;
+}
+
 export interface DetailsProviderResponse {
   projectUrl?: string;
   projectDescription?: string;
@@ -12,42 +23,17 @@ export interface DetailsProviderResponse {
 }
 
 export interface ReleaseProviderResponse {
-  latestRelease: string;
-  latestReleaseAt: Date;
+  latestRelease: string; // fix me
+  latestReleaseAt: Date; // fix me
   releaseUrl?: string;
   releaseDescription?: string;
   isPreRelease?: boolean;
-}
-
-export interface ReleasesRepository {
-  id: string;
-  identifier: string;
-  versionRegex?: string;
 }
 
 type ReleasesErrorKeys = keyof TranslationObject["widget"]["releases"]["error"]["messages"];
 
-export interface ReleasesResponse {
-  id: string;
-  latestRelease?: string;
-  latestReleaseAt?: Date;
-  releaseUrl?: string;
-  releaseDescription?: string;
-  isPreRelease?: boolean;
-  projectUrl?: string;
-  projectDescription?: string;
-  isFork?: boolean;
-  isArchived?: boolean;
-  createdAt?: Date;
-  starsCount?: number;
-  openIssues?: number;
-  forksCount?: number;
+export type ErrorResponse = { code: ReleasesErrorKeys } | { message: string };
 
-  error?:
-    | {
-        code: ReleasesErrorKeys;
-      }
-    | {
-        message: string;
-      };
-}
+export type DetailedRelease = DetailsProviderResponse & ReleaseProviderResponse;
+
+export type ReleasesResponse = { id: string } & (DetailedRelease | { error: ErrorResponse });
--- a/packages/integrations/src/interfaces/releases-providers/releases-providers-types.ts
+++ b/packages/integrations/src/interfaces/releases-providers/releases-providers-types.ts
@@ -23,8 +23,8 @@ export interface DetailsProviderResponse {
 }
 
 export interface ReleaseProviderResponse {
-  latestRelease: string; // fix me
-  latestReleaseAt: Date; // fix me
+  latestRelease: string;
+  latestReleaseAt: Date;
   releaseUrl?: string;
   releaseDescription?: string;
   isPreRelease?: boolean;
--- a/packages/integrations/src/linuxserverio/linuxserverio-integration.ts
+++ b/packages/integrations/src/linuxserverio/linuxserverio-integration.ts
@@ -43,7 +43,7 @@ export class LinuxServerIOIntegration extends Integration implements ReleasesPro
   public async getLatestMatchingReleaseAsync(
     identifier: ParsedIdentifier,
   ): Promise<DetailedRelease | ErrorResponse | null> {
-    const { owner, name } = identifier;
+    const { name } = identifier;
 
     const releasesResponse = await fetchWithTrustedCertificatesAsync(this.url("/api/v1/images"));
     if (!releasesResponse.ok) {
@@ -59,7 +59,6 @@ export class LinuxServerIOIntegration extends Integration implements ReleasesPro
     const release = data.data.repositories.linuxserver.find((repo) => repo.name === name);
     if (!release) {
       localLogger.warn(`Repository ${name} not found on provider, with LinuxServerIO integration`, {
-        owner,
         name,
       });
       return null;
--- a/packages/integrations/src/linuxserverio/linuxserverio-integration.ts
+++ b/packages/integrations/src/linuxserverio/linuxserverio-integration.ts
@@ -6,7 +6,11 @@ import { Integration } from "../base/integration";
 import { TestConnectionError } from "../base/test-connection/test-connection-error";
 import type { TestingResult } from "../base/test-connection/test-connection-service";
 import type { ReleasesProviderIntegration } from "../interfaces/releases-providers/releases-providers-integration";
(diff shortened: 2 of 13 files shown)


castielwaverly:
Bug created for visibility:
- https://github.com/homarr-labs/homarr/issues/4172

Open to any feedback and changes, as I realize I did a bit more than just add pagination to the search.

Meierschlumpf:
Hey, thank you for your contribution. I agree that the current implementation is not great, but I would suggest that we stick with the implementation fully within the integrations and just improve how we deal with the types and maybe add some reusable helpers.

For example I would suggest that the main method either returns something like

```ts
type ReleaseError = {
    type: "custom";
    message: string;
  } | {type: ReleaseErrorType};
interface ReleaseData {
  ...allPropertiesWeHave
}
type Release = {
  success: true;
  data: ReleaseData;
} | {
  success: false;
  error: ReleaseError
}
```

Sorry for only answering this late, did not really find the time the last few days to work on Homarr, but I'm back 😄

castielwaverly:
> Sorry for only answering this late...

No worries. I came in uninvited and dropped a refactor as part of a fix for some very edge-case-ey behavior, so I'm thankful for any feedback :smile:

> I would suggest that we stick with the implementation fully within the integrations and just improve how we deal with the types...

Yeah sure. I think I see where you're coming from: there is currently no complexity in the top level releases request handler, so let's just keep it that way instead of "spreading" the complexity, so to speak. Would still like to clarify some things though:

1. When you say "main method", do you mean `releasesRequestHandler.requestAsync` or `ReleasesProviderIntegration.getLatestMatchingReleaseAsync`? In general, I'd like to avoid changing the actual return structure of the former to keep the changes contained. I've already leaked into the component logic by making the type more strict, so I worry changing the structure would come with additional substantial changes there.
2. Are you against _any_ logic in the top level releases request handler? For example, something like the following would keep it simple

```ts
 async requestAsync(integration, input) {
    const integrationInstance = await createIntegrationAsync(integration);
    const response = await integrationInstance.getLatestMatchingReleaseAsync({
      id: input.id,
      identifier: input.identifier,
      versionRegex: input.versionRegex,
    });
    if (!response.success) {
      return { id: input.id, error: response.error }
    }
    return { id: input.id, ...response.data, ...otherStuff }
  }
```

Meierschlumpf:
1. with main method I mean the `getLatestMatchingReleaseAsync` method.
2. I would like to not have to much logic in there as it should (at least for integrations) mostly only be caching. IMO it would make sense to return it directly in the new structure (the id can also be added in the router). Or what do you think? Then we can benefit from the types through all of it to the component

castielwaverly:
Makes sense. I should have a few hours sometime in the next few days, so I'll see what I can do.

castielwaverly:
It was my turn to have no time last week :smile:. I got a chance to move logic out of the cache layer and it ended up being cleaner than what I had, good call :+1: 

I left the `id` in the cache layer since it's already juggling `integration` in a similar way. I think to make the cache layer return the response directly would require moving the additional info completely out of the data fetching flow (e.g. router, cache, integrations) and into the [component](https://github.com/homarr-labs/homarr/blob/eb717da79257a72a8b14c2a2b0d320eaed23250e/packages/widgets/src/releases/component.tsx#L83-L94) itself, since all that information already exists there and isn't fetched, but rather sent on a loop. 

As for a new response structure, this is what I have stashed:
```ts
export type ReleaseData = DetailsProviderResponse & ReleaseProviderResponse;

export type ReleaseError =
  | { code: ReleasesErrorKeys }
  | { code: "unexpected", message: string };

export type ReleaseResponse =
  | { success: true; data: ReleaseData }
  | { success: false; error: ReleaseError };
```

Both of the above changes (moving the local data out of the fetching flow and changing response type structure) would require changes at the component level, so hoping to get some thoughts before moving forward.

Meierschlumpf:
Okay here are my thoughts:

I think we can remove returning of the integration from the request handler and simply do that part in the router:

```ts
return await Promise.all(
        input.repositories.map(async (repository) => {
          const innerHandler = releasesRequestHandler.handler(ctx.integration, {
            id: repository.id,
            identifier: repository.identifier,
            versionRegex: formatVersionFilterRegex(repository.versionFilter),
          });

		const releases = await innerHandler.getCachedOrUpdatedDataAsync({
            forceUpdate: false,
          })

        return {
			integration: {id: ctx.integration.id, kind: ctx.integration.kind},
			releases, 
		};
        }),
      );
```

and yeah the rest I think we just have to handle on the client in the component, but it will probably make it a lot easier as we no longer will need the `"key" in release` because it either was successful or not

castielwaverly:
I think `id` should stay in the router like you initially suggested, since the client actually needs it from the server to do matching. I'll switch the types over now and see how that goes.

castielwaverly:
Not sure how I feel about this latest change haha. I suppose it's more descriptive (+1), but now the code is more verbose and the structure is overly nested for the benefit of a few if statements (-1). I also think those if statements would get the same benefits from an early `"error" in response`. A lot of the rest of the code is already doing `response.error`, which would just be replaced with `!response.success`.

Anyway, I'm on the fence. When that happens I lean towards the simpler code with fewer changes, but perhaps you have a stronger opinion.

castielwaverly:
Did some cleanup and flattened the data structure a bit, so I'm happier with it now. Part of the nesting was caused by the router, since it returns the response from the release response handler under its own `data` key, while the release response has its own `data` key, which made for a confusing structure once you get to the client. 

Also did some testing and things are behaving as expected. Ready for review again :white_check_mark:

packages/integrations/src/interfaces/releases-providers/releases-providers-types.ts
```diff
@@ -1,5 +1,11 @@
 import type { TranslationObject } from "@homarr/translation";
 
+export interface ReleasesRepository extends Record<string, unknown> {
+  id: string;
+  identifier: string;
+  versionRegex?: string;
+}
```
Meierschlumpf: Check what this type is used for

packages/integrations/src/linuxserverio/linuxserverio-integration.ts
```diff
@@ -24,56 +24,44 @@ export class LinuxServerIOIntegration extends Integration implements ReleasesPro
...
-      };
+      return null;
     }
+    return { owner, name };
+  }
 
-    const releasesResponse = await fetchWithTrustedCertificatesAsync(this.url("/api/v1/images"));
+  public async getLatestMatchingReleaseAsync(identifier: string): Promise<ReleaseResponse> {
+    const { name } = this.parseIdentifier(identifier) ?? {};
+    if (!name) return { success: false, error: { code: "invalidIdentifier" } };
```
Meierschlumpf, line 41: What is the reason we only check for name here?

Meierschlumpf: Nvm makes total sense, it is always `linuxserver/project-name`

Meierschlumpf (approved):

Meierschlumpf:
Looks fine to me, when reviewing it I also found something that has to be improved in general as the linuxserver.io integration kind of only has the latest version that it can show I've created #4351. But this one looks good. Thanks for your work ❤️

PR #4277, opened 2025-10-12
feat(users): add libravatar / gravatar support

### LINKED ISSUE #2783: feat: gravatar

add gravatar for user icons. :)

### Additional information

_No response_

HeapReaper:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Closes #2783


apps/nextjs/src/components/user-avatar.tsx
```diff
@@ -1,19 +1,22 @@
...
   const currentSession = await auth();
 
-  const user = {
+  const user: UserProps = {
     name: currentSession?.user.name ?? null,
     image: currentSession?.user.image ?? null,
+    email: currentSession?.user.email ?? null,
   };
 
+  if (!user.image && !user.email && user.name) user.image = null;
```
Meierschlumpf: Is there a reason for this? Can you explain?

packages/ui/src/components/user-avatar.tsx
```diff
@@ -13,9 +17,15 @@ interface UserAvatarProps {
...
   if (!user?.name) return <Avatar size={size} />;
+
   if (user.image) {
     return <Avatar src={user.image} alt={user.name} size={size} />;
   }
 
+  if (user.email) {
+    const emailHash = createHash("md5").update(user.email.trim().toLowerCase()).digest("hex");
+    return <Avatar src={`https://seccdn.libravatar.org/avatar/${emailHash}?d=404`} alt={user.name} size={size} />;
+  }
```
Meierschlumpf: I'll have to check if the usage of crypto actually works

apps/nextjs/src/components/user-avatar.tsx
HeapReaper: I guess it's redundant because of `?? null` inside `const user:`.
I can remove that part yeah

HeapReaper:
> Also we should add the email to all the usages of `UserAvatar` so it doesn't only work for the current user

I did add an extra user and the Libravatar/Gravatar works fine for that user too. I could be missing smth tho
image

Meierschlumpf:
What I mean is that there are a lot of other user avatars that should also have the email always available, so you should make the email property required and add it to the missing places:

image

HeapReaper:
Ah like that, fair. Gonna add it!

manuel-rw:
Can you also add a toggle (setting) that enables Gravatar? Some people may not want it.

HeapReaper:
> What I mean is that there are a lot of other user avatars that should also have the email always available, so you should make the email property required and add it to the missing places:
> image

Atm working on it

manuel-rw:
Works now. Since the image has an override, I think we can safely merge this as is without toggle.

Meierschlumpf:
> What I mean is that there are a lot of other user avatars that should also have the email always available, so you should make the email property required and add it to the missing places:
> 
> image

This should still be done, no?

Meierschlumpf:
Also quickly added the toggle functionality:

image

packages/spotlight/src/modes/user-group/users-search-group.tsx
```diff
@@ -52,7 +52,9 @@ export const usersSearchGroup = createGroup<User>({
   Component: (user) => (
     <Group px="md" py="sm">
       <UserAvatar user={user} size="sm" />
-      <Text>{user.name}</Text>
+      <Stack gap={0}>
+        <Text>{user.name}</Text>
+      </Stack>
```
Meierschlumpf: Revert

Meierschlumpf (approved):

PR #4295, opened 2025-10-15
feat(integrations): add beszel

### LINKED ISSUE #2645: feat: beszel integration

It would be great if we could have a beszel integration. Beszel is basically dash. but with more features and support for windows systems. It would be great if the widget that comes with it could look something like this:

https://camo.githubusercontent.com/f905cf0fd5cfe9a80d3f34db062e81eebce07994a83012cf4e00fed013146666/68747470733a2f2f6265737a656c2e6465762f696d6167652f64617368626f6172642e706e67

https://github.com/henrygd/beszel

### Additional information

_No response_

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

- [x] Documentation is up to date. https://github.com/homarr-labs/documentation/pull/449

image

Closes #2645


packages/definitions/src/integration.ts
```diff
@@ -298,6 +298,13 @@ export const integrationDefs = {
...
   },
+  beszel: {
+    name: "Beszel",
+    secretKinds: [["username", "password"]],
+    iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons/svg/beszel.svg",
+    category: ["healthMonitoring"],
+    documentationUrl: createDocumentationLink("/docs/integrations/beszel"),
+  },
   // This integration only returns mock data, it is used during development (but can also be used in production by directly going to the create page)
   mock: {
```
Meierschlumpf, line 309: Add systemUsage to mock

packages/integrations/src/beszel.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+    const firstSystem = systems.at(0);
+    if (!firstSystem) throw new Error("no system found");
+
+    const systemDetails = await integration.getSystemDetailsAsync(firstSystem.id);
+
+    //const test = await pb.collection("user_settings").getFirstListItem(`user='${user.record.id}'`);
+
+    console.log(JSON.stringify(systemDetails, null, 2));
+  });
+});
```
Meierschlumpf: Replace with e2e test

packages/integrations/src/beszel/beszel-integration.ts
```diff
@@ -0,0 +1,112 @@
...
+        connectionType: system.info.ct === 1 ? "ssh" : system.info.ct === 2 ? "webSocket" : null,
+        version: system.info.v,
+      },
+      usage: {
+        cpuPercentage: system.info.cpu,
+        memoryPercentage: system.info.mp,
+        diskPercentage: system.info.dp,
+        gpuPercentage: system.info.g ?? null,
+        load: {
+          status: this.calculateLoadStatus(system.status, system.info.la, system.info.t), // TODO: handle correctly with user settings maybe?
```
Meierschlumpf: Remove comment

packages/widgets/src/system-usage/index.ts
```diff
@@ -0,0 +1,25 @@
...
+import { createWidgetDefinition } from "../definition";
+import { optionsBuilder } from "../options";
+
+export const { definition, componentLoader } = createWidgetDefinition("systemUsage", {
+  icon: IconServer,
+  supportedIntegrations: ["beszel"],
+  integrationsRequired: false,
+  createOptions() {
+    return optionsBuilder.from((factory) => ({
+      // TODO: add autocomplete with getSystems call from router
```
Meierschlumpf: Pending todo

packages/api/src/middlewares/integration.ts
```diff
@@ -13,7 +13,7 @@ import type { IntegrationKind } from "@homarr/definitions";
 
 import { publicProcedure } from "../trpc";
 
-export type IntegrationAction = "query" | "interact";
+export type IntegrationAction = "query" | "interact" | "use";
```
manuel-rw, line 16: What is the difference between "use" and "interact"?

packages/api/src/router/widgets/system-usage.ts
manuel-rw, line 1: Once again, I'm not sure whether it makes sense to add a new widget, router and handler for this, since everything is so similar with the existing ones.

packages/widgets/src/system-usage/component.tsx
manuel-rw, line 1: This is quite long. Consider splitting?

manuel-rw (changes requested):

packages/api/src/middlewares/integration.ts
Meierschlumpf: - **query:** request data to show
- **interact:** do actions like pausing the downloads / disabling dns hole
- **use:** able to select the integration for a widget and therefore be able to choose the system

packages/api/src/router/widgets/system-usage.ts
Meierschlumpf: What would be your suggestion instead?

Meierschlumpf: I mean the widget shows similar data to system-resources and health-monitoring. But at the same time it shows other data as well and it also has an option to choose a system

packages/api/src/middlewares/integration.ts
manuel-rw: Can you document this? Either by comment in code or at homarr.dev

packages/api/src/router/widgets/system-usage.ts
manuel-rw: > I mean the widget shows similar data to system-resources and health-monitoring. But at the same time it shows other data as well and it also has an option to choose a system

To decide that, we need to know the differences between them.
**What** other data does it show? Remember, we can keep this new widget and make it compatible for ``Dash.``, ``TrueNAS``, ...etc.

> What would be your suggestion instead?

Since Hardware monitoring will be one of the main use cases of Homarr in general, I think we should add your new widget to the existing hardware monitoring category and then update that category to display additional information if available (e.g. by Bazel). That is more work, but better for the user, since they can choose any of the widgets and less confusing (why would the Bezel widget not be compatible with Dash. and the other way around? Why does the new widget not have "Bezel" in it's name if it's only strictly compatible with said app?)

Meierschlumpf: I agree that it and the others should be available in all 3 widgets if possible, however the structure of all 3 is quite different. Especially the heal monitoring which distinguishes between different drives etc.

This widget only has the percentage usage and additional infos that are not on the system health monitoring widget like network usage and the agent it is using. Also it shows the name of the system directly.

Do you think it makes more sense to not merge this today and discuss how we move forward further or that we release it today and if necessary change it (for example using a migration script) in an upcoming release?

I think the people would be quite happy with the addition and we can still improve it in upcoming releases

Meierschlumpf: Oh wait we are back in normal time, true 😂 
Release was just now

ghost:
Not PR related but I think the data sources that beszel displays could be sorted into the existing widget types (or new ones):
- uptime for services (like docker stats)
- system resources
- disks
- notifications, like ntfy

ajnart:
Closing as no longer relevant

PR #4311, opened 2025-10-18
fix(docker): stats not working

### LINKED ISSUE #4305: Docker integration stopped working in v1.42.0.
Docker integration stopped working in v1.42.0. Downgrading to v1.41.0 works fine. Homarr is running under Docker.

Image

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

In some environments the cache stats property is not available which causes it to crash.
Also the cpu usage calculation results in `NaN`

From official docker documentation:

```
On Linux, the Docker CLI reports memory usage by subtracting cache usage from the total memory usage. The API does not perform such a calculation but rather provides the total memory usage and the amount from the cache so that clients can use the data as needed. The cache usage is defined as the value of total_inactive_file field in the memory.stat file on cgroup v1 hosts.

On Docker 19.03 and older, the cache usage was defined as the value of cache field. On cgroup v2 hosts, the cache usage is defined as the value of inactive_file field.
```

See https://docs.docker.com/reference/cli/docker/container/stats

Closes #4305

### DIFF
--- a/packages/request-handler/src/docker.ts
+++ b/packages/request-handler/src/docker.ts
@@ -3,7 +3,6 @@ import type { ContainerInfo, ContainerStats } from "dockerode";
 
 import { db, like, or } from "@homarr/db";
 import { icons } from "@homarr/db/schema";
-import { logger } from "@homarr/log";
 
 import type { ContainerState } from "../../docker/src";
 import { DockerSingleton } from "../../docker/src";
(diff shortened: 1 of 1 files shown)


packages/request-handler/src/docker.ts
```diff
@@ -61,7 +61,17 @@ async function getContainersWithStatsAsync() {
         })?.url ?? null,
       cpuUsage: calculateCpuUsage(stats),
       // memory usage by default includes cache, which should not be shown as it is also not shown with docker stats command
-      memoryUsage: stats.memory_stats.usage - stats.memory_stats.stats.cache,
+      // The below type is probably wrong, sometimes stats can be undefined
```
manuel-rw, line 64: 🤦‍♂️

Meierschlumpf:
Yes I tested myself during development, on my windows docker desktop and on my nas running linux. Also one of the users that had the issue confirmed it to work

PR #4441, opened 2025-11-08
fix: add orphaned status for containers with invalid stats

KartulUdus:
Setting homarr up for the first time, I got a `Cannot read properties of undefined (reading online_cpus)` error when accessing manage -> tools -> docker.

In my case it was an issue of orphaned running containers where online_cpus were null.
I achieved this feat by removing services from a docker-compose stack while the containers were still running and later restarting the compose stack. 

Having exited containers exist should not stop a user from using the Docker tool.

- Display exited containers in 
- Prevent 'Cannot read properties of undefined (reading online_cpus)' error
- Fixed exited status container error in docker-table (⨯ TypeError: Cannot read properties of null (reading 'map')

<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

### DIFF
--- a/packages/request-handler/src/docker.ts
+++ b/packages/request-handler/src/docker.ts
@@ -12,9 +12,7 @@ export const dockerContainersRequestHandler = createCachedWidgetRequestHandler({
   queryKey: "dockerContainersResult",
   widgetKind: "dockerContainers",
   async requestAsync() {
-    const containers = await getContainersWithStatsAsync();
-
-    return containers;
+    return await getContainersWithStatsAsync();
   },
   cacheDuration: dayjs.duration(20, "seconds"),
 });
@@ -28,7 +26,7 @@ async function getContainersWithStatsAsync() {
     dockerInstances.map(async ({ instance, host }) => {
       const instanceContainers = await instance.listContainers({ all: true });
       return instanceContainers
-        .filter((container) => dockerLabels.hide in container.Labels === false)
+        .filter((container) => !(dockerLabels.hide in container.Labels))
         .map((container) => ({ ...container, instance: host }));
     }),
   ).then((res) => res.flat());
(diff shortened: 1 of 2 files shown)


packages/translation/src/lang/cn.json
```diff
@@ -3549,7 +3549,8 @@
           "restarting": "正在重启",
           "exited": "已退出",
           "removing": "删除中",
-          "dead": "废弃"
+          "dead": "废弃",
+          "orphaned": "孤立"
```
manuel-rw: Please revert all translation changes except English, they are provided by our community and don't need to be added via PR

packages/request-handler/src/docker.ts
```diff
@@ -47,44 +47,117 @@ async function getContainersWithStatsAsync() {
...
+          dbIcons.find((icon) => {
+            const extractedImage = extractImage(container);
+            if (!extractedImage) return false;
+            return icon.name.toLowerCase().includes(extractedImage.toLowerCase());
+          })?.url ?? null,
+        cpuUsage: 0,
+        memoryUsage: 0,
+        image: container.Image,
+        ports: container.Ports,
+      };
```
manuel-rw: Do we really want to show "failed containers" as "orphaned"?
When does this catch usually get triggered?

packages/translation/src/lang/cn.json
KartulUdus: removed all language changes

packages/request-handler/src/docker.ts
KartulUdus: True, it's not a valid docker container status. 
Removed the "orphaned" status and tested these containers to have "exited" status. 

Simplified the approach, added safety checks to CPU and memory calculations so it won't fail if cpu_usage, numberOfCpus or memory_stats.usage is null.

exited containers don't have ports either so added a check for no ports in 
apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx too

packages/request-handler/src/docker.ts
```diff
@@ -47,7 +47,18 @@ async function getContainersWithStatsAsync() {
...
-    const stats = await instance.getContainer(container.Id).stats({ stream: false, "one-shot": true });
+    // Get stats, falling back to empty stats object if fetch fails
+    // calculateCpuUsage and calculateMemoryUsage will return 0 for invalid/missing stats
+    const stats = await instance
+      .getContainer(container.Id)
+      .stats({ stream: false, "one-shot": true })
+      .catch(() => ({
+        cpu_stats: { online_cpus: 0, cpu_usage: { total_usage: 0 }, system_cpu_usage: 0 },
+        memory_stats: { usage: 0 },
+      } as ContainerStats));
```
manuel-rw: Can you use ``satisfies`` here if possible?

tristan-pflegeabc:
i was also running into this bug after updating my images and rebooting, docker intergration stopped working, searched hours, debugged the docker api, then found out it was an orphaned container.

ajnart (approved):

manuel-rw (dismissed):

PR #4506, opened 2025-11-18
feat(docker): hide containers with label homarr.hide

### LINKED ISSUE #4428: feat: Ignoring certain containers with specific label

We should be able to hide certain containers from homarr by adding labels to a container such as `"homarr.hide"` this would make homarr ignore completely the container.
It would be useful for containers such as the socket-proxy used to make so homarr can see containers on a different machine, or to hide homarr so you can't turn off homarr from homarr.

### Additional information

I did some tests for a way to implement this and found that only the `getContainersWithStatsAsync` function in the docker request handler would need to be modified to make it possible.
Specifically the creation of the `const containers` variable.
I'm far from experienced in this so I don't know if this would be the recommended way to do it, but here's what I made that can replace the variable:
```
  const containers = await Promise.all(
    dockerInstances.map(async ({  instance, host }) => {
      const instanceContainers = await instance.listContainers({ all: true});
      return instanceContainers.map((container) => {
        if (container.Labels.hasOwnProperty("homarr.hide")) {
          return;
        }
        return ({...container,instance:host,});
      });
    }),
  ).then((res) => res.flat());
```
If this seems ok, I'd be happy to make a pull request.

generalmat82:
Filter out containers with 'homarr.hide' label when obtaining containers. From feat: #4428

<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

- [x] Documentation is up to date. https://github.com/homarr-labs/documentation/pull/457

image

image

Closes #4428

### DIFF
--- a/packages/docker/src/index.ts
+++ b/packages/docker/src/index.ts
@@ -8,3 +8,4 @@ export type { Docker };
 export const containerStates = ["created", "running", "paused", "restarting", "exited", "removing", "dead"] as const;
 
 export type ContainerState = (typeof containerStates)[number];
+export * from "./labels";
--- a/packages/docker/src/labels.ts
+++ b/packages/docker/src/labels.ts
@@ -1,0 +1,4 @@
+export const dockerLabels = {
+  // Label to hide a container from Homarrs docker integration
+  hide: "homarr.hide",
+} as const;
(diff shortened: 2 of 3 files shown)


generalmat82:
Not sure how to fix the other issues

Meierschlumpf:
Hey, thank you for your contribution. I've pulled the changes locally and improved it a bit, now I'm trying the feature out and then we can probably already merge it 😉

manuel-rw (approved):

PR #4778, opened 2025-12-31
refactor: migrate from eslint & prettier to biome

ajnart:
Completely nuke prettier and eslint , switch to biome, format codebase, use mise instead of nvm, move docs to this repo


Meierschlumpf:
1. Why switch to biome? eslint and prettier work quite well
2. If we actually move the docs into this repo, please split it up from the other changes

ajnart:
> 1. Why switch to biome? eslint and prettier work quite well
> 2. If we actually move the docs into this repo, please split it up from the other changes

1. Because biome is much faster, we don't need a specific config per package, we could also move prettier and eslint to the root of the project as an alternative. I searched and there are like 400+ combined // eslint-ignore comments in the code, these tools should help you write uniform code without interfering with the creative process, it shouldn't get in the way 
2. True, I just wanted to see how it could be done, I could revert these commits. 


I think overall it's a shift that's needed for long term better DX, aiming for less annoying first-time contribution requirements

ajnart:
@Meierschlumpf just as a sidenote, if we approve this PR ill rebase by dropping all my changes in favor of the incoming ones and then push as-is without formatting other files. I might also drop all the files that have been formatted only so that it's easier to move on with this while still having active branches. At some point we'll need to push a "format codebase" commit that will do that. (And re-enable the GH actions linked to it, if you want them)

.github/workflows/code-quality.yml
```diff
@@ -35,19 +37,57 @@ jobs:
...
+                exit 0
+              else
+                echo "$OUTPUT"
+                exit $EXIT_CODE
+              fi
+            fi
+            pnpm lint:ws
+          else
+            pnpm lint && pnpm lint:ws
+          fi
```
Meierschlumpf: This is a temporary solution until we run it once through the whole codebase, right?

docs/README.md
Meierschlumpf, line 1: Please move those changes into a separate PR

.npmrc
Meierschlumpf, line 1: I think this is still needed, same for .nvmrc

biome.json
Meierschlumpf, line 1: To me it seems like the overrides contain a lot of duplicates

mise.toml
Meierschlumpf, line 1: I'm not a fan of including this, especially not with `latest`. Also most people won't need `act`, have `git` already installed, don't need `gh`. What is docker-compose?

Meierschlumpf (commented):
What do we do about the hundereds of eslint-ignores you mentioned?

.npmrc
ajnart: I switched to mise, it's all-in-one tooling that supports about every major tool, even act for local github actions

.github/workflows/code-quality.yml
ajnart: Actually I just read about https://biomejs.dev/recipes/continuous-integration/ , biome ci only executes on the changed files, so I'll switch to that and it won't bother us in the files that are untouched by a PR

ajnart:
> What do we do about the hundereds of eslint-ignores you mentioned?

~Leave them, they're not hurting anymore~ I removed them in [0ae0b2f](https://github.com/homarr-labs/homarr/pull/4778/commits/0ae0b2f418e8a69949ac9cd0e4a9724a1b24eda2)

biome.json
ajnart: I removed them, but the rules we have are HARD, I expect we'll loosen them once this is added, or we'll probably have 500 new ``// biome-ignore`` https://biomejs.dev/analyzer/suppressions/#suppression-syntax

mise.toml
ajnart: I pinned it now, it's good to include all the tools we can use, mise is very practical for that as it only activates within the repo

Meierschlumpf:
Does biome have plugins like for package-json formatting, nextjs linting plugin etc?
I still don't really see a net positive with this pr, for me until now it's just a bit faster and a switch to a tool that at least I've never used yet

ajnart:
> Does biome have plugins like for package-json formatting, nextjs linting plugin etc?
> 
> I still don't really see a net positive with this pr, for me until now it's just a bit faster and a switch to a tool that at least I've never used yet

Yes everything is included in biome even NextJS linting if we want, as long as the rules are properly enabled. 

Also, now you don't have to worry about relying on two tools and configs + one process being spun up for every different package you're editing files in

mise.toml
```diff
@@ -0,0 +1,6 @@
+[tools]
+node = "25.1.0"
+pnpm = "10.26.2"
+biome = "2.3.10"
```
Meierschlumpf, line 4: Please use the correct versions as stated in `.nvmrc` / `package.json`

.npmrc
Meierschlumpf, line 1: keep

.nvmrc
Meierschlumpf, line 1: keep

Meierschlumpf:
IMO eslint-ignore is not bad, it shows that you now something better than the linter

I've got no experience with biome and therefore the switch is a bit risky to me, especially if we write bad code and CI doesn't catch it

Meierschlumpf:
I just tested it with `_client-providers/session.tsx` and it works the same with biome-ignore and shows me the issue. A thing we definitely have to change is that it should use semicolons.

But IMO biome-ignore gets in the way even harder as you always have to specify a reason

Meierschlumpf (changes requested):

mise.toml
ajnart: The goal is to switch to Mise as it covers more tooling. I'll update the dev docs to include steps to set it up once approved / rejected

ajnart:
> I just tested it with `_client-providers/session.tsx` and it works the same with biome-ignore and shows me the issue. A thing we definitely have to change is that it should use semicolons.
> 
> 
> 
> But IMO biome-ignore gets in the way even harder as you always have to specify a reason

I don't think semicolons are necessary, I personally don't type them in, the change was intentional. 

I know you guys are C# devs so maybe it's up to preference. Anything works as long as it auto-formats automatically

Meierschlumpf:
I'll have to decline this PR, IMO the benefit of using a different formatter and linter is negative and we have less coverage of potential issues that could be caused. Sorry

PR #4807, opened 2026-01-06
feat(docker): add container logs modal and stats columns

ajnart:
<div align="center">
  
  <h3>Homarr</h3>
</div>

## Summary
Adds Docker container log viewing from the management Docker table via a dedicated logs page, and exposes CPU/memory stats columns (hidden by default).

## Changes
- Add `docker.logs` tRPC query and server-side log decoding (Docker multiplexed format).
- Add dedicated Docker logs page with xterm.js terminal that streams logs via WebSocket subscription.
- Add "View logs" action in the management Docker table.
- Add CPU% + Memory columns (hidden by default) to the management Docker table.
- Bonus: Fix a crash on macos

### Follow-up improvements
- **Font size**: Set `fontSize: 14` on both docker and server logs terminals for consistent, readable output.
- **Sidebar active state**: Nav links now use `startsWith` matching so visiting `/manage/tools/docker/logs/...` correctly highlights the Docker item and keeps the Tools section expanded.
- **Breadcrumb translations**: Added missing `navigationStructure.manage.tools.docker.logs` translation key. Fixed `DynamicBreadcrumb` to check `dynamicMappings` before `nonInteractable`, so the container name displays instead of a raw translation key.
- **xterm cleanup**: Reversed addon disposal order (canvas before terminal) to prevent `_linkifier2 is undefined` errors when navigating away from logs pages.
- **Row action menu**: Replaced the single "View logs" action icon column with a Mantine `Menu` dropdown per row containing all actions: View logs, Start, Stop, Restart, Remove, and Add to Homarr.
- **Compact table layout**: Reduced cell padding (`4px 8px`) across all table columns for a denser, more readable layout.
- **Image tooltip**: Added a Mantine `Tooltip` on the image column so truncated image names are readable on hover.
- **Typecheck fix**: Added missing `withBorder` mapping for bookmarks widget in old-import (from dev merge).

## Screenshots

- Toggles for CPU or Memory
CleanShot 2026-01-06 at 18 59 37@2x

- Docker table action bar
CleanShot 2026-06-04 at 14 03 28
CleanShot 2026-06-04 at 14 03 33

## Checklist


apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx
```diff
@@ -76,17 +80,62 @@ const createColumns = (
     accessorKey: "ports",
     header: t("docker.field.ports.label"),
     Cell({ cell }) {
-      if (!cell.row.original.ports.length) return null;
+      if (!cell?.row?.original?.ports?.length) return null;
```
Comment (Meierschlumpf) [inline apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx]: What is the reason for these optional accessors?

apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx
```diff
@@ -76,17 +80,62 @@ const createColumns = (
...
+        </Text>
+      );
+    },
+  },
 ];
 
 export function DockerTable(initialData: RouterOutputs["docker"]["getContainers"]) {
   const t = useI18n();
   const tDocker = useScopedI18n("docker");
+  const [showStats, setShowStats] = useState(false);
```
Comment (Meierschlumpf) [inline apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx]: Remove

apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx
```diff
@@ -160,6 +209,14 @@ export function DockerTable(initialData: RouterOutputs["docker"]["getContainers"
...
   });
 
+  useEffect(() => {
+    table.setColumnVisibility((prev) => ({
+      ...prev,
+      cpuUsage: showStats,
+      memoryUsage: showStats,
+    }));
+  }, [showStats, table]);
+
```
Comment (Meierschlumpf) [inline apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx]: Remove as we have `columnVisibility: { cpuUsage: false, memoryUsage: false }`

apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx
```diff
@@ -256,3 +333,22 @@ const ContainerStateBadge = ({ state }: { state: ContainerState }) => {
...
+  return "red";
+};
+
+const cpuUsageColor = (number: number, state: string) => {
+  if (number === 0 && state !== "running") return "red";
+  if (number < 40) return "green";
+  if (number < 60) return "yellow";
+  if (number < 90) return "orange";
+  return "red";
+};
```
Comment (Meierschlumpf) [inline apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx]: Maybe we could share these functions between this and the widget

packages/modals-collection/src/docker/docker-logs-modal.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  const { data, isPending, isError } = clientApi.docker.logs.useQuery(
+    { id: innerProps.id, tail: innerProps.tail },
+    {
+      refetchOnWindowFocus: false,
+      retry: 0,
+      refetchInterval: 1000,
+      refetchIntervalInBackground: true,
+    },
+  );
```
Meierschlumpf: I think it would be nicer to have a stream of the logs. Also we could use the same system for printing the logs as we have for the logs page. Then it scrolls automatically

Meierschlumpf (dismissed):

apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx
Comment (ajnart) [reply apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx]: Because it was crashing on my PC without them

Comment (ajnart) [reply apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx]: CleanShot 2026-01-06 at 21 22 04@2x
@Meierschlumpf apparently on mac docker engine ports can be null instead of just empty, it's a quick fix that has no downsides

packages/modals-collection/src/docker/docker-logs-modal.tsx
ajnart: copilot updated the logic to have these features, follow is nice, there is a cutoff so that it only auto-scrolls if you are docked to the bottom of the logs

apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx
Comment (Meierschlumpf) [reply apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx]: Okay can you then instead please only use `cell.row.original.ports?.length` and add a comment above why

Comment (ajnart) [reply apps/nextjs/src/app/[locale]/manage/tools/docker/docker-table.tsx]: it is the case now, this code is outdated

Meierschlumpf:
Please fix CI

Also I've tried your changes out and have the following feedback:

1. It's pretty bad UX when you want to view the logs of a container that is for example on the bottom of your container list
2. The logs don't really follow, at least when watching the logs from my running Homarr instance it scrolls down a few lines, however not to the bottom. As I already said I would suggest that we reuse the same library as we used for `/manage/tools/logs` which works very nicely and can also show color in the logs

PR #4811, opened 2026-01-07
feat(ui): remember mantine-react-table preferences per table

ajnart:
## Summary
Adds per-table preference persistence for Mantine React Table (MRT) when using `useTranslatedMantineReactTable`.

## How it works (and why it looks a bit “weird”)
- **There can be a small layout shift on first load:** on a full page load, Next.js server-rendered HTML cannot include `localStorage` preferences. 

- **Client-side navigation feels instant:** when navigating within the app, the table is rendered client-side and the preference hydration runs immediately (before paint) via an **isomorphic layout effect**, so the saved preferences apply almost instantly.

- **Why `useIsomorphicLayoutEffect` exists:** it uses `useLayoutEffect` only in the browser (to apply preferences before paint), but falls back to `useEffect` on the server to avoid React warnings.

- **Why localization uses `useQuery` (non-suspense):** using suspense here could delay preference hydration; we load localization asynchronously and attach it when ready.

- **Why a reducer:** preferences are stored as one state object to reduce multiple `useState` calls and keep updates consistent.


Meierschlumpf (commented):
What do you think of this code instead?

```ts
interface PersistedMrtState {
  v: 1;
  columnVisibility: MRT_VisibilityState;
  columnOrder: MRT_ColumnOrderState;
  columnSizing: MRT_ColumnSizingState;
  columnPinning: MRT_ColumnPinningState;
  density: MRT_DensityState;
  pagination: MRT_PaginationState;
  sorting: MRT_SortingState;
  globalFilter: unknown;
  columnFilters: MRT_ColumnFiltersState;
}

const storagePrefix = "homarr:mrt";

export const useLocalStorageMantineReactTable = <TData extends MRT_RowData>(
  id: string,
  tableOptions: Omit<MRT_TableOptions<TData>, "localization" | "state">,
) => {
  const storageKey = `${storagePrefix}:${id}`;
  const defaultValue: PersistedMrtState = {
    v: 1,
    columnFilters: tableOptions.initialState?.columnFilters ?? [],
    columnOrder: tableOptions.initialState?.columnOrder ?? [],
    columnPinning: tableOptions.initialState?.columnPinning ?? {},
    columnSizing: tableOptions.initialState?.columnSizing ?? {},
    columnVisibility: tableOptions.initialState?.columnVisibility ?? {},
    density: tableOptions.initialState?.density ?? "md",
    globalFilter: tableOptions.initialState?.globalFilter ?? undefined,
    pagination: tableOptions.initialState?.pagination ?? { pageIndex: 0, pageSize: 10 },
    sorting: tableOptions.initialState?.sorting ?? [],
  };
  const [state, setState] = useLocalStorage<PersistedMrtState>({
    key: storageKey,
    defaultValue,
    serialize: superjson.stringify,
    deserialize: (str) => (str === undefined ? defaultValue : superjson.parse(str)),
  });

  return useTranslatedMantineReactTable({
    ...tableOptions,
    state,
    onColumnVisibilityChange(value) {
      tableOptions.onColumnVisibilityChange?.(value);
      setState((val) => ({
        ...val,
        columnVisibility: typeof value === "function" ? value(val.columnVisibility) : value,
      }));
    },
    onColumnOrderChange(value) {
      tableOptions.onColumnOrderChange?.(value);
      setState((val) => ({
        ...val,
        columnOrder: typeof value === "function" ? value(val.columnOrder) : value,
      }));
    },
    onColumnSizingChange(value) {
      tableOptions.onColumnSizingChange?.(value);
      setState((val) => ({
        ...val,
        columnSizing: typeof value === "function" ? value(val.columnSizing) : value,
      }));
    },
    onColumnPinningChange(value) {
      tableOptions.onColumnPinningChange?.(value);
      setState((val) => ({
        ...val,
        columnPinning: typeof value === "function" ? value(val.columnPinning) : value,
      }));
    },
    onDensityChange(value) {
      tableOptions.onDensityChange?.(value);
      setState((val) => ({
        ...val,
        density: typeof value === "function" ? value(val.density) : value,
      }));
    },
    onPaginationChange(value) {
      tableOptions.onPaginationChange?.(value);
      setState((val) => ({
        ...val,
        pagination: typeof value === "function" ? value(val.pagination) : value,
      }));
    },
    onSortingChange(value) {
      tableOptions.onSortingChange?.(value);
      setState((val) => ({
        ...val,
        sorting: typeof value === "function" ? value(val.sorting) : value,
      }));
    },
    onGlobalFilterChange(value) {
      tableOptions.onGlobalFilterChange?.(value);
      setState((val) => ({
        ...val,
        // eslint-disable-next-line @typescript-eslint/no-unsafe-call
        globalFilter: typeof value === "function" ? value(val.globalFilter) : value,
      }));
    },
    onColumnFiltersChange(value) {
      tableOptions.onColumnFiltersChange?.(value);
      setState((val) => ({
        ...val,
        columnFilters: typeof value === "function" ? value(val.columnFilters) : value,
      }));
    },
  });
};

```

IMO your implementation is quite hard to follow

ajnart:
> IMO your implementation is quite hard to follow

I don't think it requires to be changed, feel free to commit the changes if you think so

lammersbjorn (changes requested):
Just messing around ;)

lammersbjorn (approved):

Meierschlumpf:
Okay I discarded my code, it causes an infinity loop. Do you think the benefit of adding this outweights the problems it has with flickering and changing of UI between server & client? Because for example if I choose to have 5 per page and open the user management page it shows me 10 until it has completed loading where it now shows me 5

ajnart:
> Okay I discarded my code, it causes an infinity loop. Do you think the benefit of adding this outweights the problems it has with flickering and changing of UI between server & client? Because for example if I choose to have 5 per page and open the user management page it shows me 10 until it has completed loading where it now shows me 5

I am not 100% sure about this, personally I don't mind a small refresh if it gets me to see the info I care about. I think we should consider this looking at all the tables used throughout the app and decide. We could make it some kind of toggle to "save layout to localstorage" as a good compromise (false by default)

Meierschlumpf:
Yeah I like the idea of having it an option to the user, then by default nobody will complain and if they use the feature they "know" why it looks like it does

ajnart:
Waiting for https://github.com/homarr-labs/homarr/pull/4915 to be merged before adding a new preference

PR #4898, opened 2026-01-23
fix(analytics): disable auto-tracking pageviews

ajnart:
Should stop umami from sending pageviews events. So it will only send events manually tagged (with `umami.track` , that we have no instance of so far in the frontend afaik) 

Adding this because currently we have **726238** rows/events for pageviews in 23 days so far


Meierschlumpf (commented):
Does this still track the general concurrent visitors? Or does this disable all?

ajnart:
> Does this still track the general concurrent visitors? Or does this disable all?

It disables it, I guess an alternative would be to add a cron job on the analytics server to delete data after 30 days

Meierschlumpf:
Yeah I would like this more 👍🏼

Meierschlumpf:
Should I close this then?

ajnart:
Yes, I changed it to #4938

PR #5088, opened 2026-02-19
feat: add slskd integration

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

- [ ] Pull request targets `dev` branch
- [ ] No shorthand variable names are used (eg. `x`, `y`, `i` or any abbrevation)

### DIFF
--- a/packages/integrations/src/download-client/slskd/slskd-integration.ts
+++ b/packages/integrations/src/download-client/slskd/slskd-integration.ts
@@ -6,10 +6,8 @@ import { fetchWithTrustedCertificatesAsync } from "@homarr/core/infrastructure/h
 import { createLogger } from "@homarr/core/infrastructure/logs";
 import { ErrorWithMetadata } from "@homarr/core/infrastructure/logs/error";
 
-import { HandleIntegrationErrors } from "../../base/errors/decorator";
-import { integrationOFetchHttpErrorHandler } from "../../base/errors/http";
-import { Integration } from "../../base/integration";
 import type { IntegrationTestingInput } from "../../base/integration";
+import { Integration } from "../../base/integration";
 import { TestConnectionError } from "../../base/test-connection/test-connection-error";
 import type { TestingResult } from "../../base/test-connection/test-connection-service";
 import type { DownloadClientJobsAndStatus } from "../../interfaces/downloads/download-client-data";
@@ -50,7 +48,6 @@ const logger = createLogger({ module: "slskd-integration" });
  * Implementation of the Slskd application for Soulseek https://github.com/slskd/slskd
  * Slskd integration cannot pause or stop downloads, it can only list downloads.
  */
-@HandleIntegrationErrors([integrationOFetchHttpErrorHandler])
 export class SlskdIntegration extends Integration implements IDownloadClientIntegration {
   // eslint-disable-next-line no-restricted-syntax
   public pauseQueueAsync(): Promise<void> {


packages/integrations/src/download-client/slskd/slskd-integration.ts
```diff
@@ -0,0 +1,247 @@
...
+  protected async testingAsync(input: IntegrationTestingInput): Promise<TestingResult> {
+    try {
+      const response = await input.fetchAsync(this.url("/api/v0/server"), {
+        headers: {
+          "X-API-Key": this.getSecretValue("apiKey"),
+        },
+      });
+
+      if (!response.ok) {
+        return TestConnectionError.UnauthorizedResult(response.status);
```
Meierschlumpf: Does it make sense to tell the user it was unauthorized if we get back 404 response? IMO use ResponseResult instead

packages/integrations/src/download-client/slskd/slskd-integration.ts
```diff
@@ -0,0 +1,247 @@
...
+    return Promise.resolve();
+  }
+
+  // eslint-disable-next-line no-restricted-syntax,id-length
+  public deleteItemAsync(_: DownloadClientItem, _2: boolean): Promise<void> {
+    return Promise.resolve();
+  }
+
+  protected async testingAsync(input: IntegrationTestingInput): Promise<TestingResult> {
+    try {
```
Meierschlumpf: I wouldn't use a try catch and instead take advantage of the global error handler. Otherwise we hide issues like certificate, request etc.

packages/integrations/src/download-client/slskd/slskd-integration.ts
```diff
@@ -0,0 +1,247 @@
...
+          down: flatFiles.reduce((acc, file) => acc + file.averageSpeed, 0),
+          up: 0, // slskd does not upload except sharing which isn't what we want to show in downloads
+        },
+        types: ["miscellaneous"],
+      },
+      items: flatFiles.slice(0, input.limit).map((file, index) => {
+        const remainingTimeParts = file.remainingTime ? file.remainingTime.split(":") : null;
+        // The file name returned from API includes the full remote path, which is not relevant for Homarr.
+        // We remove everything before the last slash to beautify the file name.
+        const actualFileName = file.filename.split("\\").at(-1);
```
Meierschlumpf, line 115: Does it always use windows paths with backslash?

packages/integrations/src/download-client/slskd/slskd-integration.ts
```diff
@@ -0,0 +1,247 @@
...
+  ): Promise<unknown> {
+    const response = await fetchWithTrustedCertificatesAsync(this.url(path), {
+      ...options,
+      headers: {
+        "X-API-Key": this.getSecretValue("apiKey"),
+      },
+    });
+
+    if (!response.ok) {
+      throw new Error(`SLSKD API error: ${response.status} ${response.statusText}`);
```
Meierschlumpf: I would use ResponseError

Meierschlumpf:
Thank you for the great work ❤️

packages/integrations/src/download-client/slskd/slskd-integration.ts
manuel-rw: I don't know, I no longer use Windows anyway. But even if it does, it won't cause any issues as it will split into one part and at(-1) will jus take the last (only one!) part. I suggest we run it as-is and fix it if Windows users are reporting it.

Meierschlumpf: True you are using linux so I guess it just uses `\`

packages/integrations/src/download-client/slskd/slskd-integration.ts
```diff
@@ -0,0 +1,250 @@
...
+      ...options,
+      headers: {
+        "X-API-Key": this.getSecretValue("apiKey"),
+      },
+    });
+
+    if (!response.ok) {
+      throw new IntegrationResponseError(this.integration, {
+        cause: new ResponseError(response),
+      });
```
Meierschlumpf: You can actually use `ResponseError` without `IntegrationResponseError` as the error handler deals with the `ResponseError`

packages/integrations/src/download-client/slskd/slskd-integration.ts
```diff
@@ -0,0 +1,247 @@
...
+  }[];
+}
+
+const logger = createLogger({ module: "slskd-integration" });
+
+/**
+ * Implementation of the Slskd application for Soulseek https://github.com/slskd/slskd
+ * Slskd integration cannot pause or stop downloads, it can only list downloads.
+ */
+@HandleIntegrationErrors([integrationOFetchHttpErrorHandler])
```
Meierschlumpf: Is there a reason we use `ofetch` error handler? `ofetch` is a custom fetch implementation different from built-in fetch and undici

PR #5102, opened 2026-02-21
feat: add glances integration

manuel-rw:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

- [ ] Pull request targets `dev` branch
- [ ] No shorthand variable names are used (eg. `x`, `y`, `i` or any abbrevation)

### DIFF
--- a/packages/integrations/src/glances/glances-integration.ts
+++ b/packages/integrations/src/glances/glances-integration.ts
@@ -1,6 +1,7 @@
 import dayjs from "dayjs";
 import { z } from "zod/v4";
 
+import { ResponseError } from "@homarr/common/server";
 import { fetchWithTrustedCertificatesAsync } from "@homarr/core/infrastructure/http";
 
 import type { IntegrationInput, IntegrationTestingInput } from "../base/integration";
@@ -63,11 +64,21 @@ export class GlancesIntegration extends Integration implements ISystemHealthMoni
 
   private async getGlancesVersionAsync() {
     const response = await fetchWithTrustedCertificatesAsync(this.url("/api/4/version"));
+
+    if (!response.ok) {
+      throw new ResponseError(response);
+    }
+
     return await response.text();
   }
 
   private async getAllStatsAsync() {
     const response = await fetchWithTrustedCertificatesAsync(this.url("/api/4/all"));
+
+    if (!response.ok) {
+      throw new ResponseError(response);
+    }
+
     return allSchema.parseAsync(await response.json());
   }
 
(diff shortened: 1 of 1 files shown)


ajnart (dismissed):

Meierschlumpf:
What about Auth? Also maybe we can use https://glances.readthedocs.io/en/develop/api/restful.html#get-all-stats instead of fetching so many endpoints

manuel-rw:
> What about Auth? Also maybe we can use https://glances.readthedocs.io/en/develop/api/restful.html#get-all-stats instead of fetching so many endpoints

True, I didn't see that endpoint.
Auth can be implemented later IMO in a second iteration

packages/integrations/src/glances/glances-integration.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (session == null) {
+      await this.sessionStore.setAsync({
+        version: await this.getGlancesVersionAsync(),
+      });
+
+      // update the store to re-use the values below.
+      session = await this.sessionStore.getAsync();
+    }
+
+    if (session == null) throw new Error("Session was unexpectitly null");
```
Meierschlumpf, line 36: We could add a method like `lookupAsync` to `SessionStore` which combines the steps we have here and expects a callback that returns the stored value (in this case the version)

packages/integrations/src/glances/glances-integration.ts
```diff
@@ -0,0 +1,131 @@
...
+    };
+  }
+
+  private async getGlancesVersionAsync() {
+    const response = await fetchWithTrustedCertificatesAsync(this.url("/api/4/version"));
+    return await response.text();
+  }
+
+  private async getAllStatsAsync() {
+    const response = await fetchWithTrustedCertificatesAsync(this.url("/api/4/all"));
```
Meierschlumpf, line 76: Add handling for non-ok status response by throwing `ResponseError` with response as parameter (same for version)

packages/integrations/src/glances/glances-integration.ts
manuel-rw: As of now, this isn't much boilerplate. See **YAGNI**

packages/integrations/src/glances/glances-integration.ts
Meierschlumpf: What about this?

loznic89:
Hi, in the System Resources widget, is it possible to add a GPU graphics card? Fortunately, Glances detects my GPU, something I was never able to achieve with Dashdot.

PR #5125, opened 2026-02-25
feat(installations): support secrets file for all environment variables

cethien:
was trying to deploy it in docker swarm with secret managment which doesnt work without some fragile entrypoint workaround.
added optional `__FILE` suffix for environment variables.
[also seems that its more or less a convention to use optional _FILE 
](https://docs.docker.com/compose/how-tos/use-secrets/ )

sidenote: this was hacked together in a minute, tho in theory shouldnt break anything from a quick grep of the environment variables


manuel-rw:
Hi, thanks for the contribution.
Could you elaborate why the suffix doesn't work for Swarms?
Perhaps instead of fighting the consequence, we could rename it and fix the root problem?

cethien:
Hi,

could be that my wording was wrong, english is not my native, sorry about that 😄 

deploying homarr on swarm does work without issues, the "problem" is when working with `secrets` property in the compose.
usually deployed `MY_SECRET=/run/secrets/my-secret`. 

in case of homarr when doing eg `SECRET_ENCRYPTION_KEY: /run/secrets/encryption-key` it results in homarr taking the directive as the value, causing an error.

many containers support _FILE or __FILE alternatives for environment variables so figured would be cleaner that way and propably help others. since it mainly an entrypoint wrapper i dont think it should break anything either

Meierschlumpf:
We already have this for the secret encryption key:
https://github.com/homarr-labs/homarr/blob/12ba14bb3f122cb3373042098bd2c1e095b9f54e/scripts/run.sh#L23-L25

However I think it makes sense to support it for all environment variables. The only thing we have to decide in my opinion is if we

1. Use the `_FILE` suffix like it already exists for the encryption key and therefore prevent a breaking change (that is required at some point, but not has to be now). However then we have problems with environment variables that actually end with `_FILE` which would then be removed
2. Use the new `__FILE` suffix like suggested while still supporting the `_FILE` suffix for secret encryption key until the next major version where we then remove it as breaking change.

Meierschlumpf:
In the secrets in compose example from docker they use the `_FILE` suffix
image

Meierschlumpf:
We probably should use something like this for the iteration:

```bash
matching_var_names=$(env | cut -d '=' -f 1 | grep "_PATH$")
cat "$(echo ${!var})"
```

cethien:
> We already have this for the secret encryption key:
> 
> https://github.com/homarr-labs/homarr/blob/12ba14bb3f122cb3373042098bd2c1e095b9f54e/scripts/run.sh#L23-L25

okay, i checked and was in this case an error on my end, the version i had running was old and the run script didnt contain the `_FILE` suffix


> However I think it makes sense to support it for all environment variables. The only thing we have to decide in my opinion is if we
> 
>     1. Use the `_FILE` suffix like it already exists for the encryption key and therefore prevent a breaking change (that is required at some point, but not has to be now). However then we have problems with environment variables that actually end with `_FILE` which would then be removed
> 
>     2. Use the new `__FILE` suffix like suggested while still supporting the `_FILE` suffix for secret encryption key until the next major version where we then remove it as breaking change.

since this is not my project i let you guys decide. my 2 cents:
the references use `_FILE`, afaik linuxcontainers.io images also use the `_FILE` directive, so `_FILE` is propably the way to go

manuel-rw:
@Meierschlumpf what is the status of this PR? Can we merge or do you want a different solution?

Meierschlumpf:
I think we have to adjust it in the above mentioned way

Meierschlumpf:
Tested it locally with `SECRET_ENCRYPTION_KEY_FILE` and worked without any further issues

Meierschlumpf (approved):

PR #5178, opened 2026-03-06
fix(system-disks): format absolute denominator as human-readable size

### LINKED ISSUE #4996: bug: System Disks widget displays absolute value denominator in non-human readable format
### Describe the bug

When I selected the display mode to "Absolute values" the numerator value displays in a human readable format (10.8TiB), but the denominator value appears to either be total bits or bytes of the storage medium, without a unit name.

Image

### Steps to reproduce

1. Setup integration that supports getting disk usage information (In my testing this happened with dash. and unraid)
2. Add "system disks" widget to a board
3. Edit "system disks" widget and set display mode to "Absolute values"
4. Observe the "system disks" widget

### Impact

Minor. It makes reading the chart fairly difficult for the average user if the display mode is set to absolute values. There is no workaround that I am aware of other than changing the display mode, which isn't really a workaround.

### Additional information

Homarr Version: 1.52.0

### Version

### Installation method

Docker Compose

### Browser

Firefox

SergioChan:
## Summary
- fix the System Disks widget absolute mode denominator to display in human-readable units instead of raw byte counts
- keep the existing value when `available` is not a numeric byte value, to avoid changing behavior for integrations that already return formatted strings

## Testing
- attempted: `pnpm --filter @homarr/widgets lint`
- could not run in this environment because repo requires Node `>=24.14.0` while runtime is `v22.22.0`
- manual verification logic: in `absolute` mode, numeric `available` values now pass through `humanFileSize(...)` before rendering (e.g., `11811160064` -> `11 GiB`), matching the numerator's human-readable style

## Related
Fixes #4996

### DIFF
--- a/packages/definitions/src/integration.ts
+++ b/packages/definitions/src/integration.ts
@@ -148,7 +148,6 @@ export const integrationDefs = {
     secretKinds: [["apiKey"]],
     iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons@master/svg/seerr.svg",
     category: ["mediaSearch", "mediaRequest", "search"],
-    // @ts-expect-error - docs page will be created when integration is merged
     documentationUrl: createDocumentationLink("/docs/integrations/seerr"),
   },
   overseerr: {


SergioChan:
Follow-up pushed in c4d0210 to fix the \ typecheck failure in \ (TS2578 unused directive).

I couldn’t fully rerun the workspace checks locally in this environment because dependencies aren’t installed for the monorepo (no prebootstrapped node_modules), so I’m relying on CI re-run for validation of this patch.

SergioChan:
Correction to my previous comment (shell mangled formatting): follow-up commit c4d0210 removes an unused @ts-expect-error on the seerr documentationUrl line in packages/definitions/src/integration.ts, which should address the TS2578 failure. I could not fully rerun the monorepo checks locally here because dependencies are not installed in this environment; please use CI as source of truth for this patch.

packages/widgets/src/system-disks/component.tsx
```diff
@@ -15,8 +16,13 @@ const getDisplayText = (item: { used: string; available: string; percentage: num
   switch (displayMode) {
     case "percentage":
       return `${Math.round(item.percentage)}%`;
-    case "absolute":
-      return `${item.used} / ${item.available}`;
+    case "absolute": {
```
manuel-rw, line 19: Could you comment why this is necessary via code comment?

manuel-rw (commented):
I don't like this, we should fix the integrations instead...

SergioChan:
Thanks, that makes sense. I agree this should be fixed at the integration layer instead of adding display-side normalization here.\n\nI’ll stop iterating on this PR and prepare a replacement fix scoped to the affected integration(s) so the widget can keep assuming normalized units.

SergioChan:
Closing this in favor of an integration-layer fix as suggested in review. I’ll open a replacement PR with the narrower upstream-normalization change.

PR #5181, opened 2026-03-06
fix(stock-price): align sparkline color with displayed price change

### LINKED ISSUE #5149: bug: Stock price graph incorrect (when compared to price)?
### Describe the bug

I am trying to configure Stock Price widget and using 1 day / 5 minutes to see the last day changes.

However, when testing with some stocks, I noticed the following: the graph appears as green (up) but the actual price is down -2.15/-0.67% (this is AVGO stock).

I have seen a previous fix in this area (https://github.com/homarr-labs/homarr/issues/4670) related to the price/percentage (the price/% looks good to me), but the graph data may not be correct, unless I am doing something wrong.

My expectation is that the graph and price change data time range represent the same period, or at least, there should be an option for this.

Image
Image

### Steps to reproduce

1. Add Stock Price widget.
2. Configure a ticker (e.g.: AVGO) with Time Range 1 day and Time Interval 5 minutes.
Note: as the issue depends on the actual stock price, this may not be easily reproducible when the price changes.

### Impact

Confusing as the visual feedback of the graph (slope, color) does not match the price change.

### Additional information

_No response_

### Version

1.54.0

### Installation method

Docker Compose

### Browser

Brave

SergioChan:
## Summary
- use the displayed change-vs-previous-close value to drive sparkline color
- keep graph data unchanged, but prevent color from implying a gain when the shown change is negative
- remove duplicated stock symbol/title stack rendered twice in widget

## Testing
- Unable to run local lint/build in this environment: repository requires Node >=24.14.0, but runtime here is Node v22.22.0 (`pnpm -F @homarr/widgets lint src/stocks/component.tsx` fails with unsupported engine).
- Manually verified code path: sparkline color now follows `stockValuesChange` sign, matching displayed delta/percentage.

## Related
Fixes #5149


packages/widgets/src/stocks/component.tsx
```diff
@@ -80,22 +81,6 @@ export default function StockPriceWidget({ options, width, height }: WidgetCompo
...
-            <IconTrendingDown size="1.5rem" color={theme.colors.red[7]} />
-          )}
-          {data.symbol}
-        </Text>
-        {width > 280 && height > 280 && (
-          <Text size="md" lh="1">
-            {data.shortName}
-          </Text>
-        )}
-      </Stack>
```
Meierschlumpf, line 98: I think this should not be removed, no?

Meierschlumpf (changes requested):

packages/widgets/src/stocks/component.tsx
SergioChan: Good catch to double-check — this block is duplicated in `dev` (same `Stack` appears at lines 51–65 and again at 84–98 in the base file). This PR keeps the first one and removes only the duplicate at the end, so rendered UI stays the same except for the sparkline color fix from #5149.

SergioChan:
Thanks for the pushback — updated now. I restored the duplicated `Stack` block to keep this PR strictly scoped, and left only the sparkline color fix for #5149 (switching from static `trendColors` to computed `color` based on current change).

Validation note: I couldn’t run eslint locally in this environment because the repo requires Node >=24.14 while this runner is on Node 22.22.

SergioChan:
Quick follow-up: the requested scope change is now addressed (duplicate Stack removal reverted, PR now only changes sparkline color logic for #5149). Marking ready for another look when you have a moment 🙏

PR #5183, opened 2026-03-06
fix(system-disks): format absolute denominator with human-readable size

### LINKED ISSUE #4996: bug: System Disks widget displays absolute value denominator in non-human readable format

SergioChan:
## Summary
- format the absolute mode denominator using `humanFileSize` so both numerator and denominator are human-readable
- preserve a safe fallback to the raw value when the denominator cannot be parsed as a finite number

## Testing
- `pnpm eslint packages/widgets/src/system-disks/component.tsx` *(fails in this environment: Homarr requires Node >=24.14.0, runner has Node v22.22.0)*
- manual verification: absolute mode now renders `used / humanFileSize(available)` instead of raw unformatted denominator values

## Related
Closes #4996

- [x] No shorthand variable names are used
- [ ] Documentation is up to date (not required for this code-only bug fix)

### DIFF
--- a/packages/definitions/src/integration.ts
+++ b/packages/definitions/src/integration.ts
@@ -148,7 +148,6 @@ export const integrationDefs = {
     secretKinds: [["apiKey"]],
     iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons@master/svg/seerr.svg",
     category: ["mediaSearch", "mediaRequest", "search"],
-    // @ts-expect-error - docs page will be created when integration is merged
     documentationUrl: createDocumentationLink("/docs/integrations/seerr"),
   },
   overseerr: {


Meierschlumpf:
Thanks!

Issue #5242, opened 2026-03-14
bug: certificate error in immich integration


sebingel:
### Describe the bug

I cannot add an immich integration due to the following error:
```
i: Unable to connect to the integration
caused by t: Request failed type=certificate reason=untrusted code=UNABLE_TO_GET_ISSUER_CERT_LOCALLY
caused by TypeError: fetch failed 
caused by Error: unable to get local issuer certificate code=UNABLE_TO_GET_ISSUER_CERT_LOCALLY
```
Immich is behind Caddy acting as a reverse proxy handling https.
I have added the root.crt to homarrs trusted certificates through the web interface.
My pihole integration uses this cert to successfully access my pihole instance. I have profen that by removing the root cert and observing the pihole integration fail. While after adding the cert it worked again.
Even adding the certificate through the integration interface when the error occurs does not change this. It is adding a new certificate every time I click the trust button, but the integration won't work.

### Steps to reproduce

1. have immich instance
2. access it through caddy handling https
3. add caddy root.crt to homarr
4. try to add immich integration

### Impact

It makes the integration unusable.

### Additional information

<img width="915" height="133" alt="Image" src="https://github.com/user-attachments/assets/ada75e26-5118-4597-a192-7625101af790" />

<img width="757" height="187" alt="Image" src="https://github.com/user-attachments/assets/ef42994f-08a1-4d6f-9217-c4578c9168d5" />

<img width="966" height="1207" alt="Image" src="https://github.com/user-attachments/assets/b0ba7bb9-b2f9-4989-9aa3-ec4652bed380" />

### Version

1.56.0

### Installation method

Docker Compose

### Browser

Firefox

manuel-rw:
Not a bug, please read https://homarr.dev/docs/management/certificates/

sebingel:
@manuel-rw 
I have read this page more than once, and I am still not able to locate any information that would explain this behavior.
As I described I added the certificate to homarr as described on said documentation page. It works for one integration but not for another. Both integrations target a service that is reached through Caddy. So for both integrations the same root certificate should work.
I am by no means a specialist in handling certificates, so please explain which information on the linked page will help me understand what I have done wrong.

Meierschlumpf:
[...]

PR #5275, opened 2026-03-20
fix: honor trusted certificates in Immich integration requests

saschabuehrle:
## Bug
https://github.com/homarr-labs/homarr/issues/5242 — Immich integration requests fail with `UNABLE_TO_GET_ISSUER_CERT_LOCALLY` even after trusting a custom/root certificate.

## Fix
Use Homarr's trusted-certificate fetch implementation for all Immich SDK API calls:
- `getServerStatistics`
- `searchUsers`
- `getAlbumInfo`
- `getAllAlbums`
- `getMyUser`

This routes Immich requests through the same trust-store-aware HTTP path already used by other integrations.

## Testing
- Verified Immich integration now passes a trusted-cert-capable fetch option on each SDK call path used in connection tests and widget data retrieval.
- Change is scoped to Immich integration request wiring only.

Happy to address any feedback.

Greetings, saschabuehrle

### DIFF
--- a/packages/integrations/src/immich/immich-integration.ts
+++ b/packages/integrations/src/immich/immich-integration.ts
@@ -106,7 +106,8 @@ export class ImmichIntegration extends Integration {
 
   private getRequestOptions(fetchAsync = fetchWithTrustedCertificatesAsync) {
     return {
-      fetch: fetchAsync,
+      // Undici and node types are not the same
+      fetch: fetchAsync as unknown as typeof fetch,
     };
   }
 
--- a/packages/integrations/src/immich/immich-integration.ts
+++ b/packages/integrations/src/immich/immich-integration.ts
@@ -1,5 +1,6 @@
 import { getAlbumInfo, getAllAlbums, getMyUser, getServerStatistics, init, searchUsers } from "@immich/sdk";
 
+import { fetchWithTrustedCertificatesAsync } from "@homarr/core/infrastructure/http";
 import { createLogger } from "@homarr/core/infrastructure/logs";
 import { ImageProxy } from "@homarr/image-proxy";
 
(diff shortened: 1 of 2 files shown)


manuel-rw (dismissed):

packages/integrations/src/immich/immich-integration.ts
```diff
@@ -89,17 +93,23 @@ export class ImmichIntegration extends Integration {
     }[]
   > {
     this.initClient();
-    const albums = await getAllAlbums({});
+    const albums = await getAllAlbums({}, this.getRequestOptions());
     return albums.map((album) => ({ id: album.id, albumName: album.albumName, assetCount: album.assetCount }));
   }
 
   protected async testingAsync(_: IntegrationTestingInput): Promise<TestingResult> {
```
Meierschlumpf: Here the input.fetchAsync should be used instead so the error shown during connection testing is correct

Meierschlumpf: But IMO we can also merge this for tonight's release and fix it later if we don't have the time

Meierschlumpf (changes requested):

Meierschlumpf:
Feel free to merge if you fixed it / decided to fix this minor issue later

saschabuehrle:
Thanks for the follow-up and for taking a look. I verified the change remains scoped to the trusted-cert fetch wiring for Immich SDK calls only.\n\nGreetings, saschabuehrle

saschabuehrle:
Thanks — good catch.

I updated the connection test path to use the provided , while keeping the trusted-cert fetch for the regular Immich SDK calls.

Greetings, saschabuehrle

saschabuehrle:
Small correction (shell quoting ate the inline code in my previous comment):

I updated the connection test path to use the provided input.fetchAsync, while keeping the trusted-cert fetch for the regular Immich SDK calls.

Greetings, saschabuehrle

saschabuehrle:
Thanks for the review — I’ve gone through the requested changes and queued follow-up updates. I’ll push the revision in a dedicated follow-up pass.

Meierschlumpf (dismissed):

manuel-rw (dismissed):

manuel-rw:
Sorry, this won't make it to this weeks release. Couldn't figure out some of the warnings in time.

Meierschlumpf (approved):

manuel-rw (approved):

PR #5362, opened 2026-03-28
feat: add Umami Analytics integration and widget

### LINKED ISSUE #5361: feat: add Umami Analytics integration and widget
<p>Add an integration and widget for <a href="https://umami.is/">Umami Analytics</a> — a privacy-focused, open-source web analytics platform and a popular self-hosted alternative to Google           
  Analytics.</p>                                                                                                                                                                                          
  <p>Umami is widely used in the homelab/self-hosted community and exposes a clean REST API for retrieving website statistics. The integration covers API key and username/password auth; the widget is   
  designed for at-a-glance analytics directly from a Homarr board.</p>                                                                                                                                    
                                                            
  <h3>Why This Belongs in Homarr</h3>                                                                                                                                                                     
  <ul>                                                      
    <li>Umami is one of the most popular self-hosted analytics platforms (~22K GitHub stars) and a natural fit alongside the existing self-hosted integrations</li>                                       
    <li>The API is simple, stable, and well-documented — low maintenance burden</li>
    <li>No existing Homarr integration covers web analytics</li>                                                                                                                                          
    <li>Homarr itself already uses Umami for its own telemetry (v1.53.0, #4938), so the project is already familiar with the platform</li>
  </ul>                                                                                                                                                                                                   
                                                            
  <h3>Integration</h3>                                                                                                                                                                                    
  <p><strong>Authentication:</strong> Both methods Umami offers are supported:</p>
  <ul>
    <li><strong>Self-hosted:</strong> Username/password → JWT token (via <code>POST /api/auth/login</code>)</li>                                                                                          
    <li><strong>Umami Cloud:</strong> API key (via <code>x-umami-api-key</code> header)</li>
  </ul>                                                                                                                                                                                                   
  <p><strong>API Endpoints Used:</strong></p>               
                                                                                                                                                                                                          
  | Endpoint | Purpose |                                    
  |---|---|
  | `GET /api/websites` | List tracked websites (for site selector); also used as health check |
  | `GET /api/websites/:id/stats` | Summary stats: pageviews, visitors, visits, bounce rate, avg duration |                                                                                               
  | `GET /api/websites/:id/pageviews` | Time-series visitor/session data for chart views |                                                                                                                
  | `GET /api/websites/:id/active` | Real-time active visitor count (polled every 30s) |                                                                                                                  
  | `GET /api/websites/:id/metrics?type=path` | Top pages by path |                                                                                                                                       
  | `GET /api/websites/:id/metrics?type=referrer` | Top referrers |                                                                                                                                       
  | `GET /api/websites/:id/metrics?type=event` | Event names (for event selectors) |                                                                                                                      
  | `GET /api/websites/:id/events` | Per-event time-series data (for event chart views) |                                                                                                                 
                                                                                                                                                                                                          
  <h3>Widget</h3>                                                                                                                                                                                         
  <p><strong>Header row (all view modes):</strong> Domain, time frame label, unique visitor count, optional real-time active count. Stats row: pageviews, visits, bounce rate, avg visit duration.</p>    
                                                                                                                                                                                                          
  <p><strong>View modes (configurable per widget):</strong></p>
  <ul>                                                                                                                                                                                                    
    <li><strong>Chart</strong> — bar chart or sparkline of visitors over the selected time frame. Optionally overlay a single named event as a second series (grouped or overlay style for bar
  charts).</li>                                                                                                                                                                                           
    <li><strong>Events chart</strong> — multi-select one or more named events; charts each as its own colored series with no visitors baseline. Supports bar and sparkline.</li>
    <li><strong>Top pages</strong> — ranked list of top URLs by pageview count; configurable row count (1–500, default 5).</li>                                                                           
    <li><strong>Top referrers</strong> — ranked list of top referrer sources; same configuration.</li>                                                                                                    
  </ul>                                                                                                                                                                                                   
                                                                                                                                                                                                          
  <p><strong>Time frame options:</strong> Today, Last 24 hours, Last 7 days, Last 30 days, This month, Last month.</p>                                                                                    
                                                            
  <p><strong>Multi-site support:</strong> Website selector dropdown populated from the connected Umami instance. Consistent with the pattern from the Coolify widget.</p>                                 
                                                            
  <h3>Implementation</h3>                                                                                                                                                                                 
  <p>Implemented on a fork off <code>dev</code>, following the established pattern from recent integration PRs (Coolify #4789, Immich #5165):</p>
  <ul>                                                                                                                                                                                                    
    <li><code>packages/definitions/</code> — integration definition (API key + URL, or username/password + URL)</li>
    <li><code>packages/integrations/src/umami/</code> — API client, Zod schemas, types</li>                                                                                                               
    <li><code>packages/request-handler/src/umami.ts</code> — cached request handlers (5 min TTL; active visitors 30s)</li>
    <li><code>packages/widgets/src/umami/</code> — widget definition, options, and component</li>                                                                                                         
    <li><code>packages/api/src/router/widgets/umami.ts</code> — tRPC router</li>                                                                                                                          
    <li><code>packages/cron-jobs/</code> — cron job registration for each data type</li>                                                                                                                  
    <li><code>packages/translation/src/lang/en.json</code> — English translations</li>                                                                                                                    
  </ul>                                                                                                                                                                                                   
  <p>Planning to open a PR against dev shortly after this issue is filed.</p> 

### Additional information

  - Umami API docs: https://umami.is/docs/api/website-stats 
  - Umami v3 was recently released with a new UI but the API endpoints remain compatible with v2
  - Auth is the only area with real complexity — two auth modes (JWT vs API key). Both are handled and tested against self-hosted and Umami Cloud instances.                                              
  - Note: `/metrics?type=url returns` 400 on all tested Umami versions; the correct param for top pages is `type=path`

febrile42:
## Summary

Adds an integration and widget for Umami Analytics — a privacy-focused, open-source web analytics platform and a popular self-hosted alternative to Google Analytics.

Closes #5361 

## Integration

**Authentication:** Both methods Umami offers are supported:
- **API key** (`x-umami-api-key` header) — works with self-hosted and Umami Cloud
- **Username/password** → JWT token — self-hosted only

**Endpoints used:**
- `GET /api/websites` — website selector + health check
- `GET /api/websites/:id/stats` — summary stats
- `GET /api/websites/:id/pageviews` — time-series data
- `GET /api/websites/:id/active` — real-time active visitors (30s poll)
- `GET /api/websites/:id/metrics` — top pages (`type=path`), referrers, event names
- `GET /api/websites/:id/events` — per-event time-series

## Widget

**Header (all views):** domain, time frame, unique visitors, active visitors, pageviews, visits, bounce rate, avg duration.

**View modes:**
- **Chart** — bar or sparkline of visitors over time; optionally overlay a single named event (grouped or overlay style)
- **Events chart** — multi-select one or more events, each as its own colored series; no visitor baseline
- **Top pages** — ranked list of URLs by pageview count
- **Top referrers** — ranked list of referrer sources; direct traffic labeled "Direct"

**Time frames:** Today, Last 24h, Last 7d, Last 30d, This month, Last month

## Implementation notes

- Handles timestamp format divergence between self-hosted (`YYYY-MM-DD HH:MM:SS`) and Umami Cloud (ISO 8601) by normalizing both to epoch ms before joining event and pageview series
- Cache: 5-minute TTL for all data; 30s for active visitor count
- Cron jobs skip cache warming for view modes that aren't active (e.g. top pages job skips widgets in chart mode)

## Screenshots

homarr-umami-widgets

<table><tr><td>homarr-visitors-events-widget-edit</td><td>homarr-top-referrers-widget-edit</td></tr></table>

## Documentation

Docs PR: [homar-labs/documentation#520](https://github.com/homarr-labs/documentation/pull/520)


development/development.docker-compose.yml
```diff
@@ -49,6 +49,31 @@ services:
...
+    container_name: umami
+    image: ghcr.io/umami-software/umami:postgresql-latest
+    ports:
+      - "3010:3000"
+    environment:
+      DATABASE_URL: postgresql://umami:umami@umami-db:5432/umamidb
+      APP_SECRET: homarr-dev-secret-change-in-prod
+    depends_on:
+      - umami-db
+    restart: unless-stopped
```
manuel-rw: I don't think this should be part of the development compose. Otherwise we end up with a compose for every service we support

packages/definitions/src/integration.ts
```diff
@@ -28,6 +28,7 @@ interface integrationDefinition {
   category: AtLeastOneOf<IntegrationCategory>;
   documentationUrl: string | null;
   defaultUrl?: string; // optional default URL for the integration
+  urlDescription?: string; // optional hint shown below the URL field in the integration form
```
manuel-rw: This should be a separate change and PR.

packages/integrations/src/umami/umami-integration.ts
```diff
@@ -0,0 +1,356 @@
...
+export class UmamiIntegration extends Integration {
+  protected async testingAsync(input: IntegrationTestingInput): Promise<TestingResult> {
+    const authHeaders = await this.getAuthHeadersAsync();
+    const url = this.url("/websites");
+    const response = await input.fetchAsync(url, {
+      headers: authHeaders,
+    });
+
+    if (!response.ok) {
+      throw new ResponseError(response);
```
manuel-rw: Please use
https://github.com/homarr-labs/homarr/blob/4f2cc9af2a42e19624def3bdea60fcb00334c483/packages/integrations/src/homeassistant/homeassistant-integration.ts#L108

packages/widgets/src/umami/component.tsx
```diff
@@ -0,0 +1,443 @@
...
+function formatDuration(seconds: number): string {
+  if (seconds < 60) return `${seconds}s`;
+  const mins = Math.floor(seconds / 60);
+  const secs = seconds % 60;
+  return secs > 0 ? `${mins}m ${secs}s` : `${mins}m`;
+}
+
+function formatTimeFrameLabel(timeFrame: string): string {
+  switch (timeFrame) {
+    case "today":     return "Today";
```
manuel-rw: Must be translated

packages/widgets/src/umami/component.tsx
```diff
@@ -0,0 +1,443 @@
...
+    case "24h":       return "24h";
+    case "7d":        return "7d";
+    case "30d":       return "30d";
+    case "month":     return new Date().toLocaleString("en-US", { month: "short", timeZone: "UTC" });
+    case "lastMonth": {
+      const d = new Date();
+      d.setUTCMonth(d.getUTCMonth() - 1);
+      return d.toLocaleString("en-US", { month: "short", timeZone: "UTC" });
```
manuel-rw: Not everyone uses US format. Please make it international

manuel-rw (changes requested):

packages/widgets/src/umami/component.tsx
```diff
@@ -0,0 +1,481 @@
...
+  }
+}
+
+function formatDuration(seconds: number): string {
+  if (seconds < 60) return `${seconds}s`;
+  const mins = Math.floor(seconds / 60);
+  const secs = seconds % 60;
+  return secs > 0 ? `${mins}m ${secs}s` : `${mins}m`;
+}
+
```
manuel-rw: We should already have a function for this

packages/widgets/src/umami/component.tsx
```diff
@@ -0,0 +1,481 @@
...
+  const date = new Date(timestamp);
+  switch (timeFrame) {
+    case "today":
+    case "24h":
+      return `${date.getUTCHours().toString().padStart(2, "0")}:00`;
+    case "7d":
+    case "30d":
+    case "month":
+    case "lastMonth":
+      return date.toLocaleDateString(undefined, { month: "short", day: "numeric", timeZone: "UTC" });
```
manuel-rw: This will always use UTC. Please use the TZ of the client or server to calculate this.

packages/widgets/src/umami/component.tsx
manuel-rw, line 1: This file contains too many components and has too many responsibilities. Please split it into multiple files and check for existing functions as we already have some (e.g. time duration)

manuel-rw (changes requested):

manuel-rw:
@febrile42 please check the builds, ``format`` is still failing

manuel-rw:
Please format again and typecheck is still failing. I'll review after

febrile42:
@manuel-rw I believe everything is fixed. Thanks for the help!

manuel-rw:
@Meierschlumpf can you please review too?

manuel-rw (dismissed):
lgtm now, @febrile42 since this is a pretty big widget and will require effort to main, do you agree to maintain this component & integration if it breaks?
We have limited time available and your help would support us considerably.

febrile42:
I am happy to keep it updated for the foreseeable future. I use Homarr every day and plan to use the Umami widget myself -- so I have an active interest in keeping it working, beside just helping out.

ajnart:
> I am happy to keep it updated for the foreseeable future. I use Homarr every day and plan to use the Umami widget myself

Thank you very much for your PR and for being a part of the community. Feel free to ping us on Discord if you'd like to get the Contributors role and be part of the internal discussions there

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+import { createManyIntegrationMiddleware } from "../../middlewares/integration";
+import { createTRPCRouter, publicProcedure } from "../../trpc";
+
+export const umamiRouter = createTRPCRouter({
+  getWebsites: publicProcedure
+    .input(z.object({ integrationIds: z.array(z.string()) }))
+    .concat(createManyIntegrationMiddleware("query", "umami"))
+    .query(async ({ ctx }) => {
+      const integration = ctx.integrations[0];
+      if (!integration) return [];
```
Meierschlumpf: If only one integration is actually used I would suggest to also use the `createOneIntegrationMiddleware`

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+      try {
+        const instance = await createIntegrationAsync(integration);
+        return await instance.getWebsitesAsync();
+      } catch {
+        return [];
+      }
```
Meierschlumpf: Either allow the error to occure or add a log statement to warn the user that it did not work

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+  umamiTopPagesRequestHandler,
+  umamiTopReferrersRequestHandler,
+} from "@homarr/request-handler/umami";
+
+import { createManyIntegrationMiddleware } from "../../middlewares/integration";
+import { createTRPCRouter, publicProcedure } from "../../trpc";
+
+export const umamiRouter = createTRPCRouter({
+  getWebsites: publicProcedure
+    .input(z.object({ integrationIds: z.array(z.string()) }))
```
Meierschlumpf: This input is not necessary as it is already part of the middleware

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+        return await instance.getWebsitesAsync();
+      } catch {
+        return [];
+      }
+    }),
+
+  getVisitorStats: publicProcedure
+    .input(
+      z.object({
+        integrationIds: z.array(z.string()),
```
Meierschlumpf: This input property can be removed due to the usage of the middleware

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+            updatedAt: timestamp,
+          };
+        }),
+      );
+
+      return results;
+    }),
+
+  getEventNames: publicProcedure
+    .input(z.object({ integrationIds: z.array(z.string()), websiteId: z.string() }))
```
Meierschlumpf: Remove `integrationIds` as property from input due to its appearence in middleware

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+    .concat(createManyIntegrationMiddleware("query", "umami"))
+    .query(async ({ ctx, input }) => {
+      const integration = ctx.integrations[0];
+      if (!integration) return [];
```
Meierschlumpf: Use `createOneIntegrationMiddleware` instead

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+
+      try {
+        const innerHandler = umamiEventNamesRequestHandler.handler(integration, { websiteId: input.websiteId });
+        const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
+        return data;
+      } catch {
+        return [];
+      }
```
Meierschlumpf, line 77: Either allow the error to occure or add a log statement to warn the user that it did not work

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+    }),
+
+  getTopPages: publicProcedure
+    .input(
+      z.object({
+        integrationIds: z.array(z.string()),
```
Meierschlumpf: Remove `integrationIds` as property from input due to its appearence in middleware

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+        websiteId: z.string(),
+        timeFrame: z.string(),
+        limit: z.number().int().min(1).max(500),
+      }),
+    )
+    .concat(createManyIntegrationMiddleware("query", "umami"))
+    .query(async ({ ctx, input }) => {
+      const integration = ctx.integrations[0];
+      if (!integration) return [];
```
Meierschlumpf: Use `createOneIntegrationMiddleware` instead

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+        const innerHandler = umamiTopPagesRequestHandler.handler(integration, {
+          websiteId: input.websiteId,
+          timeFrame: input.timeFrame,
+          limit: input.limit,
+        });
+        const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
+        return data;
+      } catch {
+        return [];
+      }
```
Meierschlumpf, line 107: Either allow the error to occure or add a log statement to warn the user that it did not work

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+    }),
+
+  getTopReferrers: publicProcedure
+    .input(
+      z.object({
+        integrationIds: z.array(z.string()),
```
Meierschlumpf: Remove `integrationIds` as property from input due to its appearence in middleware

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+        return data;
+      } catch {
+        return [];
+      }
+    }),
+
+  getMultiEventTimeSeries: publicProcedure
+    .input(
+      z.object({
+        integrationIds: z.array(z.string()),
```
Meierschlumpf: Remove `integrationIds` as property from input due to its appearence in middleware

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+        });
+        const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
+        return data;
+      } catch {
+        return [];
+      }
+    }),
+
+  getActiveVisitors: publicProcedure
+    .input(z.object({ integrationIds: z.array(z.string()), websiteId: z.string() }))
```
Meierschlumpf: Remove `integrationIds` as property from input due to its appearence in middleware

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+        integrationIds: z.array(z.string()),
+        websiteId: z.string(),
+        timeFrame: z.string(),
+        limit: z.number().int().min(1).max(500),
+      }),
+    )
+    .concat(createManyIntegrationMiddleware("query", "umami"))
+    .query(async ({ ctx, input }) => {
+      const integration = ctx.integrations[0];
+      if (!integration) return [];
```
Meierschlumpf: Use `createOneIntegrationMiddleware` instead

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+        integrationIds: z.array(z.string()),
+        websiteId: z.string(),
+        timeFrame: z.string(),
+        eventNames: z.array(z.string()),
+      }),
+    )
+    .concat(createManyIntegrationMiddleware("query", "umami"))
+    .query(async ({ ctx, input }) => {
+      const integration = ctx.integrations[0];
+      if (!integration) return [];
```
Meierschlumpf: Use `createOneIntegrationMiddleware` instead

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+        return [];
+      }
+    }),
+
+  getActiveVisitors: publicProcedure
+    .input(z.object({ integrationIds: z.array(z.string()), websiteId: z.string() }))
+    .concat(createManyIntegrationMiddleware("query", "umami"))
+    .query(async ({ ctx, input }) => {
+      const integration = ctx.integrations[0];
+      if (!integration) return 0;
```
Meierschlumpf: Use `createOneIntegrationMiddleware` instead

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+        const innerHandler = umamiTopReferrersRequestHandler.handler(integration, {
+          websiteId: input.websiteId,
+          timeFrame: input.timeFrame,
+          limit: input.limit,
+        });
+        const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
+        return data;
+      } catch {
+        return [];
+      }
```
Meierschlumpf, line 137: Either allow the error to occure or add a log statement to warn the user that it did not work

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+        const innerHandler = umamiMultiEventRequestHandler.handler(integration, {
+          websiteId: input.websiteId,
+          timeFrame: input.timeFrame,
+          eventNames: sortedNames,
+        });
+        const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
+        return data;
+      } catch {
+        return [];
+      }
```
Meierschlumpf, line 168: Either allow the error to occure or add a log statement to warn the user that it did not work

packages/api/src/router/widgets/umami.ts
```diff
@@ -0,0 +1,174 @@
...
+      const integration = ctx.integrations[0];
+      if (!integration) return 0;
+
+      try {
+        const innerHandler = umamiActiveVisitorsRequestHandler.handler(integration, { websiteId: input.websiteId });
+        const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
+        return data;
+      } catch {
+        return 0;
+      }
```
Meierschlumpf, line 186: Either allow the error to occure or add a log statement to warn the user that it did not work

packages/cron-jobs/src/jobs/integrations/umami.ts
Meierschlumpf, line 1: I would suggest to only add cron-jobs if you also update the data automatically on the client (using subscriptions)
Therefore either
1. Add subscription endpoints for all endpoints you want to update
2. Remove the jobs

packages/definitions/src/integration.ts
```diff
@@ -370,6 +370,14 @@ export const integrationDefs = {
     category: ["speedtest"],
     documentationUrl: createDocumentationLink("/docs/integrations/speedtest-tracker"),
   },
+  umami: {
+    name: "Umami",
+    secretKinds: [["username", "password"], ["apiKey"]],
+    iconUrl: "https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons@master/svg/umami.svg",
+    category: ["analytics"],
+    defaultUrl: "https://api.umami.is/v1",
```
Meierschlumpf, line 378: Currently we have seeding logic in `seed.ts` that automatically created integrations with defaultUrl, I suggest we update the seeding logic to only do this if the secret kinds contain an empty list (so that there is an option with no auth to use that integration)

packages/definitions/src/integration.ts
```diff
@@ -370,6 +370,14 @@ export const integrationDefs = {
...
+    documentationUrl: null,
```
Meierschlumpf: You can already add it (like the others) and add a `@ts-expect-error` or similar above to ignore the type error until the docs have been released

packages/integrations/src/umami/test/umami-integration.spec.ts
Meierschlumpf, line 1: Not sure how @manuel-rw feels about this, but IMO this does not bring that much benefits and I would suggest to remove it. Maybe at some point we'll add integration tests with a real umami instance (for example the one of homarr?)

packages/integrations/src/umami/umami-integration.ts
```diff
@@ -0,0 +1,392 @@
...
+
+  public async getWebsitesAsync(): Promise<UmamiWebsite[]> {
+    const authHeaders = await this.getAuthHeadersAsync();
+    const url = this.url("/websites");
+    const response = await fetchWithTrustedCertificatesAsync(url, { headers: authHeaders });
+    if (!response.ok) throw new ResponseError(response);
+    const json = await response.json();
+    return Array.isArray(json)
+      ? umamiWebsiteSchema.array().parse(json)
+      : umamiWebsiteSchema.array().parse((json as { data: unknown }).data);
```
Meierschlumpf, line 62: Can you maybe add a comment why this differs and which format is used by which hosting type?

packages/integrations/src/umami/umami-integration.ts
```diff
@@ -0,0 +1,392 @@
...
+        startAt = endAt - 24 * 60 * 60 * 1000;
+        unit = "hour";
+    }
+
+    return { startAt, endAt, unit };
+  }
+
+  public async getVisitorStatsAsync(
+    websiteId: string,
+    timeFrame = "24h",
```
Meierschlumpf, line 128: IMO it would make more sense to send start and end date from the client (calculate them there from the timeframe)

packages/integrations/src/umami/umami-integration.ts
```diff
@@ -0,0 +1,392 @@
...
+      eventName
+        ? this.getWebsiteEventTimeSeriesAsync(website.id, startAt, endAt, unit, eventName, authHeaders)
+        : Promise.resolve(null),
+    ]);
+
+    const eventCount = eventName && eventMetrics ? (eventMetrics.find((m) => m.x === eventName)?.y ?? 0) : undefined;
+
+    // Umami instances return timestamps in different formats ("YYYY-MM-DD HH:MM:SS" vs ISO 8601).
+    // Normalise to epoch ms for a format-agnostic lookup.
+    const parseUmamiDate = (s: string) => new Date(s.includes("T") ? s : `${s.replace(" ", "T")}Z`).getTime();
```
Meierschlumpf: Extract to separate method

packages/integrations/src/umami/umami-integration.ts
```diff
@@ -0,0 +1,392 @@
...
+  }
+
+  public async getEventNamesAsync(websiteId: string): Promise<string[]> {
+    const authHeaders = await this.getAuthHeadersAsync();
+    const now = Date.now();
+    const startAt = dayjs().subtract(30, "day").valueOf();
+    const url = this.url(`/websites/${websiteId}/events`, {
+      startAt: startAt.toString(),
+      endAt: now.toString(),
+      pageSize: "100000",
```
Meierschlumpf, line 178: Not sure if this is a good idea?

Meierschlumpf: Isn't 100 more than enough?

packages/integrations/src/umami/umami-integration.ts
```diff
@@ -0,0 +1,392 @@
...
+    });
+    const response = await fetchWithTrustedCertificatesAsync(url, { headers: authHeaders });
+    if (!response.ok) return [];
```
Meierschlumpf: We should not silently fail (add log or throw error)

packages/integrations/src/umami/umami-integration.ts
Meierschlumpf: Oh wait we fetch all events to get the names of them

packages/integrations/src/umami/umami-integration.ts
```diff
@@ -0,0 +1,392 @@
...
+      }
+    }
+    return [...names].sort();
+  }
+
+  public async getActiveVisitorsAsync(websiteId: string): Promise<number> {
+    const authHeaders = await this.getAuthHeadersAsync();
+    const url = this.url(`/websites/${websiteId}/active`);
+    const response = await fetchWithTrustedCertificatesAsync(url, { headers: authHeaders });
+    if (!response.ok) return 0;
```
Meierschlumpf: We should not silently fail (log / throw)

packages/integrations/src/umami/umami-integration.ts
```diff
@@ -0,0 +1,392 @@
...
+      endAt: endAt.toString(),
+      type,
+    };
+    if (limit !== undefined) params.limit = limit.toString();
+    const url = this.url(`/websites/${websiteId}/metrics`, params);
+    const response = await fetchWithTrustedCertificatesAsync(url, {
+      headers: authHeaders,
+    });
+
+    if (!response.ok) return [];
```
Meierschlumpf: Do not silently fail (log / throw)

packages/integrations/src/umami/umami-integration.ts
```diff
@@ -0,0 +1,392 @@
...
+
+    const json = await response.json();
+    if (Array.isArray(json)) return umamiMetricItemSchema.array().parse(json);
+    return [];
```
Meierschlumpf, line 298: Is there a reason to fail silently here?

packages/integrations/src/umami/umami-integration.ts
```diff
@@ -0,0 +1,392 @@
...
+      unit,
+      timezone: "UTC",
+      eventName,
+      pageSize: "100000",
+    });
+    const response = await fetchWithTrustedCertificatesAsync(url, {
+      headers: authHeaders,
+    });
+
+    if (!response.ok) return undefined;
```
Meierschlumpf: Do not fail silently

packages/widgets/src/umami/component.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { useScopedI18n } from "@homarr/translation/client";
+
+import type { WidgetComponentProps } from "../definition";
+import { NoIntegrationDataError } from "../errors/no-data-integration";
+import { UmamiContent } from "./umami-content";
+
+export default function UmamiWidget({ options, integrationIds }: WidgetComponentProps<"umami">) {
+  if (integrationIds.length === 0) {
+    throw new NoIntegrationDataError();
+  }
```
Meierschlumpf: Can be removed as if no integration is selected it already shows an error by an upper component

packages/widgets/src/umami/umami-content.tsx
```diff
@@ -0,0 +1,217 @@
...
+  const [results] = clientApi.widget.umami.getVisitorStats.useSuspenseQuery({
+    integrationIds,
+    websiteId,
+    timeFrame,
+    eventName,
+  });
+
+  const { data: activeVisitors } = clientApi.widget.umami.getActiveVisitors.useQuery(
+    { integrationIds, websiteId },
+    { refetchInterval: 30_000 },
```
Meierschlumpf: I sugest to instead rely on a subscription with the created job. The benefit of it is, that the user using it can customize the interval of which this is updated

packages/widgets/src/umami/umami-utils.ts
```diff
@@ -0,0 +1,29 @@
...
+    case "30d":
+    case "month":
+    case "lastMonth":
+      return date.toLocaleDateString(undefined, { month: "short", day: "numeric" });
+    default:
+      return timestamp;
+  }
+}
+
+const validTimeFrames = ["today", "24h", "7d", "30d", "month", "lastMonth"] as const;
```
Meierschlumpf: We have this defined twice, I think you should be able to import this into index.ts to reuse instead of `timeFrameValues `

Meierschlumpf (changes requested):

febrile42:
Addressed all comments from this review. Summary of changes:

### API router (`packages/api/src/router/widgets/umami.ts`)
- Switched `getWebsites`, `getEventNames`, `getTopPages`, `getTopReferrers`, `getMultiEventTimeSeries`, and `getActiveVisitors` from `createManyIntegrationMiddleware` to `createOneIntegrationMiddleware` — each of these only ever uses a single integration
- Updated all client call sites accordingly (`integrationIds[]` → `integrationId: string`)
- Removed the redundant `integrationIds` property from the `getVisitorStats` outer input schema (the middleware already declares it)
- Replaced all silent `catch {}` blocks with `logger.warn(message, { error })` using `createLogger` from `@homarr/core/infrastructure/logs`

### Cron jobs
Removed all four Umami cron jobs (`umami`, `umamiTopPages`, `umamiTopReferrers`, `umamiMultiEvent`) rather than adding subscription endpoints.

**Tradeoff:** Without the cron jobs, widget data is fetched on demand. The first load per 5-minute cache window hits the Umami API live (~200–600ms spinner depending on the instance); subsequent loads within that window return from cache instantly. The active visitors counter is unaffected — it already polled client-side every 30s independently of the cron jobs.

The cron jobs made sense in theory, but as you noted, they only pre-warm a server-side cache with no mechanism to push those updates to connected clients. On-demand fetching with the existing cache TTL is the right default for now.

Adding proper subscription endpoints (following the `health-monitoring.ts` pattern) so the widget auto-updates without client polling is tracked as a future improvement.

### Integration definitions
- `documentationUrl: null` → `// @ts-expect-error` + `createDocumentationLink("/docs/integrations/umami")`, matching the pattern used for `coolify`

### Seed logic
- Updated `seedDefaultIntegrationsAsync` in `seed.ts` to only auto-create integrations where `secretKinds` includes at least one empty-array option (no-auth). Umami requires credentials for both auth methods, so it no longer gets auto-seeded despite having a `defaultUrl`.

febrile42:
@Meierschlumpf let me know if there are any questions about the fixes. I missed marking the original feedback comments as resolved, but they were as noted in the comment above.

Meierschlumpf:
Awesome thanks, I'll give this another review 👍🏼

packages/api/src/router/widgets/umami.ts
```diff
@@ -10,28 +11,28 @@ import {
...
+import { createManyIntegrationMiddleware, createOneIntegrationMiddleware } from "../../middlewares/integration";
 import { createTRPCRouter, publicProcedure } from "../../trpc";
 
+const logger = createLogger({ module: "umami-router" });
+
 export const umamiRouter = createTRPCRouter({
   getWebsites: publicProcedure
-    .input(z.object({ integrationIds: z.array(z.string()) }))
-    .concat(createManyIntegrationMiddleware("query", "umami"))
+    .input(z.object({ integrationId: z.string() }))
```
Meierschlumpf: Remove

packages/api/src/router/widgets/umami.ts
```diff
@@ -10,28 +11,28 @@ import {
...
     .query(async ({ ctx }) => {
-      const integration = ctx.integrations[0];
-      if (!integration) return [];
       try {
-        const instance = await createIntegrationAsync(integration);
+        const instance = await createIntegrationAsync(ctx.integration);
         return await instance.getWebsitesAsync();
-      } catch {
+      } catch (e) {
+        logger.warn("umami: getWebsites failed", { error: e });
```
Meierschlumpf: Use something like `logger.warn("Failed to load websites")` instead as that it is umami is already included through module property

packages/api/src/router/widgets/umami.ts
```diff
@@ -62,112 +63,107 @@ export const umamiRouter = createTRPCRouter({
     }),
 
   getEventNames: publicProcedure
-    .input(z.object({ integrationIds: z.array(z.string()), websiteId: z.string() }))
-    .concat(createManyIntegrationMiddleware("query", "umami"))
+    .input(z.object({ integrationId: z.string(), websiteId: z.string() }))
```
Meierschlumpf: Remove `integrationId` as input

packages/api/src/router/widgets/umami.ts
```diff
@@ -62,112 +63,107 @@ export const umamiRouter = createTRPCRouter({
...
-      if (!integration) return [];
-
       try {
-        const innerHandler = umamiEventNamesRequestHandler.handler(integration, { websiteId: input.websiteId });
+        const innerHandler = umamiEventNamesRequestHandler.handler(ctx.integration, { websiteId: input.websiteId });
         const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
         return data;
-      } catch {
+      } catch (e) {
+        logger.warn("umami: getEventNames failed", { error: e });
```
Meierschlumpf: Adjust log similar to above

packages/api/src/router/widgets/umami.ts
```diff
@@ -62,112 +63,107 @@ export const umamiRouter = createTRPCRouter({
...
         return [];
       }
     }),
 
   getTopPages: publicProcedure
     .input(
       z.object({
-        integrationIds: z.array(z.string()),
+        integrationId: z.string(),
```
Meierschlumpf: Remove this field

packages/api/src/router/widgets/umami.ts
```diff
@@ -62,112 +63,107 @@ export const umamiRouter = createTRPCRouter({
...
+        const innerHandler = umamiTopPagesRequestHandler.handler(ctx.integration, {
           websiteId: input.websiteId,
           timeFrame: input.timeFrame,
           limit: input.limit,
         });
         const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
         return data;
-      } catch {
+      } catch (e) {
+        logger.warn("umami: getTopPages failed", { error: e });
```
Meierschlumpf: Adjust log similar to above

packages/api/src/router/widgets/umami.ts
```diff
@@ -62,112 +63,107 @@ export const umamiRouter = createTRPCRouter({
...
         return [];
       }
     }),
 
   getTopReferrers: publicProcedure
     .input(
       z.object({
-        integrationIds: z.array(z.string()),
+        integrationId: z.string(),
```
Meierschlumpf: Remove

packages/api/src/router/widgets/umami.ts
```diff
@@ -62,112 +63,107 @@ export const umamiRouter = createTRPCRouter({
...
+        const innerHandler = umamiTopReferrersRequestHandler.handler(ctx.integration, {
           websiteId: input.websiteId,
           timeFrame: input.timeFrame,
           limit: input.limit,
         });
         const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
         return data;
-      } catch {
+      } catch (e) {
+        logger.warn("umami: getTopReferrers failed", { error: e });
```
Meierschlumpf: Adjust log similar to above

packages/api/src/router/widgets/umami.ts
```diff
@@ -62,112 +63,107 @@ export const umamiRouter = createTRPCRouter({
...
         return [];
       }
     }),
 
   getMultiEventTimeSeries: publicProcedure
     .input(
       z.object({
-        integrationIds: z.array(z.string()),
+        integrationId: z.string(),
```
Meierschlumpf: Remove

packages/api/src/router/widgets/umami.ts
```diff
@@ -62,112 +63,107 @@ export const umamiRouter = createTRPCRouter({
...
+        const innerHandler = umamiMultiEventRequestHandler.handler(ctx.integration, {
           websiteId: input.websiteId,
           timeFrame: input.timeFrame,
           eventNames: sortedNames,
         });
         const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
         return data;
-      } catch {
+      } catch (e) {
+        logger.warn("umami: getMultiEventTimeSeries failed", { error: e });
```
Meierschlumpf: Adjust log similar to above

packages/api/src/router/widgets/umami.ts
```diff
@@ -62,112 +63,107 @@ export const umamiRouter = createTRPCRouter({
...
         return [];
       }
     }),
 
   getActiveVisitors: publicProcedure
-    .input(z.object({ integrationIds: z.array(z.string()), websiteId: z.string() }))
-    .concat(createManyIntegrationMiddleware("query", "umami"))
+    .input(z.object({ integrationId: z.string(), websiteId: z.string() }))
```
Meierschlumpf: Remove `integrationId` property

packages/api/src/router/widgets/umami.ts
```diff
@@ -62,112 +63,107 @@ export const umamiRouter = createTRPCRouter({
...
       try {
-        const innerHandler = umamiActiveVisitorsRequestHandler.handler(integration, { websiteId: input.websiteId });
+        const innerHandler = umamiActiveVisitorsRequestHandler.handler(ctx.integration, {
+          websiteId: input.websiteId,
+        });
         const { data } = await innerHandler.getCachedOrUpdatedDataAsync({ forceUpdate: false });
         return data;
-      } catch {
+      } catch (e) {
+        logger.warn("umami: getActiveVisitors failed", { error: e });
```
Meierschlumpf: Adjust log similar to above

Meierschlumpf (changes requested):
Beside the below, please take a look at the previous review again, it is missing some requested changes for `umami-integration`, `component` etc.

Meierschlumpf:
@manuel-rw can you give your opinion regarding the tests? See `umami-integration.spec.ts`

> IMO this does not bring that much benefits and I would suggest to remove it. Maybe at some point we'll add integration tests with a real umami instance (for example the one of homarr?)

febrile42:
@Meierschlumpf Addressed all comments from this review round + caught a date-window off-by-one during QA. Summary of changes:

### Router (`packages/api/src/router/widgets/umami.ts`)
- Dropped redundant `integrationId` inputs from 6 procedures (middleware supplies it via `createOneIntegrationMiddleware`).
- Reworded all `logger.warn` messages to drop the `"umami:"` prefix — the `umami-router` module name auto-prefixes via `createLogger`.

### Integration (`packages/integrations/src/umami/umami-integration.ts`)
- Five previously-silent failure paths now log via the module logger and return safe defaults (`[]` / `0` / `undefined`).
- Extracted `parseUmamiDate` to a private method.
- Added comments explaining the cloud-vs-self-hosted divergences (response envelope wrapping, aggregated buckets vs raw event records, date format).
- Added comment explaining why `timeFrame` is resolved server-side (cache-key stability across users on the same dashboard).
- Fixed an off-by-one I caught during QA: `computeTimeRange` produced 30×24h rolling windows where Umami's "Last N days" preset means today + N prior days. Bounds now mirror Umami UI semantics; verified against the Umami Cloud dashboard for "Last 24 hours" and "Last 30 days".

### Widget (`packages/widgets/src/umami/`)
- Removed the dead empty-integration check in `component.tsx` (handled by the upper component).
- Replaced `validTimeFrames` in `umami-utils.ts` with the shared `timeFrameValues` exported from `index.ts`.

### Active visitors subscription
- Replaced client-side 30s polling with a single `umamiActiveVisitorsJob` cron + a `subscribeActiveVisitors` tRPC subscription that the widget consumes via `useSubscription`. The widget updates the React Query cache for `getActiveVisitors` directly so there's a single source of truth. Pattern mirrors `subscribeClusterHealthStatus` + `healthMonitoringJob`. The four other cron jobs you removed last round stay removed.

### Tests
- Removed `umami-integration.spec.ts` per your suggestion. Archived locally in case we add real-instance integration tests later.

packages/integrations/src/umami/test/umami-integration.spec.ts
febrile42: Removed in this round; archived the file locally in case we add real-Umami integration tests against a homarr instance later.

packages/integrations/src/umami/umami-integration.ts
febrile42: Kept this server-side and added a comment explaining the rationale: keeping `timeFrame` (a stable string like `"7d"`) in the cache key means all users viewing the same dashboard share one Redis cache entry per `(integration, website, timeFrame)` for the 5-min TTL. Client-supplied dates would include a per-second `now` and fragment the cache.

Side note — found and fixed an off-by-one in this function during QA: bounds now mirror Umami UI's "Last N days" preset (today + N prior days). Verified against the Umami Cloud dashboard.

packages/integrations/src/umami/umami-integration.ts
febrile42: We need to read all events in the window to extract the unique name set (the endpoint returns one record per event, no aggregated names list). Open to a smaller cap if you have a number in mind — in practice the per-record payload is small (just `eventName` / `eventType` / `createdAt`).

packages/integrations/src/umami/umami-integration.ts
febrile42: Now logs via the module logger and returns `[]` on HTTP failure. Caller behaviour unchanged on the success path.

packages/integrations/src/umami/umami-integration.ts
febrile42: Now logs via the module logger and returns `0` on HTTP failure. Caller behaviour unchanged on the success path.

packages/integrations/src/umami/umami-integration.ts
febrile42: Now logs via the module logger and returns `[]` on HTTP failure. Caller behaviour unchanged on the success path.

packages/integrations/src/umami/umami-integration.ts
febrile42: Now logs an "unexpected metrics response shape" warning when the parser fallback is hit, then returns `[]`. Caller behaviour unchanged on the success path.

packages/integrations/src/umami/umami-integration.ts
febrile42: Now logs via the module logger and returns `undefined` on HTTP failure. Caller behaviour unchanged on the success path.

packages/widgets/src/umami/umami-content.tsx
febrile42: Done — added a `subscribeActiveVisitors` tRPC subscription backed by a single `umamiActiveVisitorsJob` cron at 30s. Widget consumes via `useSubscription` and pushes updates into the React Query cache for `getActiveVisitors` so there's a single source of truth. Pattern mirrors `subscribeClusterHealthStatus` + `healthMonitoringJob`. The four other cron jobs you removed last round stay removed.

A per-widget configurable interval would need a widget-options field — happy to follow up with that as a separate change if it's worth it.

packages/api/src/router/widgets/umami.ts
```diff
@@ -17,18 +18,15 @@ import { createTRPCRouter, publicProcedure } from "../../trpc";
...
-        logger.warn("umami: getWebsites failed", { error: e });
-        return [];
-      }
-    }),
+  getWebsites: publicProcedure.concat(createOneIntegrationMiddleware("query", "umami")).query(async ({ ctx }) => {
+    try {
+      const instance = await createIntegrationAsync(ctx.integration);
+      return await instance.getWebsitesAsync();
+    } catch (e) {
+      logger.warn("Failed to load websites", { error: e });
```
Meierschlumpf: We could use `new Error` here so we see the actual error that occured

packages/integrations/src/umami/umami-integration.ts
Meierschlumpf: Makes sense, thanks

packages/widgets/src/umami/umami-content.tsx
```diff
@@ -46,10 +46,14 @@ export function UmamiContent({
     eventName,
   });
 
-  const { data: activeVisitors } = clientApi.widget.umami.getActiveVisitors.useQuery(
-    { integrationId: integrationIds[0] ?? "", websiteId },
-    { refetchInterval: 30_000 },
-  );
+  const activeVisitorsInput = { integrationId: integrationIds[0] ?? "", websiteId };
+  const { data: activeVisitors } = clientApi.widget.umami.getActiveVisitors.useQuery(activeVisitorsInput);
```
Meierschlumpf: consider using `useSuspenseQuery` instead

packages/widgets/src/umami/umami-utils.ts
```diff
@@ -18,11 +20,10 @@ export function formatXLabel(timestamp: string, timeFrame: string): string {
...
 }
 
-const validTimeFrames = ["today", "24h", "7d", "30d", "month", "lastMonth"] as const;
-type TimeFrame = (typeof validTimeFrames)[number];
+type TimeFrame = (typeof timeFrameValues)[number];
 
 export function formatTimeFrameLabel(timeFrame: string, t: ReturnType<typeof useScopedI18n<"widget.umami">>): string {
-  if ((validTimeFrames as readonly string[]).includes(timeFrame)) {
+  if ((timeFrameValues as readonly string[]).includes(timeFrame)) {
     return t(`option.timeFrame.option.${timeFrame as TimeFrame}`);
```
Meierschlumpf: These conversions are not necessary if you change type of `UmamiContentProps` to `TimeFrame` and the `timeFrame` property above as well

Meierschlumpf:
Can you verify that it still works with my changes?

Meierschlumpf (dismissed):

febrile42:
Verified locally on the merged branch — `pnpm typecheck`/`lint`/`format` are clean, suspense + subscription work as expected, and all five widget queries (`getVisitorStats`, `getMultiEventTimeSeries`, `getTopPages`, `getTopReferrers`, `getActiveVisitors`) return successfully against a self-hosted Umami v2 instance. Thanks for the cleanup pass — the `ErrorWithMetadata` payloads are a nicer log shape.

One bug I caught while QAing: `umamiActiveVisitorsSchema` had the wrong field name (`{x: number}` instead of `{visitors: number}`). The `/websites/:id/active` endpoint returns `{"visitors": N}` on both self-hosted and Cloud, so the active-visitors counter never actually worked — it was silently failing parse and falling back to 0 via the catch in the router. Pushed a 2-line fix on top of your refactor commit. Schema is in `umami-types.ts:60-62`, call site in `umami-integration.ts:209`.

Good to merge from my side after this.

Meierschlumpf (approved):

ajnart:
Great work @febrile42

PR #5412, opened 2026-04-05
fix(security): remove salt from database

Meierschlumpf:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>


FlammeGamer:
You could also remove `createSaltAsync` entirely. When you pass an integer (salt rounds) to `bcrypt.hash` instead of a salt, bcrypt will automatically generate one. See https://github.com/kelektiv/node.bcrypt.js?tab=readme-ov-file#to-hash-a-password
This would remove the need to create salts manually, as they are usually just passed directly to `hashPasswordAsync` anyway.

Meierschlumpf:
You are right, thanks for the tip!

manuel-rw (approved):
Did you test with an existing setup? Losing the salt for thousands of setups and unable to login would be *really* bad.

Meierschlumpf:
Yes I tested it with my own setup and it worked without any issue as the salt was already included previously in the password hash

PR #5510, opened 2026-04-14
fix: nextjs high memory usage

KimP0ssible-no:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

- [X] Builds without warnings or errors (`pnpm build`, autofix with `pnpm format:fix`)
- [X] Pull request targets `dev` branch
- [X] No shorthand variable names are used (eg. `x`, `y`, `i` or any abbrevation)
- [X] Documentation is up to date. Create a pull request [here](https://github.com/homarr-labs/documentation/).

Multiple issues idenfitied and tested locally:

https://github.com/vercel/next.js/issues/89091 -  The retainer chain was traced by a community member — when compression is enabled and a client disconnects mid-stream, the ServerResponse is held alive by AfterContext closures that reference the zlib Gzip instance. The socket never gets closed, sits in CLOSE-WAIT, and the native zlib buffers leak.

Setting compress: false eliminates that entire retention path. Multiple reporters on the issue confirmed it resolved their memory growth. The trade-off is that responses won't be gzip'd by Next.js. If you ever put this behind a reverse proxy (nginx), that handles compression instead.

https://github.com/vercel/next.js/issues/89091
https://github.com/vercel/next.js/issues/92287 - The httpBatchStreamLink vs httpBatchLink swap isn't tied to a single GitHub issue — it's the general recommendation from the tRPC community and multiple reporters on vercel/next.js#89091 and vercel/next.js#92287. The reasoning is straightforward: httpBatchStreamLink opens a long-lived streaming HTTP response that the client may abandon (page navigation, refresh), leaving the server-side socket in CLOSE-WAIT. httpBatchLink sends a single atomic response — the connection completes before the client can abandon it.

custom socket-cleanup-cjs to cleanup CLOSE_WAIT sockets after 60 seconds.

reproducible in original dev branch with constant refreshing of homarr homepage with multiple apps, rss feeds and plex integration:

```
ss -tnp | grep ":3000" | awk '{print $1}' | sort | uniq -c | sort -rn # check for connection states
ps aux | grep -E 'tasks\.cjs|wssServer|next-server' | grep -v grep #check RSS for homarr apps
```


manuel-rw:
Thank you for the contribution. I assume you would like to participate in the bounty mentioned in #3759 ?
Can you outline how memory consumption has changed after this change? We'll review this within the next days.

KimP0ssible-no:
> Thank you for the contribution. I assume you would like to participate in the bounty mentioned in #3759 ? Can you outline how memory consumption has changed after this change? We'll review this within the next days.

Yes thank you. memory consumption on nextjs increased rapidly as refreshes were done against homarr, corresponding to an increase in CLOSE_WAIT connections - after 100 or so refreshes RSS was at 1.5GB - after the changes RSS stabilised around the 550MB Mark - and the CLOSE_WAIT cleanup script after 60 seconds only left 2 ESTABLISHED connections and zero CLOSE_WAIT. 

the gzip compression had the biggest effect and the rest of the changes focussed around removing stale CLOSE_WAIT connections in nextjs.

Meierschlumpf:
Not sure, but I don't see any improvement regarding the memory usage:

old [NO TRAFFIC]:
- 685
- 555
- 542
- 544
- 544
- 544
- 562

new [NO TRAFFIC]:
- 636
- 558
- 560
- 568
- 575
- 579

Maybe I don't understand it correctly, but at least the base usage is the same

Meierschlumpf:
I've published it under the image tag `kimpossible-potential-fix` if you want to test it yourself (I had to fix the require in Dockerfile)

KimP0ssible-no:
> kimpossible-potential-fix

Base usage at startup should be the same, (my previous testing was locally on linux, not the docker image built) - the changes should affect memory growth over time, I would say run the ghcr.io/homarr-labs/homarr:kimpossible-potential-fix image for a few days and see how it handles things.

KimP0ssible-no:
> kimpossible-potential-fix

Running the docker image, just ran a quick test by spamming the F5 refresh on homarr, and you can see the spike in memory usage from next-server, but a quick recovery/gc:

homarr_process_memory

I'll run a 24 hour test to check stability further.

KimP0ssible-no:
@manuel-rw I ran homarr for 72 hours and accessed from 2 different locations and kept the tabs open for websocket updates, everything seems very stable and no memory growth, you can actively see the GC reducing memory occasionally:

homarr_process_memory

I'd consider this fixed, but outside testing would confirm.

apologies for my messy committing and force-pushing, hopefully if this works it can be squash merged.

manuel-rw:
Hi, I did have a look on the weekend. I think you may have misread the graphs that you have created. The total RSS is the total amount of RAM, allocated by all three processes. That amount is still mostly unchanged from the latest Homarr version. Your screenshot shows a peak of 1GiB with an average of 800MiB. Some of your changes may definitely help with some minor allocations and we might merge this, but sadly it doesn't qualify for our bounty aa far as I can tell. The bounty is for "significant reductions in memory consumption". @Meierschlumpf do you agree, that this doesn't qualify as "significant"?

If you do manage to reduce usage significantly without breaking functionality, feel free to submit another PR (ideally separate from this) and we're happy to pay out the bounty!

manuel-rw:
Also, a quick recovery is expected for most frameworks unless there is a memory leak. But as @Meierschlumpf also posted above, the base usage seems unchanged or barely measureable.

KimP0ssible-no:
> Hi, I did have a look on the weekend. I think you may have misread the graphs that you have created. The total RSS is the total amount of RAM, allocated by all three processes. That amount is still mostly unchanged from the latest Homarr version. Your screenshot shows a peak of 1GiB with an average of 800MiB. Some of your changes may definitely help with some minor allocations and we might merge this, but sadly it doesn't qualify for our bounty aa far as I can tell. The bounty is for "significant reductions in memory consumption". @Meierschlumpf do you agree, that this doesn't qualify as "significant"?
> 
> If you do manage to reduce usage significantly without breaking functionality, feel free to submit another PR (ideally separate from this) and we're happy to pay out the bounty!

I think my fixes address unfettered memory growth, not base usage, in which case I would agree with your conclusion. Either way hopefully the fixes help overall.

manuel-rw:
Yes, definitely. As promised, we will review this soon when we have time besides our job.

apps/nextjs/src/app/[locale]/_client-providers/trpc.tsx
```diff
@@ -90,7 +90,7 @@ export function TRPCReactProvider(props: PropsWithChildren) {
               url: getTrpcUrl(),
               headers: createHeadersCallbackForSource("nextjs-react (form-data)"),
             }),
-            false: httpBatchStreamLink({
+            false: httpBatchLink({
```
Comment (manuel-rw) [inline apps/nextjs/src/app/[locale]/_client-providers/trpc.tsx:93]: I'm not familiar enough with trpc.
@Meierschlumpf will the batching work here? Does it break anything?
https://tanstack.com/intent/registry/%40trpc__client/links#httpbatchlink----batch-multiple-calls-into-one-request

manuel-rw (approved):

apps/nextjs/src/app/[locale]/_client-providers/trpc.tsx
Comment (Meierschlumpf) [reply apps/nextjs/src/app/[locale]/_client-providers/trpc.tsx:93]: Not sure, but I would imagine experience wise it is better to have stream link as then the results will stream in bit by bit (instead of it having to collect all data and return at the end)

manuel-rw:
I am closing because the original issue was resolver by other means

PR #5560, opened 2026-04-23
perf(integrations): lazy-load integration classes to reduce memory at startup

alexanderxfgl-bit:
## Problem

The `@homarr/integrations` package eagerly imports all 40+ integration classes (AdGuardHome, Jellyfin, Plex, Proxmox, etc.) at module resolution time via static `import` statements in `creator.ts` and `index.ts`. This causes ~12,000 lines of integration code and their transitive dependencies (octokit, xml2js, ical.js, proxmox-api, jellyfin-sdk, etc.) to be loaded into every Node.js process that touches integrations, even if the user only configured 3-5 integrations.

This is a significant contributor to the high base memory usage reported in #3759.

## Solution

**`packages/integrations/src/base/creator.ts`**: Replaced 40+ static imports with dynamic `import()` calls. Each integration is now loaded lazily - only when `createIntegrationAsync()` is called for that specific integration kind. The `integrationCreators` map now holds async factory functions instead of class references.

**`packages/integrations/src/index.ts`**: Changed integration class re-exports to `export type` (TypeScript-only, erased at compile time). This prevents the barrel file from triggering eager loading when other packages import types or `createIntegrationAsync` from `@homarr/integrations`.

## Impact

- **tasks process**: No longer loads all integration code at startup. Only loads integrations actually configured by the user.
- **websocket process**: Same - imports `@homarr/api` which transitively imports `@homarr/integrations` types only.
- **Next.js process**: Tree-shaking handles this client-side, but server-side API routes benefit from the same lazy loading.

The transitive dependency savings are significant. For example, a user with only Pi-hole and Plex configured will no longer load `@gitbeaker/rest` (GitLab), `octokit` (GitHub), `@ctrl/deluge`, `tsdav` (CalDAV), `ical.js`, `xml2js`, `proxmox-api`, etc. in the tasks process.

## Testing

- All dynamic import paths verified against actual file locations
- Export names verified against actual class/function names
- Type-only exports preserve backward compatibility for `import * as homarrIntegrations from "@homarr/integrations"` usage in tests

Relates to #3759

### DIFF
--- a/packages/integrations/src/base/creator.ts
+++ b/packages/integrations/src/base/creator.ts
@@ -1,55 +1,12 @@
 import type { IntegrationKind } from "@homarr/definitions";
 
-import { AdGuardHomeIntegration } from "../adguard-home/adguard-home-integration";
-import { AnchorIntegration } from "../anchor/anchor-integration";
-import { CodebergIntegration } from "../codeberg/codeberg-integration";
-import { CoolifyIntegration } from "../coolify/coolify-integration";
-import { DashDotIntegration } from "../dashdot/dashdot-integration";
-import { DockerHubIntegration } from "../docker-hub/docker-hub-integration";
-import { Aria2Integration } from "../download-client/aria2/aria2-integration";
-import { DelugeIntegration } from "../download-client/deluge/deluge-integration";
-import { NzbGetIntegration } from "../download-client/nzbget/nzbget-integration";
-import { QBitTorrentIntegration } from "../download-client/qbittorrent/qbittorrent-integration";
-import { SabnzbdIntegration } from "../download-client/sabnzbd/sabnzbd-integration";
-import { SlskdIntegration } from "../download-client/slskd/slskd-integration";
-import { TransmissionIntegration } from "../download-client/transmission/transmission-integration";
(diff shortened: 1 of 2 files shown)


alexanderxfgl-bit:
Hi maintainers! Just a quick follow-up on this PR. All CI checks are passing and DeepSource gave it a Grade A. This addresses the memory optimization from #3759 by lazy-loading integration classes so they're only imported when actually needed. Happy to make any adjustments if needed. Thanks for your time!

MrRubberDucky:
image

Just putting this out here for maintainers that this seems to be an bounty hunting AI bot opening a PR to cash-in. Leaving this here as an information for maintainers so they can focus on PRs that matter, not to be malicious. (They were seen opening PRs (+91 PRs for 31 repos from April 9 to April 21) for [other bounties](https://github.com/keephq/keep/pull/6235), [another example](https://github.com/arakoodev/EdgeChains/pull/452))

manuel-rw:
I tested this PR and did not see any meaningful improvements. Therefore it does not qualify for the bounty mentioned in #3759. #5562 is superseeding these changes with more improvements. Therefore I am closing this PR.

Issue #3511, opened 2025-06-30
feat(tasks): presence detection


Meierschlumpf:
### Describe the feature you'd like to request

Add presence detection for widget cron-jobs so they only run when necessary
- Add an server setting to disable presence detection
- Add logic to add / remove something from having presence when connecting / disconnecting from a subscription
- Clear something at startup (so a crash does not let you run with unnecessary stuff)

something here means something in redis that stores the present items

### Additional information

Related #3408 
Related #2584 
Related #3278

ajnart:
Automated cleanup — this issue was closed as part of routine triage.

Reason: Zero-comment enhancement request with no maintainer engagement since 2025.

If you still think this is relevant, feel free to re-open this issue and add any new details.

PR #5562, opened 2026-04-24
perf: [3759] Memory fix

### LINKED ISSUE #3759: [Bounty Claimed] High Memory Usage
Homarr v1.0 is consistently using around 600MB of RAM on my setup. This seems excessively high for a dashboard application. I'd typically expect usage closer to 200-300MB.

Image

#### To participate in the bounty, please see https://github.com/homarr-labs/homarr/issues/3759#issuecomment-4144830306

Deses:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

- [X] Builds without warnings or errors (`pnpm build`, autofix with `pnpm format:fix`)
- [X] Pull request targets `dev` branch
- [X] No shorthand variable names are used (eg. `x`, `y`, `i` or any abbrevation)
- [X] Documentation is up to date. Create a pull request [here](https://github.com/homarr-labs/documentation/).

Continuation of some ramblings on: https://github.com/homarr-labs/homarr/issues/3759#issuecomment-4308188024

Image

# What I did!

1.- I started by Lazy-loading integration SDKs: `packages/integrations/src/base/creator.ts` was importing all 48 integration classes at startup. Even if you only use Proxmox and Pi-hole like in my case, we are  loading every single integration, which is crazy.

2.- Then I moved tasks and websockets to Nextjs to delete 2 node processes:
Both got merged into Next.js via `instrumentation.ts`, which Next.js already calls on startup , so taht one process runs everything.

3.- Another thing I changed was to stop the integrations from polling when the page is not being used. I might be overreaching but IMO if the page is closed it should not be working on refreshing the widgets since no one is looking at them. I did this by checking the websocket and setting a 1 minute timer after which the integrations stop working. This will save more cpu than memory but i guess it's nice to have.

4.- I also now force a heap limit and try to call the garbage collection more. It helps but since node seems to not unload libraries once loaded, this isn't helping as much as i'd like.
That's why if I reboot the container the memory sits at 200MB but as long as I open the page for the first time it goes to 340MB or so and sits there.

5.- To address this I first tried doing a small hardcoded test reloading homarr after 5 minutes of inactivity and, when I saw it working, I moved on to adding a setting in the configuration so users can toggle the setting and choose how much inactivity before reloading.
This change **REQUIRES** that the docker `--restart=unless-stopped` flag is set or it'll just die and never restart.

image

I added translations for all the languages. I used DeepL to translate these where they had the language available and Claude Free for the rest. I can only personally verify the correctness of English, Spanish Catalan, Portuguese and Italian. A German friend verified German and the rest... let's hope for the best.

In all honestly this commit is a bit... idk... maybe you want to take it out. It did work for me in two environments but I can see it leading to issues. That being said, this is the memory footprint after doing a reload:

Image

6.- A change you might not like or has to be tested more is the removal of `isomorphic-dompurify` from the SVG upload route. It alone needed an entire jsdom environment just to sanitize uploads which seems excessive. I replaced it with a small regexp that cleans user input.

----

Let me know what you think! I'm really excited about these changes! And I really needed these!! I have about 1Gb free on my proxmox machine. 🫠


manuel-rw:
Thanks, will review asap

Meierschlumpf:
First of thank you for taking the time and tinkering with the codebase. It seems like you might have found a good solution on how we can combine the node processes without the limited custom server for standalone nextjs, great 👍🏼 

Some thoughts I currently have about this pull request:

- It's great that it reduces the usage by about half
- It's pretty bad, that we have duplicated a lot of code (like widget definitions), but of course this can be still improved
- Some type definitions probably need a few improvements (like integration creator)
- Features like pausing jobs was kind of already an issue raised in #3511, however the current solution will cause issues for manually paused jobs. I would suggest to not make this part of the changes
- It's an interesting idea to stop the server when nobody is using it, but I think this will most likely break some setups and not sure if it actually improves memory usage by a lot
- We should move the websocket and tasks code outside the instrumentation.ts file and import it instead of the many `await import` inside there
- The change with dompurify probably needs to be reverted due to security issues with a simple regex (we might be able to optimize it however using some notes from [npmjs.com/package/isomorphic-dompurify](https://www.npmjs.com/package/isomorphic-dompurify))
- Translations should only be done for `en.json`, all other languages are translated on Crowdin by the community.

manuel-rw:
Regarding the idle restart, I don't like it for several reasons:
- It relies on the host restarting the server. In Docker, this requires the mentioned flag. This change may be breaking and might require downstream changes (e.g. proxmox, runtipi, unraid, ...)
- I doubt it frees up much resources. The most frequent complaint is of the high idle ram usage, which this PR is addressing. We can further reduce CPU usage by monitoring user presence and reducing intervals for jobs when no presence.
- It is not fixing the root problem but is trying to combat the symptoms. I'd rather focus on the root cause.

apps/nextjs/src/app/[locale]/manage/settings/_components/idle-restart-settings-form.tsx
Comment (manuel-rw) [inline apps/nextjs/src/app/[locale]/manage/settings/_components/idle-restart-settings-form.tsx:1]: As mentioned, I think we can revert this

apps/nextjs/src/app/api/user-medias/[id]/route.ts
```diff
@@ -1,11 +1,19 @@
 import { notFound } from "next/navigation";
 import { NextResponse } from "next/server";
 import type { NextRequest } from "next/server";
-import DOMPurify from "isomorphic-dompurify";
 
 import { db, eq } from "@homarr/db";
 import { medias } from "@homarr/db/schema";
 
+// Lightweight SVG sanitizer (This replaces DOMPurify!!)
+function sanitizeSvg(svg: string): string {
```
Comment (manuel-rw) [inline apps/nextjs/src/app/api/user-medias/[id]/route.ts]: 1. We should put this into a separate file; we may need to reuse it
2. Does this roughly behave the same? We disclosed the vulnerability publicly and a regression in the protection against the exploit would be fatal.

apps/nextjs/src/instrumentation.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    const exitProcess = process.exit;
+    const scheduleRestart = async () => {
+      const { enabled, gracePeriodMinutes } = await getServerSettingByKeyAsync(db, "idleRestart");
+      if (!enabled) return;
+      restartTimer = setTimeout(() => {
+        restartTimer = null;
+        logger.info(`No clients for ${gracePeriodMinutes} minutes - restarting process to free memory`);
+        exitProcess(0);
+      }, gracePeriodMinutes * 60_000);
```
manuel-rw: IMO can be reverted / removed.

apps/nextjs/src/instrumentation.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    process.on("SIGTERM", () => {
+      if (idleTimer !== null) clearTimeout(idleTimer);
+      if (restartTimer !== null) clearTimeout(restartTimer);
+      wss.clients.forEach((ws) => ws.close());
+      wss.close();
+    });
+
+    // Periodically hint V8 to collect garbage that accumulates from cron jobs and
+    // request handlers. --expose-gc must be set in NODE_OPTIONS for this to work;
```
manuel-rw: Is this flag safe for production?
Also, as a .NET developer, isn't GC supposed to run in background organized by the runtime?

apps/nextjs/src/instrumentation.ts
```diff
@@ -0,0 +1,162 @@
...
+    // if not available it is silently skipped.
```
manuel-rw: Instead of describing what, can you describe why?
Why would it not be available?  If the flag mentioned above is not set?

apps/nextjs/src/instrumentation.ts
manuel-rw, line 1: Very interesting! Thank you for this change :)

packages/integrations/src/base/creator.ts
manuel-rw, line 1: Interesting approach, @Meierschlumpf can you proofread this? I am not deep enough to fully review this code.

packages/translation/src/lang/ca.json
manuel-rw, line 1: Translations should never be done in a developer change, they are done by the community using Crowdin. You can leave them as is, the community can correct them if they are bad. For next time, you don't have to do this step :)

packages/widgets/src/bookmarks/definition.ts
manuel-rw, line 1: Does this mean now we have duplicate definitions?

SECURITY.md
manuel-rw, line 1: This change is somewhat unrelated, would be nice to have such changes separate next time :)

manuel-rw (changes requested):

manuel-rw:
I also pushed some commits to fix the linting. Can you check those out too? You can run lint and format locally to check.
``pnpm lint`` / ``pnpm format``

Deses:
I'll check out your comments and do the changes as soon as possible! probably between Thursday and Friday. I've been very busy and severely sleep deprived these past few days lol

Meierschlumpf:
No pressure - now that we have a solution, we don't have to release it right away. Please take care of yourself; I hope you'll be getting enough sleep again soon.

manuel-rw:
No worries, please sleep enough. Sleep is more important :)

apps/nextjs/src/app/api/user-medias/[id]/route.ts
Comment (Deses) [reply apps/nextjs/src/app/api/user-medias/[id]/route.ts]: It does roughly behave the same but on a surface level. 

After understanding what dompurify does, which analyzes all the DOM of the svg, it's way more thorough stripping anything bad that an svg could have. I think the small code I wrote could be bypassed with enough persistance... plus the DOMPurify team actively maintains its code and addresses new security vulnerabilities which is nice.

Replacing DOMPurify or making its footprint smaller belongs in another pr, but I'll keep what i wrote in a new file in case we want to reuse it in the future.

apps/nextjs/src/app/[locale]/manage/settings/_components/idle-restart-settings-form.tsx
Comment (Deses) [reply apps/nextjs/src/app/[locale]/manage/settings/_components/idle-restart-settings-form.tsx:1]: Yeah i'll take it out, but I think that as an optional feature for people that know what they do or for seriously memory constrained users could be nice.

apps/nextjs/src/instrumentation.ts
Deses: I've reverted the gc changes too.
`--expose-gc` is used to enable the GC hint calls which it just nudged the gc to work but it can also ignore it. You know how temperamental gc is
`--max-old-space-size=400` can help on constrained hosts by capping heap growth, but 400 MB could be a tad aggressive for a nextjs app and could cause OOM crashes. In my testing I haven't had problems and the app did grow its legs to 800mbs just to go back after some time but without extensive testing I can't vouch for it to work for everyone. That's another option we could expose to an user that knows what's doing,

packages/translation/src/lang/ca.json
Deses: I understand. I personally hate crowdin because it takes a crazy amount of time to change each line when I can just use my IDE and change them directly... iirc i tried to help`translate speedtest-tracker once and got frustrated with the tool and just gave up.

It would be great to have the ability to submit translation changes through a PR and have others review them in Crowdin.
Does Crowdin updates what's been changed from github?

packages/widgets/src/bookmarks/definition.ts
Deses: Yeah, but it's intentional. The definition.ts files are server-side stubs so the server can compute default widget option values without dragging in React or UI dependencies. The full definitions in index.tsx are still used client-side as before. It's more files but it keeps the server bundle clean. 

Before my changes the server was importing full widget modules (React components, Mantine, Tabler icons, client API etc etc) just to call createOptions(). The definition.ts stubs cut that down to just the schema. It's more of a bundle/startup improvement than a runtime heap reduction though. The big memory improvement comes from consolidating the websocket and tasks processes into Next.js instead of running them separately.

SECURITY.md
Deses: guilty as charged   🥲

Deses:
After addressing some of the cahnges you wanted me to make, this is how much memory Homarr is using now

image
And still very little usage after just rebooting and not opening the page.
image

I think it's worth it to explore an auto-reset again in the future as an advanced user option.

I'll test a bit more before pushing :)

manuel-rw:
Can you @ me as soon as we can discuss the bounty? You can also DM me on Discord so we can organise the payment (Manicr..)

Deses:
> Can you @ me as soon as we can discuss the bounty? You can also DM me on Discord so we can organise the payment (Manicr..)

Oh btw, done! You have a friend request. :)

ajnart:
Closing this as no longer relevant

PR #5595, opened 2026-04-30
feat(app): support path-only hrefs for multi-hostname access

mairas:
### Motivation

We run Homarr as the home dashboard of a small Linux distribution ([HaLOS](https://docs.halos.fi)) where Traefik fronts a set of containerized apps and exposes each one at a distinct path under the same origin (`/cockpit/`, `/grafana/`, `/signalk-server/`, etc.), with Authelia providing SSO across them.

A given device can be reached on multiple hostnames in different contexts: locally over mDNS as `halos.local`, and remotely over a VPN as something like `halos.example.com`. Our Traefik and SSO setup copes with both — TLS SANs, Authelia per-host cookies, and OIDC redirect URIs all adapt per request — but Homarr does not. App hrefs must be absolute URLs today, so a card created while reaching the dashboard at `halos.local` permanently links to `https://halos.local/cockpit/`. Clicking it from the VPN jumps the browser to a hostname that doesn't resolve in that context.

Path-only hrefs (`/cockpit/`) fix this cleanly: the browser resolves them against the current origin natively, so the same card works under every hostname the dashboard answers on. The change is purely additive — existing absolute hrefs are unaffected at every call site.

#### 1. Schema relaxation — `packages/validation/src/app.ts`

`appHrefSchema` now accepts either:

- An absolute URL (existing behavior, byte-identical), OR
- A path-only form: `/[^/]...` — must start with a single forward slash followed by a non-slash character.

Path-only branch rejects:

- `javascript:` and other dangerous schemes (existing protection preserved by branch ordering).
- Protocol-relative hrefs (`//host/...`).
- Single-slash root (`/`) — too ambiguous to be a useful link target.
- Consecutive slashes mid-path (`/foo//bar`) — defends against accidental URL-parsing surprises in downstream consumers.
- Backslashes anywhere in the path.
- Whitespace and JavaScript `\s` characters anywhere in the path.
- C0/C1 control characters.
- Unicode zero-width, bidi, and formatting characters anywhere in the path (display-spoofing protection — paths like `/admin‮‭` that visually resemble `/admin` but route differently).

Empty-string-to-null transform behavior is preserved (existing).

#### 2. Server-side resolver — `packages/common/src/url.ts`

New `resolveServerUrl(app)` helper centralizes the existing `pingUrl ?? href` pattern:

```ts
function resolveServerUrl(app: { href: string | null; pingUrl: string | null }): string | null {
  if (app.pingUrl) return app.pingUrl;          // step 1 — explicit override
  if (isAbsoluteUrl(app.href)) return app.href; // step 2 — existing behavior
  return null;                                   // step 3 — path-only without pingUrl
}
```

**Design choice: path-only hrefs intentionally resolve to `null` server-side.** Path-only hrefs are a browser-resolved form. The server has no canonical hostname for an app whose href is path-only, and must not synthesize one from request headers — that would be a header-spoofing / SSRF vector (a malicious client setting `x-forwarded-host: internal.target` could redirect server-side fetches arbitrarily).

An earlier draft of this resolver expanded path-only hrefs against `extractBaseUrlFromHeaders(headers)`. We dropped that branch on review — it adds an attack surface for a feature path-only-href apps don't actually need. Apps that require server-side coverage under multi-hostname deployments should set an explicit `pingUrl`.

#### 3. Ping widget — `packages/api/src/router/widgets/app.ts` + `packages/widgets/src/app/{component,ping/ping-indicator}.tsx`

The `pingUrl ?? href` pattern in the router is routed through `resolveServerUrl`. When it returns `null` (path-only without explicit `pingUrl`), the router responds with a tRPC `CONFLICT` error instead of attempting a fetch.

**Client-side: switch from `useSuspenseQuery` to `useQuery`.** The previous indicator used `useSuspenseQuery` wrapped in a `<Suspense>` boundary in `component.tsx`. With Suspense, a tRPC error propagates to the nearest error boundary — which would treat the new "no ping URL available" CONFLICT response as a render-breaking error rather than the expected absence it actually is. The refactor:

- `ping-indicator.tsx`: `useSuspenseQuery` → `useQuery` with `retry: false`. Reads `query.data` / `query.error` directly. While loading or pending data, renders the blue `IconLoader` dot inline. On CONFLICT (path-only without pingUrl), renders a permanent blue `IconLoader` dot with the server's error message as tooltip — an explicit indeterminate state. Other tRPC errors (FORBIDDEN, NOT_FOUND, etc.) are re-thrown so the widget error boundary handles them as before.
- `component.tsx`: drops the `<Suspense fallback={<PingDot icon={IconLoader}…/>}>` wrapper since the loader is now handled inline by the indicator.

Net behavior: existing apps with absolute href / explicit pingUrl render byte-identically to today (loader → dot transition is now in the indicator instead of via Suspense, but visually the same). Path-only-without-pingUrl apps display the indeterminate blue loader dot indefinitely rather than tripping the widget error boundary — visually communicating "ping not configured" without breaking the card.

#### 4. Integration middleware — `packages/api/src/middlewares/integration.ts`

`externalUrl: rest.app?.href ?? null` is replaced with `externalUrl: resolveServerUrl(rest.app)`. This collapses path-only hrefs to `null`, letting the existing `externalUrl ?? integration.url` fallback at `packages/integrations/src/base/integration.ts:85` take over.

Without this, integrations bound to path-only-href apps crash at runtime via `new URL("/cockpit/" + path)` → `TypeError [ERR_INVALID_URL]`.

#### 5. Bookmarks widget — `packages/widgets/src/bookmarks/sub-label.ts`

Extracted `getHrefSubLabel` helper. Branch-free given schema-validated inputs:

- Absolute href → render hostname (`docs.example.com`).
- Path-only href → render trailing-slash-trimmed path (`/cockpit/` → `/cockpit`).
- Null → empty.

### Backward compatibility

Every change is additive:

- Schema: existing absolute-URL inputs continue to validate identically; the path-only branch is an OR alternative.
- Resolver: for apps with explicit `pingUrl` or absolute `href`, the resolver returns the same string today's `pingUrl ?? href` pattern would. The new `null` case is reachable only for path-only hrefs without `pingUrl` — a state that is impossible under the old schema.
- Bookmarks widget: previously rendered `new URL(app.href).hostname`; the new helper produces the same hostname for all currently-valid hrefs.
- Integration middleware: previously could only see absolute hrefs; the resolver returns those identically.

There is no migration; existing data continues to validate and render unchanged.

### Tests

- `packages/validation/src/test/app.spec.ts` — 33 schema cases: 12 acceptance, 21 rejection (backslash, whitespace, control chars, zero-width / bidi / formatting, consecutive slashes, javascript:, protocol-relative, bare strings, single-slash root).
- `packages/common/src/test/url.spec.ts` — 11 resolver cases.
- `packages/widgets/src/bookmarks/test/sub-label.spec.ts` — 10 sub-label render cases.
- `packages/api/src/router/test/widgets/app.spec.ts` — 3 router-level ping cases (path-only-without-pingUrl CONFLICT, path-only-with-pingUrl, absolute-href passthrough).

### PR template checklist

- [x] Documentation is up to date.

> *Documentation is included in this PR. Since `homarr-labs/documentation` was archived and the docs moved into the monorepo under `apps/docs`, the app URL field documentation is updated here in `apps/docs/docs/management/apps/index.mdx` rather than in a separate docs PR.*

## Summary by CodeRabbit

* **New Features**
  * App links now support validated path-only hrefs (in addition to full URLs).
  * Bookmarks now display a smarter sub-label for both absolute and path-based links.
  * Ping status widgets now use improved loading and dedicated conflict error handling.
* **Bug Fixes**
  * Ping URL checks now require a full ping target for path-only apps, preventing incorrect ping behavior.
* **Documentation**
  * Updated “Ping URL” documentation to clarify full ping URL requirements and improved section guidance.
* **Documentation**
  * Expanded app href validation messaging for invalid formats and consecutive slashes.


Meierschlumpf:
Hey, thanks for your contribution we'll review and discuss this change in the team. If I understand it correctly you were able to configure Homarr to use subpaths `https://halos.local/homar/` how did you do that?

mairas:
> Hey, thanks for your contribution we'll review and discuss this change in the team. If I understand it correctly you were able to configure Homarr to use subpaths `https://halos.local/homar/` how did you do that?

Thanks!

Homarr is the main landing page on the device, so it's always at `https://hostname.local/`, not on a sub-path. Was that what you were asking?

The container apps are exposed to subpaths, e.g. `https://hostname.local/cockpit/`, but Traefik just 302 redirects to the exposed ports such as `https://hostname.local:9090/` because many of the apps don't like to live at sub-paths.

Meierschlumpf:
Oh makes sense (yes that is what I meant)
We actually also do not support subpaths because the web framework we're using only supports them when specified before build

Meierschlumpf:
I'll take a look at this pr later tonight, sorry for the delay

packages/api/src/middlewares/integration.ts
```diff
@@ -66,7 +67,7 @@ export const createOneIntegrationMiddleware = <TKind extends IntegrationKind>(
       ctx: {
         integration: {
           ...rest,
-          externalUrl: rest.app?.href ?? null,
+          externalUrl: rest.app ? resolveServerUrl(rest.app) : null,
```
Meierschlumpf: I think here we can actually keep the `app?.href ?? null` because it is used on the client. Otherwise it will use the pingUrl. Same below

packages/validation/src/test/app.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  test.each([
+    ["https://example.com/path", "https://example.com/path"],
+    ["http://example.com/path", "http://example.com/path"],
+    ["https://example.com", "https://example.com"],
+    ["/cockpit/", "/cockpit/"],
+    ["/signalk-server/@signalk/freeboard-sk/", "/signalk-server/@signalk/freeboard-sk/"],
+    ["/x", "/x"],
+  ])("accepts %s", (input, expected) => {
+    expect(appHrefSchema.parse(input)).toBe(expected);
+  });
```
Meierschlumpf, line 15: How about only specifying one input? Then it would just check it to be the same

packages/validation/src/app.ts
```diff
@@ -1,10 +1,51 @@
...
+const pathOnlyHrefSchema = z
+  .string()
+  .trim()
+  // Leading "/" followed by at least one non-"/" / non-disallowed char,
+  // then any mix of allowed chars (including "/" for internal segments).
+  .regex(new RegExp(`^/[^/${PATH_DISALLOWED}][^${PATH_DISALLOWED}]*$`))
+  // Reject consecutive slashes anywhere ("/foo//bar"). Cosmetic on its own
+  // but keeps the rendered sub-label canonical and matches the leading-"/"
+  // exclusion of "//host/...".
+  .refine((s) => !s.includes("//"), { message: "consecutive slashes are not allowed" });
```
Meierschlumpf: We could add a translated error message here
You can use 

```ts
createCustomErrorParams({
          key: "passwordsDoNotMatch", // use a custom key here (translation in common.zod.errors.custom.{key})
          params: {},
})
```

packages/widgets/src/app/ping/ping-indicator.tsx
Meierschlumpf, line 1: I would suggest to use an error-boundary instead (we could consider moving the suspense and error-boundary into a component within this file)

packages/widgets/src/app/ping/ping-indicator.tsx
```diff
@@ -32,6 +35,21 @@ export const PingIndicator = ({ appId }: PingIndicatorProps) => {
     },
   );
 
+  // Apps without a server-pingable URL (e.g. path-only href without an explicit
+  // pingUrl) yield a CONFLICT. Render an indeterminate dot for that case so the
+  // card stays usable. Other tRPC errors (FORBIDDEN, NOT_FOUND) still bubble to
+  // the widget error boundary as before.
+  if (query.error) {
+    if (query.error.data?.code === "CONFLICT") {
+      return <PingDot icon={IconLoader} color="blue" tooltip={query.error.message} />;
```
Meierschlumpf: We could use orange for this case

Meierschlumpf (changes requested):

mairas:
> I'll take a look at this pr later tonight, sorry for the delay

No worries, real life intervenes etc. I'll go through your comments. They don't seem to bad. :-)

packages/widgets/src/app/ping/ping-indicator.tsx
mairas: Orange equals conflict. Yes Mr. President.

packages/validation/src/app.ts
mairas: Sicher.

packages/validation/src/test/app.spec.ts
mairas: Doh.

packages/validation/src/app.ts
mairas: I let Claude also create a part of the translations. Couldn't review them myself, for obvious reasons. ;-)

packages/api/src/middlewares/integration.ts
mairas: Claude Code wanted to push back because externalUrl is also used server-side. I'm not sure I can completely follow the logic -- Claude's response is below. I'll implement accordingly. If you disagree, I'll revert. :-)

----

Good catch — `externalUrl` does flow back to the client through `super.externalUrl(...).toString()` in Jellyfin, Emby, Sonarr/Radarr/Readarr, Overseerr, Nextcloud and Ntfy, so for path-only-bound apps we genuinely want the path to pass through, not be collapsed to `null`.

The literal `app?.href ?? null` revert crashes, though: `externalUrl` is also consumed server-side in `packages/integrations/src/base/integration.ts:85` (`new URL(`${baseUrl}${path}`)`), and a path-only base like `/cockpit/` throws `Invalid URL`. So I'd like to fix this properly rather than swap one crash for another.

**Plan**:

1. Teach `createUrl` to handle path-only bases — when `inputUrl` is path-only, return a small `RenderablePath` shim (`toString()` / `pathname` / `hostname`) instead of constructing a `URL`. 17 of 19 callers just `.toString()` the result; the two outliers (`ntfy-integration.ts:57` reading `.hostname`/`.pathname`, `overseerr-integration.ts:319` returning up to a `.toString()` caller) still behave sensibly.
2. Revert this middleware to `app?.href ?? null` at both call sites (70 and 132) as you suggested — safe once `createUrl` handles path-only.
3. Also patch `packages/request-handler/src/lib/cached-request-integration-job-handler.ts:97`, same `app?.href ?? null` pattern, same crash risk via the same code path. Pre-existing miss in our PR, surfaced while looking at this.

Net effect: integrations bound to path-only-href apps render their hrefs against the user's current origin, instead of falling back to `integration.url` (the server-internal address). Materially better for the multi-hostname use case.

`resolveServerUrl` stays in the ping router — the server genuinely can't ping a hostless URL, so CONFLICT is still right there. Pushing the change in a follow-up commit shortly.

mairas: Implemented. I'll leave the conversation open for your verification.

Meierschlumpf:
Only add translation for locale `en.json`, all other languages are translated on crowdin by the community

mairas:
Reverted the translations.

Meierschlumpf:
IMO we should make it more explicit what the externalUrl and url are. Therefore I'm currently working on some nice typing improvements

mairas:
> IMO we should make it more explicit what the externalUrl and url are. Therefore I'm currently working on some nice typing improvements

So, in practice - I should wait and then rebase and refactor to use the new types?

Meierschlumpf:
I would push my changes into your branch containing the type changes and then we can review it again and if okay, merge it

Meierschlumpf:
Okay nevermind I don't have access to your fork, can you give me access?

mairas:
> Okay nevermind I don't have access to your fork, can you give me access?

Eh. I'm stumped - I usually tick the box while creating the PR, but now that it's not there, I can't find it... Gimme a moment.

mairas:
> Okay nevermind I don't have access to your fork, can you give me access?

Looks like I can't:

```
  Caveats:
  - If the head branch is on an organization fork (e.g., hatlabs/homarr), GitHub disables this option — org-owned forks don't support
  maintainer edits regardless of the checkbox. Workaround: push the branch to a personal fork and open the PR from there.
```

I can reopen the PR from my personal namespace but that'll of course wipe the comments so far.

Meierschlumpf:
Hmm understood, no I think we can keep this open. I'll look into it tomorrow how we get my changes to you, for today I'll leave. See you soon

mairas:
> Hmm understood, no I think we can keep this open. I'll look into it tomorrow how we get my changes to you, for today I'll leave. See you soon

Pull this PR branch into a branch on another fork and do your edits there, then comment here and ask me to pull from that fork.

Meierschlumpf:
I've created a PR on https://github.com/hatlabs/homarr/pull/4

mairas:
Reminder: could you provide your opinion on https://github.com/hatlabs/homarr/pull/4#issuecomment-4596712131, and I'll merge it.

Meanwhile, I'll rebase this PR and https://github.com/hatlabs/homarr/pull/4.

mairas:
**Path-only `externalUrl`: URL fragments are preserved (deliberate)**

Flagging a behavior decision for reviewers. The typed `externalUrl` / `buildUrl` implementation in this PR was developed and reviewed in hatlabs/homarr#4, where one open question was whether a path-only `externalUrl` should preserve the URL **fragment** ([discussion](https://github.com/hatlabs/homarr/pull/4#issuecomment-4596712131)).

Jellyfin and Emby deep-link through `externalUrl` using hash-bang routes, e.g. `/web/index.html#!/details?id=...`. With a full-URL external base the fragment survives (`URL.toString()` keeps it); only the path-only branch was at risk of dropping it. (Plex builds the same hash-bang via `super.url(...)`, so it is unaffected.)

Dropping fragment support for path-only external URLs was on the table, but we decided to **keep it**: the fragment is a valid part of a URL, and dropping it would silently break item-level deep-links for those integrations. The path-only branch appends `result.hash`, covered by a regression test in `packages/integrations/test/base.spec.ts` (`with path-only externalUrl preserves hash-bang fragment`).

Minor note: extra `queryParams` land in the search component (before `#`), not inside the hash. The Jellyfin/Emby callers embed everything in the path and pass no extra params, so this doesn't arise in practice.

packages/validation/src/app.ts
```diff
@@ -1,10 +1,65 @@
...
+//   - empty string -> null
+//   - absolute URL with http/https scheme (or any non-javascript scheme)
+//   - path-only URL starting with "/" followed by a non-"/" character
+//
+// Path-only hrefs are resolved against the current origin in the browser, and
+// against the request origin server-side via `resolveServerUrl`. This lets a
+// single dashboard work across multiple hostnames (mDNS, VPN FQDN, DHCP DNS).
+//
+// Rejects: javascript: scheme, protocol-relative ("//host/..."), single-slash
+// root ("/"), bare strings without scheme or leading slash.
```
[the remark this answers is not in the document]
mairas: Fixed in c23123a9c — corrected the comment: path-only hrefs are browser-resolved only; `resolveServerUrl` returns null for them, which is what drives the `CONFLICT` ping branch.

packages/widgets/src/app/ping/ping-indicator.tsx
```diff
@@ -11,23 +16,48 @@ interface PingIndicatorProps {
...
 
-  const utils = clientApi.useUtils();
+  const [pingResult, setPingResult] = useState<RouterOutputs["widget"]["app"]["ping"]>(ping);
+
   clientApi.widget.app.updatedPing.useSubscription(
     { id: appId },
     {
       onData(data) {
-        utils.widget.app.ping.setData({ id: appId }, data);
+        setPingResult(data);
```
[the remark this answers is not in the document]
mairas: Fixed in 04d9bd05f — keyed `PingIndicatorInner` by `appId` so it remounts with fresh ping state when the app changes, instead of showing the previous app status.

mairas:
Is there _anything_ I could do to get a new review on this PR?

ajnart:
> Is there _anything_ I could do to get a new review on this PR?

We're talking about it internally it's just that we are worried it'll break things. If you agree to be the maintainer for the feature / added to codeowners then we can merge it, 

But there is a risk that it breaks for a larger amount of users than the amount of users who would benefit from that feature or swift their usage to take advantage of it when it is released, the ROI of it is quite low given the amount of code change introduced

mairas:
> We're talking about it internally it's just that we are worried it'll break things. If you agree to be the maintainer for the feature / added to codeowners then we can merge it,
> 
> But there is a risk that it breaks for a larger amount of users than the amount of users who would benefit from that feature or swift their usage to take advantage of it when it is released, the ROI of it is quite low given the amount of code change introduced

Hi, thanks for the reply!

I understand the concern. I don't object being maintainer of the feature; I would notice immediately if it breaks etc, and I'm also committed to using Homarr in Halos (very happy how they work together!).

However, from my point of view, the PR has snowballed quite a bit. Halos only auto-generates and uses app cards, and those are the ones that would need to work with relative URLs for us: a user accessing Homarr locally might be using a different hostname than when accessing Homarr over, say, WireGuard or Tailscale VPN. And in this PR, only the _first commit_ is needed for the app card functionality. The second commit generalizes the relative URLs to other integrations, and @Meierschlumpf's refactorings are only relevant to the other integrations.

Thus, if a partial feature (support relative URLs only for app hrefs) would be deemed an acceptable solution, everything else could be dropped away, and the PR would shrink to a small fraction of its original size, and only touch a small handful of files in a fairly benign way.

The remaining generalizations towards integrations could be mothballed and only revived if issues are filed to request support for them.

What do you think? And yes, if you prefer having a complete solution, I can put my neck on the block for it, if necessary.

mairas:
Here's an alternative minimal PR for kicks: https://github.com/homarr-labs/homarr/pull/6148

PR #5629, opened 2026-05-03
fix(calendar): keep working integrations when one fails

### LINKED ISSUE #5597: bug: Calendar fails completely if any integration associated with it fails
### Describe the bug

The calendar fails completely if any associated integration fails.

For instance, I may have Lidarr and Radarr added as calendar integrations. If either Lidarr or Radarr is down, the whole calendar fails with 'TRPCClientError: Request to integration failed.'

### Steps to reproduce

Shutdown either Lidarr or Radarr, but it can be affected by any integration with the calendar.

### Impact

Cannot use calendar.

### Additional information

The calendar should function whether all associated integrations are up or not.
For instance, if both Lidarr and Radarr are integrated with the calendar, and Radarr is down, the calendar should still work with Lidarr.

### Version

1.59.3

### Installation method

Docker Compose

### Browser

Chrome

mvanhorn:

- [x] Builds without warnings or errors (\`pnpm build\`, autofix with \`pnpm format:fix\`) -- typecheck via \`pnpm --filter @homarr/api run typecheck\` passes; \`pnpm format\` clean
- [x] Pull request targets \`dev\` branch
- [x] Commits follow the [conventional commits guideline](https://www.conventionalcommits.org/en/v1.0.0/) -- \`fix(calendar):\`
- [x] No shorthand variable names are used (eg. \`x\`, \`y\`, \`i\` or any abbrevation)
- [x] Documentation is up to date. Create a pull request [here](https://github.com/homarr-labs/documentation/) -- N/A, internal-only behavior change

---

## Summary

Fixes #5597. When the calendar widget had multiple integrations and any one of them was offline (e.g. Lidarr down while Radarr was up), the whole calendar query rejected with \`TRPCClientError: Request to integration failed.\` -- the rest of the calendar went blank instead of rendering events from the still-working integrations.

## Root cause

\`packages/api/src/router/widgets/calendar.ts:27\` wrapped all integrations in a single \`Promise.all(...)\`. \`Promise.all\` rejects on the first failure, so any single integration outage propagates as a complete tRPC error and the suspense query in \`packages/widgets/src/calendar/component.tsx:42\` never resolves.

\`subscribeToEvents\` (line 54-77) was already iterative -- one \`subscribe\` failure does not affect the others -- so it is unaffected.

## Fix

Switch \`findAllEvents\` to \`Promise.allSettled\`. Fulfilled results are returned as before; rejected results are logged through the existing \`calendarRouter\` logger and dropped from the response. The frontend already groups events per \`integration.id\` and re-renders on subscription updates, so the partial result preserves the existing data shape -- callers see fewer entries in the array, not a thrown error.

\`\`\`ts
const settled = await Promise.allSettled(...)
return settled.flatMap((result, index) => {
  if (result.status === "fulfilled") return [result.value];
  logger.warn("Calendar integration request failed; skipping events for this integration", {...});
  return [];
});
\`\`\`

## Validation

- \`pnpm --filter @homarr/api run typecheck\` -- clean.
- \`pnpm format\` -- clean.
- Manually traced consumer behavior in \`component.tsx\`: failed integration drops out of \`data\`, \`useMemo\` over \`data.flatMap\` produces fewer events, no error throw.

## AI assistance

This PR was written with assistance from Claude (Anthropic). The bug analysis (Promise.all -> Promise.allSettled) and patch were reviewed before submission.

Closes #5597


packages/api/src/router/widgets/calendar.ts
```diff
@@ -40,6 +43,19 @@ export const calendarRouter = createTRPCRouter({
...
+      return settled.flatMap((result, index) => {
+        if (result.status === "fulfilled") {
+          return [result.value];
+        }
+        const integration = ctx.integrations[index];
+        logger.warn("Calendar integration request failed; skipping events for this integration", {
+          integrationId: integration?.id,
+          integrationKind: integration?.kind,
+          error: result.reason instanceof Error ? result.reason.message : String(result.reason),
+        });
```
Meierschlumpf: I would suggest to use `ErrorWithMetadata` so we can put the `result.reason` in the `cause` of it

Meierschlumpf (changes requested):
Beside this great change, thanks ❤️

mvanhorn:
Done in 9ff2aabf - the per-integration warn now uses `ErrorWithMetadata` with `result.reason` as the cause, matching the existing pattern in `apps/nextjs/src/app/api/trpc/[trpc]/route.ts` and `apps/tasks/src/main.ts`. Local `tsc --noEmit`, prettier, and eslint all pass on `packages/api`.

manuel-rw (approved):
lgtm besides the other comment from Meier

Meierschlumpf (approved):

Issue #5342, opened 2026-03-27
bug: Homarr generates malformed DNS queries (<host>https) for Tracearr integration


mshorey:
### Describe the bug

When configuring a Tracearr integration, Homarr repeatedly generates malformed DNS queries where the protocol string (https) is appended to the hostname. This results in DNS lookups like:

tracearr.mydomain.nethttps

This causes continuous DNSSEC validation failures and SERVFAIL responses from the DNS server.

### Steps to reproduce

1. Add a Tracearr integration in Homarr
2. Set URL to: https://tracearr.mydomain.net
3. Save and enable integration
4. Add Tracearr widget to dashboard
5. Observe DNS logs on upstream resolver (Blocky in this case)

### Impact

Excessive query load on DNS.

### Additional information

<img width="1247" height="746" alt="Image" src="https://github.com/user-attachments/assets/c4796455-5d77-476f-a06f-773c6a821ce1" />

### Version

1.56.1

### Installation method

Docker Compose

### Browser

Brave

manuel-rw:
Can you try disabling DNS caching?
https://homarr.dev/docs/advanced/environment-variables/#general

mshorey:
> Can you try disabling DNS caching? https://homarr.dev/docs/advanced/environment-variables/#general

I just tried that but it doesn't look to have changed behavior.  Homarr keeps appending https to the end of the hostname.

manuel-rw:
@StrandedTurtle can you take a look?

StrandedTurtle:
I will try and take a look soon, although I haven't had the issue yet, I'll try to recreate it next time I get a chance

StrandedTurtle:
Have patched this here: 
https://github.com/homarr-labs/homarr/pull/5654

PR #5654, opened 2026-05-06
fix(tracearr): support absolute image urls

StrandedTurtle:
When Tracearr returns absolute URLs for avatars or posters, the `proxyImageAsync` method incorrectly passed them through
  `this.url()`, concatenating the integration base URL with the absolute URL. This produced malformed URLs like
  `https://tracearr.mydomain.nethttps://tracearr.mydomain.net/...`, causing repeated DNS lookups for
  `tracearr.mydomain.nethttps`.

https://github.com/homarr-labs/homarr/issues/5342


packages/integrations/src/tracearr/tracearr-integration.ts
```diff
@@ -208,7 +208,9 @@ export class TracearrIntegration extends Integration {
       // This is critical because ImageProxy uses bcrypt which truncates at 72 bytes.
       // Without this, all long avatar/poster URLs from the same session hash identically
       // since they share the same first 72+ bytes.
-      const baseUrl = this.url(cleanUrl as `/${string}`).toString();
+      const baseUrl = cleanUrl.startsWith("http://") || cleanUrl.startsWith("https://")
```
manuel-rw: Can't we remove ``http://`` and ``https://`` using regex and then just pass the resulting value to ``this.url()``?

manuel-rw (changes requested):

manuel-rw (dismissed):

ajnart (approved):

manuel-rw (approved):

PR #5714, opened 2026-05-15
feat(board): persist category collapse state in localStorage for guests

### LINKED ISSUE #3355: feat: save category state in local storage
### Describe the bug

I have my Homarr icons organised in categories per host system. When I visit my Homarr instance all categories are always shrunken (not expanded). It used to be the case that the default was to expand all categories. When I expand the categories on any browser and reload they are shrunken again.

### Steps to reproduce

1. Create a dashboard with categories add some apps
2. Open the page, categories are shrunken
3. Expand with down-arrow of category
4. Reload
5. Rinse, repeat

### Impact

It's just very annoying!

### Additional information

Tested on multiple browsers, both desktop and mobile.

### Version

1.23.0

### Installation method

Portainer

### Browser

Firefox

Gilmoursa:
## What does this PR do?

Extends category collapse state persistence to unauthenticated (guest) users via `localStorage`.

**Before:** Guests could collapse/expand categories but the state was lost on reload (the server-side `changeCollapsed` mutation requires authentication).

**After:** Guest users get the same persistence behaviour as signed-in users — their per-section collapse state is stored in `localStorage` under `homarr-section-collapsed-<sectionId>` and restored on page load.

## How it works

- On mount, if `status === "unauthenticated"`, reads localStorage and applies the stored state via `open()`/`close()`
- On toggle, guests write to localStorage; authenticated users continue using the existing protected mutation
- No change to behaviour for signed-in users
- The `status === "loading"` case is intentionally ignored — the component starts with the server-provided default (`collapsed: false`) and only syncs once auth state resolves

## Related issue

Closes #3355

### DIFF
--- a/apps/nextjs/src/components/board/sections/category-section.tsx
+++ b/apps/nextjs/src/components/board/sections/category-section.tsx
@@ -1,8 +1,10 @@
+import { useEffect } from "react";
 import { Card, Collapse, Group, Stack, Title, UnstyledButton } from "@mantine/core";
 import { useDisclosure } from "@mantine/hooks";
 import { IconChevronDown, IconChevronUp } from "@tabler/icons-react";
 
 import { clientApi } from "@homarr/api/client";
+import { useSession } from "@homarr/auth/client";
 import { useRequiredBoard } from "@homarr/boards/context";
 
 import type { CategorySection } from "~/app/[locale]/boards/_types";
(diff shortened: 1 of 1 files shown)


ajnart (dismissed):
I think it's a neet change, using localstorage for small preferences makes sense

apps/nextjs/src/components/board/sections/category-section.tsx
```diff
@@ -1,31 +1,51 @@
...
+      }
     },
   });
 
+  useEffect(() => {
+    if (status !== "unauthenticated") return;
+    const stored = localStorage.getItem(localStorageKey(section.id));
+    if (stored === "true") open();
+    else if (stored === "false") close();
+  }, [status, section.id, open, close]);
```
Meierschlumpf: Not a fan of using `useEffect` here, can't we set it as default value directly?

Meierschlumpf (changes requested):

Gilmoursa:
Thanks for the feedback! I've removed the `useEffect` and now compute the initial value directly from localStorage, passing it to `useDisclosure` as the default.

One thing worth flagging: since `initialCollapsed` is now read during render rather than post-mount, there is a potential hydration mismatch on SSR if the localStorage value differs from the server-provided `section.collapsed`. The `typeof window !== "undefined"` guard handles the SSR case itself, but wanted to make sure you were comfortable with that tradeoff before this is merged.

Gilmoursa:
Hi @Meierschlumpf, just flagging that I've addressed your feedback. The `useEffect` has been removed and the initial value is now computed directly from localStorage and passed to `useDisclosure` as the default. Happy to make any further changes if needed!

ajnart (approved):

Meierschlumpf (approved):

manuel-rw (approved):

PR #5735, opened 2026-05-18
perf(boards): remove loading overlay and show board immediately

ajnart:
## Summary
- Remove the `LoadingOverlay` and hidden-until-ready visibility gating from the board client — the board now paints as soon as the server-rendered content arrives
- Remove dead duplicate `updateBoardName` / `useUpdateBoard` exports from `_client.tsx` (already live in `@homarr/boards/updater`)
- Remove unused `isReady` / `useIsBoardReady` from the ready context (only consumer was the overlay)

## Test plan
- [ ] Open a board with multiple sections — should render immediately without a spinner
- [ ] Edit mode still works (section actions still call `markAsReady`)
- [ ] No layout shift or flash of empty content on slow connections

### DIFF
--- a/apps/nextjs/src/app/[locale]/boards/(content)/_client.tsx
+++ b/apps/nextjs/src/app/[locale]/boards/(content)/_client.tsx
@@ -1,70 +1,27 @@
 "use client";
 
-import { useCallback, useRef } from "react";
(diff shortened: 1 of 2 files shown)


manuel-rw (approved):
lgtm, tested on preview deployment, consider waiting for @Meierschlumpf

PR #5736, opened 2026-05-18
perf(boards): parallelize board page data fetching

ajnart:
## Summary
- Run `getInitialBoard` and `auth()` concurrently instead of sequentially
- Run all widget prefetches (`prefetchForKindAsync`) in parallel alongside integration permission loading

Single file change in `_creator.tsx`.

## Test plan
- [ ] Board page loads correctly with widgets and integrations
- [ ] No regressions on home board or named boards


manuel-rw (dismissed):
lgtm, but please test this thoroughly before merging

Meierschlumpf (dismissed):
If well tested, this is a good change, thanks

apps/nextjs/src/app/[locale]/boards/(content)/_creator.spec.ts
Comment (Meierschlumpf) [inline apps/nextjs/src/app/[locale]/boards/(content)/_creator.spec.ts:1]: Sorry but there is no value in these tests, with well tested I meant that it has been tested manually that it is working and is good user experience

Meierschlumpf (changes requested):

apps/nextjs/src/app/[locale]/boards/(content)/_creator.spec.ts
Comment (ajnart) [reply apps/nextjs/src/app/[locale]/boards/(content)/_creator.spec.ts:1]: Yeah.. on my tests of the basic ui interactions on the docker build as a ose-user of Homarr it's fine with multiple boards but I haven't really thought about a way to test it deeper than "it works on my machine for my usecase"

Comment (Meierschlumpf) [reply apps/nextjs/src/app/[locale]/boards/(content)/_creator.spec.ts:1]: I think that's probably enough for these changes 👍🏼

Meierschlumpf (approved):

PR #5737, opened 2026-05-18
feat(boards): cache named board data with tag-based invalidation

ajnart:
## Summary
- Add `getSharedBoardWithWhereAsync` — fetches board structure without per-user data (permissions, collapse states), safe to cache across requests
- Wrap named board fetches in `unstable_cache` with a 1h TTL and `board-{name}` tag
- Add `invalidateBoardCacheAsync` server action that busts cache on board save, layout save, and partial settings mutations
- Home board path is unchanged (no caching, still uses tRPC `getHomeBoard`)

Depends on #5736.

## Test plan
- [ ] Named board loads from cache on repeat visits (check server logs for fewer DB queries)
- [ ] Editing a board (save, settings, layout) invalidates cache — next load reflects changes
- [ ] Home board still works without caching
- [ ] Anonymous users see the correct board data


manuel-rw (dismissed):
this looks good to me, but wait for @Meierschlumpf 's approval.
What consequences does this have, when the cache is outdated? Does it cause race conditions when saving?

Meierschlumpf (dismissed):
Based on my understanding these changes will result in a few problems:
1. User specific data is no longer delivered, so groups / permissions of the user and collapse states will no longer be available.
2. ~⚠️ **Anybody** can access even **private** boards as there no longer is a `throwIfActionForbiddenAsync` with `view` permission~ - at least if the `_layout-creator` would not check for it additionally. You could consider caching the values from getInitialBoard for the specific request so it does not send 3 calls to that function (layout, page and metadata) instead and maybe also improve the `throwIfActionForbiddenAsync` which is currently also sending db requests for info that could be available through the requested board

ajnart:
> Based on my understanding these changes will result in a few problems:
> 
> 1. User specific data is no longer delivered, so groups / permissions of the user and collapse states will no longer be available.
> 2. ~⚠️ **Anybody** can access even **private** boards as there no longer is a `throwIfActionForbiddenAsync` with `view` permission~ - at least if the `_layout-creator` would not check for it additionally. You could consider caching the values from getInitialBoard for the specific request so it does not send 3 calls to that function (layout, page and metadata) instead and maybe also improve the `throwIfActionForbiddenAsync` which is currently also sending db requests for info that could be available through the requested board

I replaced the unstable_cache approach with React cache() for per-request deduplication. Layout, page, and metadata now share one tRPC call (3 DB queries instead of 9), while keeping the full auth check and all user-specific data intact

PR #5739, opened 2026-05-18
feat(seed): add demo mode with pre-seeded user

ajnart:
## Summary

- **Demo mode seed**: When `DEMO_MODE=true`, the seed creates a `demo` / `demo` user with admin permissions, a default board, and completed onboarding — ready to use out of the box.
- **`.dockerignore`**: Added `.env` and `apps/nextjs/.next` so local dev config and build cache don't leak into the Docker build context. The local `.env` (with `DB_DRIVER=better-sqlite3`) was causing `next build` to resolve a production default DB path (`/appdata/db/db.sqlite`) whose directory doesn't exist in the builder stage.

## Test plan

- [x] `docker build . -t homarr:demo` completes successfully
- [x] Container starts with `DEMO_MODE=true`, seed logs show `Created demo user through seeding`
- [x] Login with `demo` / `demo` succeeds and lands on the default board

### DIFF
--- a/packages/db/migrations/seed.ts
+++ b/packages/db/migrations/seed.ts
@@ -273,5 +273,7 @@ const seedDemoUserAsync = async (db: Database) => {
   });
   await db.update(users).set({ homeBoardId: boardId }).where(eq(users.id, userId));
 
-  console.log("Your instance has enabled demo mode. Homarr therefore created example data for your demo. You can disable demo mode anytime by setting the env 'DEMO_MODE' to 'false' or unset it.");
+  console.log(
+    "Your instance has enabled demo mode. Homarr therefore created example data for your demo. You can disable demo mode anytime by setting the env 'DEMO_MODE' to 'false' or unset it.",
+  );
 };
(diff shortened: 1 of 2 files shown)


packages/db/migrations/seed.ts
```diff
@@ -191,3 +208,70 @@ const seedDefaultIntegrationsAsync = async (db: Database) => {
...
+  await db.insert(layouts).values({
+    id: createId(),
+    name: "Base",
+    columnCount: 12,
+    breakpoint: 0,
+    boardId,
+  });
+  await db.update(users).set({ homeBoardId: boardId }).where(eq(users.id, userId));
+
+  console.log("Created demo user through seeding");
```
manuel-rw:
```suggestion
  console.log("Your instance has enabled demo mode. Homarr therefore created example data for your demo. You can disable demo mode anytime by setting the env 'DEMO_MODE' to 'false' or unset it.");
```

manuel-rw (dismissed):

PR #5740, opened 2026-05-18
ci: add --affected filtering and fix Turbo cache invalidation

ajnart:
## Summary

Fixes the Turborepo configuration that was causing near-100% cache miss rates and adds `--affected` filtering so CI only processes packages that actually changed.

### Turbo cache fixes
- **Removed `globalDependencies: ["**/.env"]`** — this was invalidating every cache entry across all packages whenever any `.env` file changed
- **Moved 40+ env vars from `globalEnv` to `build.env`** — lint, format, and typecheck don't need `DB_URL`, `AUTH_LDAP_*`, etc. Only `CI`, `NODE_ENV`, and `SKIP_ENV_VALIDATION` remain global
- **Added per-task `inputs`** with `$TURBO_DEFAULT$` and test file exclusions — changing a test file no longer triggers build or typecheck
- **Enabled `affectedUsingTaskInputs`** — `--affected` now operates at task level (which specific task inputs changed) not just package level

### CI workflow changes
- **Added `--affected`** to all turbo commands (lint, format, typecheck, build) — a PR touching 1 package runs checks on ~1-3 packages instead of ~30
- **Added `fetch-depth: 0`** to all jobs using `--affected` (required for git history comparison)
- **Added Docker Buildx + GHA layer caching** to E2E job (`cache-from: type=gha`, `cache-to: type=gha,mode=max`)
- **Added `SKIP_ENV_VALIDATION: true`** as workflow-level env

### Expected impact
- Most PRs: CI feedback in ~2-5 min instead of ~15-30 min
- Typecheck (previously the slowest job): only runs on affected packages
- E2E Docker build: cached layers on subsequent runs

## Test plan
- [ ] CI passes on this PR (validates the turbo.json changes don't break anything)
- [ ] Verify `--affected` correctly identifies changed packages (check turbo output logs)
- [ ] Verify E2E Docker cache works on a follow-up push to this PR
- [ ] Verify typecheck only runs on affected packages by checking turbo output


.github/workflows/code-quality.yml
```diff
@@ -70,28 +83,30 @@ jobs:
         run: pnpm test
 
       - name: "Report Coverage"
-        # Set if: always() to also generate the report if tests are failing
-        # Only works if you set `reportOnFailure: true` in your vite config as specified above
```
manuel-rw, line 74: Why remove this?

manuel-rw (dismissed):

turbo.json
```diff
@@ -1,73 +1,38 @@
 {
   "$schema": "https://turborepo.org/schema.json",
-  "globalDependencies": ["**/.env"],
-  "globalEnv": [
```
ajnart, line 4: Basically this was holding us back as far as I understand it was baking the hash of the .env into the build and thus forcing cache-miss every time

.env.example
ajnart, line 1: I deployed https://github.com/AdiRishi/turborepo-remote-cache-cloudflare on a cloudflare worker and set these variables for the project (using gh cli) , it uses r2 to store build artifacts and since we have 15gb per month of free storage it should be more than enough for our needs. 
Ideally we'd have a hook / vscode config to build the code easily (and fast) so that ci takes even less time because it would leverage that cache

.github/workflows/code-quality.yml
ajnart: I'll put it back, I just don't like comments x)

turbo.json
```diff
@@ -1,73 +1,38 @@
...
       "outputLogs": "new-only"
     },
     "lint": {
       "dependsOn": ["^topo"],
+      "inputs": ["$TURBO_DEFAULT$"],
       "outputs": ["node_modules/.cache/.eslintcache"]
     },
     "typecheck": {
       "dependsOn": ["^topo"],
+      "inputs": ["$TURBO_DEFAULT$", "!**/*.test.ts", "!**/*.test.tsx", "!**/*.spec.ts", "!**/__tests__/**"],
```
Meierschlumpf: We shpuld not exclude test files from typecheck

turbo.json
```diff
@@ -1,73 +1,38 @@
...
-    "DOCKER_PORTS",
-    "DOCKER_SOCKET_PATHS",
-    "NODE_ENV",
-    "PORT",
-    "LOCAL_CERTIFICATE_PATH",
-    "LOG_LEVEL",
-    "SECRET_ENCRYPTION_KEY",
-    "SKIP_ENV_VALIDATION"
-  ],
+  "remoteCache": { "signature": false },
```
Meierschlumpf, line 3: I guess this is okay because we do not share the cache token publically?

Meierschlumpf (changes requested):

turbo.json
ajnart: Yes, and from what I saw this didn't work with my self-hosted endpoint for caching. I'll share my env privately if you want to leverage that caching locally, but it should work on the CI

PR #5741, opened 2026-05-18
refactor: replace ESLint + Prettier with OXC (oxlint + oxfmt)

ajnart:
## Summary

Nukes the entire ESLint + Prettier toolchain and replaces it with OXC (oxlint + oxfmt) — the Rust-based toolchain by VoidZero (Evan You).

### What's removed
- **37 `eslint.config.js` files** across the monorepo
- **`tooling/eslint/`** source files (base.js, react.js, nextjs.js, types.d.ts)
- **`tooling/prettier/`** source files (index.mjs)
- **ESLint, Prettier, and all related plugins** from every package.json (410 npm packages removed from the dep tree)
- **Catalog entries** for eslint, prettier, typescript-eslint, eslint-plugin-*, prettier-plugin-*

### What's added
- **`.oxlintrc.json`** — single root config with all rules mapped from the old ESLint setup (no-unused-vars, consistent-type-imports, no-non-null-assertion, id-length, import/consistent-type-specifier-style + React, Next.js, import, jsx-a11y plugins)
- **`.oxfmtrc.json`** — single root config (printWidth: 120, endOfLine: lf, sortPackageJson)
- **`oxlint` (1.64.0) + `oxfmt` (0.49.0)** as root devDeps
- **`lint-format-autofix` CI job** that auto-commits lint/format fixes on PRs instead of blocking

### Performance
- **oxlint**: 77ms on 45 files (vs ESLint's multi-second type-checked runs)
- **oxfmt**: 26ms on 45 files
- **No per-package config needed** — one root config covers everything

### DX improvements
- No more `eslint.config.js` in every package
- No more `"prettier": "@homarr/prettier-config"` in every package.json
- CI auto-fixes formatting instead of blocking PRs
- `pnpm lint` and `pnpm format` just work from root

### What's NOT changed
- `tsc --noEmit` typecheck stays as-is (oxlint type-aware mode is experimental, can be added later)
- The `no-restricted-syntax` custom selectors (async naming convention, homarr.dev links) are dropped for now — can be re-added via `oxlint-plugin-eslint` jsPlugins when it stabilizes

## Test plan
- [ ] CI passes (validate oxlint/oxfmt run correctly)
- [ ] `pnpm lint` works locally
- [ ] `pnpm format` works locally
- [ ] Autofix job commits formatting changes on a test push
- [ ] Verify no ESLint/Prettier references remain in the codebase


turbo.json
```diff
@@ -1,73 +1,75 @@
...
       "outputLogs": "new-only"
     },
     "lint": {
       "dependsOn": ["^topo"],
-      "outputs": ["node_modules/.cache/.eslintcache"]
+      "inputs": ["$TURBO_DEFAULT$"]
     },
     "typecheck": {
       "dependsOn": ["^topo"],
+      "inputs": ["$TURBO_DEFAULT$", "!**/*.test.ts", "!**/*.test.tsx", "!**/*.spec.ts", "!**/__tests__/**"],
```
Meierschlumpf: What is the reason for this?

.oxfmtrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "./node_modules/oxfmt/configuration_schema.json",
+  "printWidth": 120,
+  "endOfLine": "lf",
+  "sortPackageJson": false,
```
Meierschlumpf: I think we previously had this one enabled

tooling/eslint/package.json
Meierschlumpf, line 1: Package should probably be removed (same for prettier)

Meierschlumpf (commented):
Beside that, good change 👍🏼

tooling/eslint/package.json
ajnart: yup, cleaned it up

.oxfmtrc.json
ajnart: i didn't turn it on because it would've formatted a bunch of files but I guess that's fine

turbo.json
ajnart: it was a remanant from the other pr for the turbo cache, i removed it (and removed the tests exclusion from the list)

packages/api/package.json
```diff
@@ -4,6 +4,8 @@
   "private": true,
   "license": "Apache-2.0",
   "type": "module",
+  "main": "./index.ts",
+  "types": "./index.ts",
```
Meierschlumpf: Shouldn't this be `./src/index.ts` instead? There is no `index.ts` file in root directory (check for other packages as well)

packages/db/package.json
```diff
@@ -4,6 +4,8 @@
   "private": true,
   "license": "Apache-2.0",
   "type": "module",
+  "main": "./index.ts",
+  "types": "./index.ts",
```
Meierschlumpf, line 8: Is there a reason these only exist for certain packages? Consider to either revert adding them or add them to all packages or describe why it has been added to only some of them

static-data/opencollective-contributors.json
Meierschlumpf, line 1: Exclude static files from formating (as they are generated automatically and should be as small as possible)

turbo/generators/templates/package.json.hbs
Meierschlumpf, line 1: We could consider to remove the turbo generators as I have never really used them (also because the versions specified were never updated), therefore I've always just copied the existing files from another package

.deepsource.toml
```diff
@@ -3,9 +3,9 @@ version = 1
...
-  plugins = ["react"]
-  environment = ["nodejs"]
+[analyzers.meta]
+plugins = ["react"]
+environment = ["nodejs"]
 
 [[transformers]]
-name = "prettier"
\ No newline at end of file
+name = "prettier"
```
Meierschlumpf: I guess we can remove this one and rely on the auto-fix if needed (oxfmt is not supported: https://docs.deepsource.com/docs/platform/reference/code-formatters)

CHANGELOG.md
Meierschlumpf, line 1: I would consider to also ignore the format of CHANGELOG.md as it is also generated automatically

pnpm-workspace.yaml
```diff
@@ -33,7 +31,7 @@ catalog:
   "@mantine/notifications": ^8.3.18
   "@mantine/spotlight": ^8.3.18
   "@mantine/tiptap": ^8.3.18
-  "@next/eslint-plugin-next": 16.2.6
+  "@million/lint": 1.0.14
```
Meierschlumpf: This one should not be added again

Meierschlumpf (changes requested):

pnpm-workspace.yaml
ajnart: I somehow based this PR on an older commit and it slipped through, sorry about that

.deepsource.toml
ajnart: Yup I removed the block

static-data/opencollective-contributors.json
ajnart: I excluded the whole static-data folder from the formatting

packages/db/package.json
ajnart: This is just the case of formatting, the keys got reordered when I added the re-ordering commit so the diff looks like an addition but it's just moving things around

Meierschlumpf: You're right, sorry I missed that

Meierschlumpf (dismissed):

manuel-rw:
I generally agree with this change, but is OXC future safe? Or will we have to migrate again in 2 weeks? :D

ajnart:
> I generally agree with this change, but is OXC future safe? Or will we have to migrate again in 2 weeks? :D

Since it's part of Vite+ / Voidzero I'm bullish on it being future safe

PR #5751, opened 2026-05-19
fix: parse Speedtest Tracker created_at as local time

VIVAAN-DHAWAN:
Fixes #5744

## Summary
- Parse Speedtest Tracker `created_at` as a local wall-time value instead of forcing it through UTC with a trailing `Z`.
- Add a regression spec that reproduces the Berlin example from the issue in a child process with `TZ=Europe/Berlin`.

## Reproduction before fix
- `pnpm exec vitest run packages/integrations/src/speedtest-tracker/test/speedtest-tracker-types.spec.ts --coverage.enabled=false` failed because `2026-05-18 18:30:01` parsed as `20:30` in Europe/Berlin.

## Validation after fix
- `pnpm exec vitest run packages/integrations/src/speedtest-tracker/test/speedtest-tracker-types.spec.ts --coverage.enabled=false` -> 1 test passed.
- `pnpm -F @homarr/integrations typecheck` -> passed.
- `pnpm -F @homarr/integrations lint` -> passed.
- `pnpm -F @homarr/integrations format` -> all matched files use Prettier style.
- `git diff --check` -> passed.

PR #5753, opened 2026-05-19
fix: [5744] Incorrect Speed Test Tracker parsing

### LINKED ISSUE #5744: Speedtest Tracker integration: timestamps shown in UTC instead of local time
### Describe the bug

The Speedtest Tracker widget displays timestamps in UTC instead of the local timezone. 
The Homarr integration reads `data.timestamp` from the Speedtest Tracker API, which 
contains the raw UTC timestamp from Ookla (e.g. `"2026-05-18T16:30:38Z"`), instead 
of using the `created_at` field which already contains the correct local time.

### Steps to reproduce

1. Set up Speedtest Tracker with `APP_TIMEZONE=Europe/Berlin` and `DISPLAY_TIMEZONE=Europe/Berlin`
2. Add the Speedtest Tracker integration in Homarr
3. Run a speedtest at e.g. 18:30 CEST
4. Homarr displays 16:30 (UTC) instead of 18:30 (CEST)

### Impact

Low impact, the widget still works but timestamps are always 2 hours behind 
(UTC instead of CEST). Workaround: none known.

### Additional information

Timestamps should be displayed in the local timezone configured in Speedtest Tracker, 
or the `created_at` field from the API should be used instead of `data.timestamp`.

Speedtest Tracker version:  v1.14.1 
Running on: Proxmox LXC (bare metal, not Docker)
APP_TIMEZONE=Europe/Berlin 
DISPLAY_TIMEZONE=Europe/Berlin 
LXC system timezone: Europe/Berlin (CEST, UTC+2) 

API response example:
- data.timestamp: "2026-05-18T16:30:38Z"  ← UTC from Ookla (wrong)
- created_at: "2026-05-18 18:30:01"       ← correct local time

### Version

1.61.0

### Installation method

### Browser

Firefox

Deses:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

Fixes https://github.com/homarr-labs/homarr/issues/5744

In my case STT widget was always showing timestamps 2 hours ahead of what STT itself displayed. So if the test ran at 00:42 local time, Homarr was showing 02:42.

The created_at field from the STT API looks like this: "2026-05-19 00:42:01", which is already a local-time string with no timezone.

The old cdoe was doing this:
```
const parseTimestamp = (timestamp: string): Date => new Date(`${timestamp.replace(" ", "T")}Z`);
created_at: z.string().transform(parseTimestamp),
```
Which converted that string into a Date with a UTC timezone. The browser then localized that UTC date to my local timezone, UTC+2 in this case and that's why i was always 2 hours ahead.

The fix is to not convert it to a Date on the server at all. Just normalize the format (replace the space with a T so it's valid ISO) and send the string as is to the browser, and the browser converts the timezone.

The four new Date() calls were added because created_at is now a string and `.getTime()` needs a Date or explodes.

image
image

Now if only I could choose to display it in 24h mode


packages/integrations/src/speedtest-tracker/speedtest-tracker-types.ts
```diff
@@ -2,15 +2,13 @@ import { z } from "zod/v4";
...
-const parseTimestamp = (timestamp: string): Date => new Date(`${timestamp.replace(" ", "T")}Z`);
-
 export const speedtestTrackerResultSchema = z.object({
   id: z.number().int(),
   ping: z.number().nullable(),
   download_bits: z.number().int().nullish(),
   upload_bits: z.number().int().nullish(),
   healthy: z.boolean().nullable(),
-  created_at: z.string().transform(parseTimestamp),
+  created_at: z.string().transform((ts) => ts.replace(" ", "T")),
```
manuel-rw, line 11: Could you please transform this using zod to a Date instead of in the component?
Then you can revert most other changes.

manuel-rw (changes requested):

manuel-rw:
Also it seems that this is a duplicate of #5751 . Can we close this?

Deses:
> Also it seems that this is a duplicate of #5751 . Can we close this?

Yep, we are doing largely the same thing.

PR #5764, opened 2026-05-21
fix(auth): relax password complexity requirements to suggestions-only

ajnart:
## Summary

Relaxes password complexity validation to suggestions-only while keeping the minimum 8-character requirement enforced.

### Changes

- `userPasswordSchema` now only enforces `min(8).max(255)` — no more rejection for missing uppercase, lowercase, number, or special chars
- Password popover shows neutral indicators (gray dot) for unmet suggestions instead of red X
- Added "Password suggestions" title to the popover
- Created `UserCreatePasswordFields` shared component used in onboarding and user management
- Updated translation labels to reflect "suggested" wording
- Controlled/uncontrolled input fix for `CustomPasswordInput`

### Test plan

- [x] Unit test for passwords without complexity (8+ chars) pass validation  
- [x] `LoveHomarr<3` hashes correctly with bcrypt
- [x] E2E onboarding with special character password succeeds


manuel-rw (dismissed):
generally looks good to me, but @Meierschlumpf should review your redis changes

Meierschlumpf:
I never had such an issue, can you explain the way to reproduce the sqlite issue?
Also can we consider not putting things into this PR that are unrelated to password / auth? Like the changes in Dockerfile, regarding redis and the many null checks that seem unnecessary to me

Meierschlumpf (changes requested):
See above comment

ajnart:
@Meierschlumpf I honestly couldn't reproduce it again either, and AUTH_SESSION_EXPIRY_TIME fell-back to the default "30d". The createDurationSchema issue covers the case where someone passes a bare number (e.g. "86400"), but that wasn't the trigger here, I had this issue consistently at the time of the PR but I am not able to reproduce it anymore on the :latest image.. 🤷🏻‍♂️ 

At that time, I was doing a git stash and applying it to get the app working locally

We can disregard the changes to this if it's not relevant, but it might save someone that has a weird issue like mine. Perhaps it's a strange macos issue/a package update fixed it. 

We can only keep the password-related changes if you think that's better

Meierschlumpf:
Okay understand, thanks
Yeah I would suggest that we only change the password requirements stuff then, if the other issue happens again we can look into it again, but I think it might have been just a local development issue due to either a package / migrations in the db that caused it

packages/auth/test/security.spec.ts
```diff
@@ -53,4 +53,12 @@ describe("comparePasswordsAsync should compare passwords correctly", () => {
...
     expect(result).toBe(false);
   });
+  it("should return true for passwords with special characters like LoveHomarr<3", async () => {
+    const password = "LoveHomarr<3";
+    const hash = await hashPasswordAsync(password);
+
+    const result = await comparePasswordsAsync(password, hash);
+
+    expect(result).toBe(true);
+  });
```
Meierschlumpf: I don't see a benefit in this test as hashing and comparing should work for any password

packages/translation/src/lang/en.json
```diff
@@ -132,12 +132,13 @@
...
-          "length": "Includes at least 8 characters",
-          "lowercase": "Includes lowercase letter",
-          "uppercase": "Includes uppercase letter",
-          "number": "Includes number",
-          "special": "Includes special symbol"
+          "length": "At least 8 characters",
+          "lowercase": "Suggested: includes lowercase letter",
+          "uppercase": "Suggested: includes uppercase letter",
+          "number": "Suggested: includes number",
+          "special": "Suggested: includes special symbol"
```
Meierschlumpf: What about changing these from `requirement` to `suggestion` object, because otherwise in other languages these changes will not really happen

e2e/onboarding.spec.ts
```diff
@@ -42,6 +42,37 @@ describe("Onboarding", () => {
     await homarrContainer.stop();
   }, 60_000);
 
+  test("Credentials onboarding with special character password should be successful", async () => {
```
Meierschlumpf: What is the difference to the test above?

packages/ui/src/components/password-input/password-input.tsx
```diff
@@ -20,15 +20,16 @@ export const CustomPasswordInput = ({ withPasswordRequirements, ...props }: Cust
...
 
   const onChange = (event: ChangeEvent<HTMLInputElement>) => {
-    setValue(event.currentTarget.value);
+    setInternalValue(event.currentTarget.value);
     props.onChange?.(event);
   };
 
   return (
-    <PasswordRequirementsPopover password={value}>
+    <PasswordRequirementsPopover password={String(password)}>
```
Meierschlumpf: Why shouldn't this be a string?

packages/ui/src/components/password-input/password-input.tsx
```diff
@@ -20,15 +20,16 @@ export const CustomPasswordInput = ({ withPasswordRequirements, ...props }: Cust
 };
 
 const WithPasswordRequirements = (props: PasswordInputProps) => {
-  const [value, setValue] = useState("");
+  const [internalValue, setInternalValue] = useState("");
+  const password = props.value ?? internalValue;
```
Meierschlumpf: What is the benefit / reason for using the value from outside?

packages/ui/src/components/password-input/password-requirement.tsx
```diff
@@ -1,14 +1,23 @@
...
+
+const requirementDisplay = {
+  met: { color: "teal", Icon: IconCheck },
+  unmet: { color: "dimmed", Icon: IconPoint },
+} as const;
+
+const displayKey = {
+  true: "met",
+  false: "unmet",
+} as const;
```
Meierschlumpf: I would simplify this by converting `requirementDisplay` into a function that returns the met values for true as input and unmet values for false as input.

Meierschlumpf (changes requested):

manuel-rw (approved):

PR #5600, opened 2026-04-30
perf(memory): combine 3 node processes into one

Meierschlumpf:
<div align="center">
  
  <h3>Homarr</h3>
</div>

**Thank you for your contribution. Please ensure that your pull request meets the following pull request:**

- [x] Builds without warnings or errors (`pnpm build`, autofix with `pnpm format:fix`)
❌ Pull request targets `dev` branch
- [x] Commits follow the [conventional commits guideline](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] No shorthand variable names are used (eg. `x`, `y`, `i` or any abbrevation)
- [x] Documentation is up to date. Create a pull request [here](https://github.com/homarr-labs/documentation/).

Credits to #5562 
Closes #3759

- During idle between 200 and 300MB
- During actual usage (with some integrations etc.) max 500MB

Usage before: 600MB - 1GB

PR #5765, opened 2026-05-21
fix: resolve Redis null client crashes on startup

ajnart:
> [!WARNING]
> This was completely vibe-coded and I'm not 100% sure it works in all scenarios. Please review carefully.

## Problem

After #5600 merged the tasks/websocket processes into Next.js, the server crashes on every startup with:

```
TypeError: Cannot read properties of null (reading 'publish')
TypeError: Cannot read properties of null (reading 'get')
```

These come from `invalidateUpdateCheckerCache` and the `analytics` cron job (`runOnStart: true`), both of which try to use Redis immediately.

## Root cause

`createRedisConnection()` returns `null` when `CI=true`, which is always set during `pnpm build` / Dockerfile build. Next.js inlines this at build time, so the bundled server code permanently has null Redis clients. This was fine when tasks ran as a separate esbuild bundle, but now that they're in-process via `instrumentation.ts`, the null gets baked in.

## Fix

1. **Remove the null guard** from `createRedisConnection()` — always return a real client
2. **Lazy-init Redis clients** in `channel.ts` and `channel-subscription-tracker.ts` so they're created at runtime, not at module evaluation during build
3. **Wait for Redis readiness** in `main.ts` before starting cron jobs (retry loop with `handshakeAsync`)

The `DISABLE_REDIS_LOGS` guard remains in `packages/core/src/infrastructure/logs/transports/index.ts` where it belongs (skipping Redis log transport during DB migrations).

## Test plan

- [x] `pnpm tsc --noEmit` passes for both `@homarr/redis` and `@homarr/tasks`
- [x] `CI=true pnpm -F @homarr/nextjs build` succeeds
- [x] Bundled output no longer contains `null as unknown` pattern
- [x] `pnpm test` — all tests pass (1 pre-existing Pi-Hole test failure, unrelated)
- [x] Docker build + run: zero startup errors, all cron jobs start cleanly
- [x] `/api/health/live` returns `{"status":"healthy"}` with both database and Redis healthy
- [x] UI renders correctly

### DIFF
--- a/apps/tasks/src/main.spec.ts
+++ b/apps/tasks/src/main.spec.ts
@@ -1,0 +1,108 @@
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+
+import { handshakeAsync } from "@homarr/redis";
+
+import { waitForRedisAsync } from "./main";
+
+const { mockLogger } = vi.hoisted(() => ({
+  mockLogger: {
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+  },
+}));
+
+vi.mock("./overrides", () => ({}));
+
+vi.mock("@homarr/core/infrastructure/logs", () => ({
+  createLogger: () => mockLogger,
+}));
+
+vi.mock("@homarr/redis", () => ({
+  handshakeAsync: vi.fn(),
+}));
+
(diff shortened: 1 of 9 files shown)


Meierschlumpf:
I'm unable to reproduce the issue you have with it not starting up, can you explain when it happens? Because when I run the latest dev image it works without issues

PR #5768, opened 2026-05-22
fix(widgets): show calendar immediately instead of blocking on data fetch

ajnart:
## Idea
It's frustrating that if I have a calendar widget and open my dashboard I get to see a ugly full-size loading spinner within the widget for a few seconds until the data is fetched / API call are done (or cache hits) 

I want to instead show the calendar immediately with a small absolute-positioned loader / spinner / indicator somewhere that shows it's still fetching (or it could be an animation that triggers when it's loaded, like a quick spin of the current date, something cool. No idea.) 

## Summary

- Replace `useSuspenseQuery` with `useQuery` in the calendar widget so the calendar grid renders immediately instead of showing a full-screen spinner while waiting for event data.
- Add a small absolute-positioned `Loader` (16px) in the top-right corner as a subtle loading indicator.
- Event dots populate naturally once the query resolves.

## Note

This same `useSuspenseQuery` anti-pattern exists across ~30 other widget files. This PR addresses the calendar widget only; a follow-up can migrate the remaining widgets.

This is a draft because we might want to apply this to all/most widgets if that's possible. I think that blocking rendering create friction / frustration for the user


ajnart:
Closing in favor of a new PR from `feat/global-refresh-indicator` with improved branch naming and additional features (global query refresh indicator).

PR #5770, opened 2026-05-22
perf(board): server-render board layout and deduplicate fetches

ajnart:
## Summary

- **Split board rendering** into a static CSS Grid (view mode, SSR) and lazy-loaded GridStack (edit mode only), eliminating widget "pop-in" on page load
- **Server-side prefetch** for weather and RSS feed widgets so content renders on first paint
- **Fallback dimensions** for widgets during hydration to prevent layout shifts when `useElementSize` hasn't measured yet

## How it works

1. `_dynamic-client.tsx` no longer wraps `ClientBoard` in `dynamic(..., { ssr: false })` — the board now renders on the server
2. View mode uses a pure CSS Grid (`static-grid.tsx`) with container queries to match GridStack's square-cell sizing
3. Edit mode lazy-loads GridStack via `next/dynamic` with `ssr: false`
4. `useEditMode()` returns `false` during SSR, so the static grid always renders server-side

## Test plan

- [ ] Load a board page — widgets should appear instantly without pop-in
- [ ] Toggle edit mode — GridStack should load and all widgets remain in correct positions
- [ ] Verify category sections collapse/expand correctly in view mode
- [ ] Verify dynamic (container) sections render at correct size without item stretching
- [ ] Test responsive layouts — resize viewport and confirm layout switches work
- [ ] Verify weather and RSS widgets show data on first paint (no loading spinner)


Meierschlumpf:
I see a few problems with these changes, especially regarding the server rendering. If you open the default board right now it loads with smaller texts and 11:45 as time because it uses the server local time and a predefined screen size, therefore we have to decide if we really want these changes or if we are okay with it loading on the client, but at least showing it without layout shift & wrong data

ajnart:
> I see a few problems with these changes, especially regarding the server rendering. If you open the default board right now it loads with smaller texts and 11:45 as time because it uses the server local time and a predefined screen size, therefore we have to decide if we really want these changes or if we are okay with it loading on the client, but at least showing it without layout shift & wrong data

I agree there are some drawbacks, maybe it's more of a proof of concept stage right now. I do really like the idea of having the edit mode as a whole separate thing that loads up gridstack for re-organizing and then just having a simpler read-only dashboard in view mode

Meierschlumpf:
Yeah I also already though about spliting it even from an url standpoint. Similar to confluence where you go into edit mode and it sends you to a different url than the page you can view it

ajnart:
> Yeah I also already though about spliting it even from an url standpoint. Similar to confluence where you go into edit mode and it sends you to a different url than the page you can view it

I wouldn't necessarily create another page, ideally it's possible to leverage dynamic imports like we did for integrations so that it only loads on demand

Meierschlumpf:
I'll convert it back to draft as it is currently a poc

Issue #5805, opened 2026-05-28
feat(auth): OIDC - Support for configurable token_endpoint_auth_method


d-lindahl:
### Describe the feature you'd like to request

Not all OIDC providers support the default `token_endpoint_auth_method=client_secret_basic` and instead require `client_secret_post` for security reasons. 

authjs already supports this and all that's needed is the possibility to configure the method.

PR: https://github.com/homarr-labs/homarr/pull/5806 (https://github.com/homarr-labs/documentation/pull/554 for doc)

PR #5806, opened 2026-05-28
feat(auth): support custom token endpoint auth methods

d-lindahl:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

       ~~did not manage to run pnpm format:fix (Error: --write cannot be used at the same time as --check)~~ Format error fixed
  https://github.com/homarr-labs/documentation/pull/554

Issue: https://github.com/homarr-labs/homarr/issues/5805

I have verified in my installation (building the docker container) against Stalwart (https://github.com/stalwartlabs/stalwart). Without my change the auth flow fails after authenticating at the OIDC side and getting redirected back to Homarr with the following error:
```
Read more at https://errors.authjs.dev#callbackrouteerror module="authConfiguration" name="s" type="CallbackRouteError" kind="error"
    at r3 (/app/apps/nextjs/.next/server/chunks/_0wichgl._.js:404:20321)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async is (/app/apps/nextjs/.next/server/chunks/_0wichgl._.js:404:27797)
    at async io (/app/apps/nextjs/.next/server/chunks/_0wichgl._.js:404:32788)
    at async u (/app/apps/nextjs/.next/server/chunks/[root-of-the-server]__0rddkaw._.js:1:17594)
caused by {"err":{"cause":{"error":"invalid_client"},"code":"OAUTH_RESPONSE_BODY_ERROR","error":"invalid_client","status":400,"name":"et"},"error":"invalid_client","provider":"oidc"}
```
With my change and setting `AUTH_OIDC_TOKEN_ENDPOINT_AUTH_METHOD=client_secret_post` I can complete the auth flow. I have not verified the old behavior more than seeing that I still get the old error when I remove the new `AUTH_OIDC_TOKEN_ENDPOINT_AUTH_METHOD` env variable.

Not 100% sure if the "none" option should be added as a valid option but it is one of the supported values in authjs.


ajnart:
Thanks for the PR ! Looks okay to me, but I am not an expert in the matter.
Can you rectify the format ? 
I think @Meierschlumpf will be able to review the PR more in depth to make sure there are no regressions
Hopefully we can get this shipped before tomorrow’s release 🤞🏼

d-lindahl:
> Thanks for the PR ! Looks okay to me, but I am not an expert in the matter. Can you rectify the format ?

I'll see if I can figure it out!

Meierschlumpf:
Based on the switch statement from https://github.com/nextauthjs/next-auth/blob/dab3cfb1a04aad0b20e616bf9f1d21b5d6e318cd/packages/core/src/lib/actions/callback/oauth/callback.ts#L95-L126 I would suggest that we add `client_secret_jwt` as well. For `private_key_jwt` we currently don't have enough configuration options and would probably need to extend support for oauth2 (as it is using `token.clientPrivateKey`)

As the current value is undefined and it also is the current path for `client_secret_basic`, this is the correct default value.

packages/auth/env.ts
```diff
@@ -40,6 +40,9 @@ export const env = createEnv({
           AUTH_OIDC_NAME_ATTRIBUTE_OVERWRITE: z.string().optional(),
           AUTH_OIDC_FORCE_USERINFO: createBooleanSchema(false),
           AUTH_OIDC_ENABLE_DANGEROUS_ACCOUNT_LINKING: createBooleanSchema(false),
+          AUTH_OIDC_TOKEN_ENDPOINT_AUTH_METHOD: z
+            .enum(["client_secret_basic", "client_secret_post", "none"])
```
Meierschlumpf:
```suggestion
            .enum(["client_secret_basic", "client_secret_post", "client_secret_jwt", "none"])
```

Meierschlumpf (changes requested):

packages/auth/providers/oidc/oidc-provider.ts
```diff
@@ -14,6 +14,9 @@ export const OidcProvider = (headers: ReadonlyHeaders | null): OIDCConfig<Profil
   type: "oidc",
   clientId: env.AUTH_OIDC_CLIENT_ID,
   clientSecret: env.AUTH_OIDC_CLIENT_SECRET,
+  client: {
+    token_endpoint_auth_method: env.AUTH_OIDC_TOKEN_ENDPOINT_AUTH_METHOD,
+  },
```
d-lindahl, line 19: @Meierschlumpf  After reading https://github.com/nextauthjs/next-auth/blob/dab3cfb1a04aad0b20e616bf9f1d21b5d6e318cd/packages/core/src/lib/actions/callback/oauth/callback.ts#L96 "TODO: in the next breaking major version have undefined be `client_secret_post`" maybe better to not add this at all if the env does not exist? That was my first draft but I didn't like the ugliness of `client: undefined` if `env.AUTH_OIDC_TOKEN_ENDPOINT_AUTH_METHOD` is undefined (and let AUTH_OIDC_TOKEN_ENDPOINT_AUTH_METHOD be optional)


```suggestion
  client: env.AUTH_OIDC_TOKEN_ENDPOINT_AUTH_METHOD
    ? {
        token_endpoint_auth_method: env.AUTH_OIDC_TOKEN_ENDPOINT_AUTH_METHOD,
       }
    :  undefined,
```

d-lindahl: Though after reading this myself - just removing the default would mean `token_endpoint_auth_method: undefined` would  give the same result.

Meierschlumpf: The problem with removing the default is, that if the underlying library would change their default, auth for some users might no longer work. Therefore I suggest we configure `client_secret_basic` as default value

d-lindahl: Gotcha! I'll put back the default!

d-lindahl:
Changed according to discussion. Format check is failing for a different module, passing for @homarr/auth.

Meierschlumpf (approved):

Meierschlumpf:
Format issue will be fixed in #5815

d-lindahl:
Thanks! #5815 was exactly what I was running into locally. Appreciate the quick work! 🙇

Meierschlumpf:
Thanks for your contribution

PR #5807, opened 2026-05-28
fix: prevent GitHub releases check from blocking app on API failure

### LINKED ISSUE #5786: bug: Occasionally unable to open Homarr — GitHub API request quota exhausted even after providing token
### Describe the bug

Title:
Occasionally unable to open Homarr — GitHub API request quota exhausted even after providing token

Body:
Sometimes when I try to open Homarr, it fails to load. The backend logs show the following error repeated multiple times:

text
homarr  | Request quota exhausted for request GET /repos/{owner}/{repo}/releases
homarr  | Request quota exhausted for request GET /repos/{owner}/{repo}/releases
homarr  | Request quota exhausted for request GET /repos/{owner}/{repo}/releases
homarr  | Request quota exhausted for request GET /repos/{owner}/{repo}/releases
I have already configured my GitHub API token in Homarr, but the issue still persists. The problem occurs intermittently — Homarr works most of the time, but occasionally becomes inaccessible due to this quota exhaustion.

Could you please advise on how to resolve this? Is there any additional configuration needed to ensure the token is used correctly for all API requests?

Thank you.

Image

### Steps to reproduce

Image

### Impact

This problem severely affects usage — Homarr cannot be opened at all.

### Additional information

Image

### Version

1.62.0

### Installation method

Docker Compose

### Browser

Edge (Proprietary)

ajnart:
## Summary

- Add optional `fallbackToStaleOnError` to the cached request handler so failed upstream fetches can return last-known-good Redis data instead of throwing
- Harden the update checker against GitHub API failures (401, quota exhausted, unreachable): disable Octokit throttling for fast failure, extend cache TTL to 24h, and enable stale fallback
- Reduce GitHub API pressure by running the update checker cron daily and respecting cache TTL instead of force-refreshing every hour

Closes #5786

## Context

Users reported Homarr becoming completely inaccessible when the GitHub releases endpoint returns quota errors — especially visible in regions where GitHub is unreliable. The update checker was hitting `GET /repos/homarr-labs/homarr/releases` too often and Octokit would retry throttled requests, blocking admin page loads.

## Test plan

- [x] `NODE_ENV=development pnpm exec vitest run packages/request-handler/src/test/update-checker.spec.ts`
- [ ] Verify admin dashboard loads when GitHub API is unreachable (no update indicator shown, no page hang)
- [ ] Verify update indicator still appears when GitHub API is reachable and a newer release exists
- [ ] Verify stale cached update data is served when GitHub fails after a prior successful fetch

### DIFF
--- a/packages/request-handler/src/lib/cached-request-handler.ts
+++ b/packages/request-handler/src/lib/cached-request-handler.ts
@@ -40,10 +40,16 @@ export const createCachedRequestHandler = <TData, TInput extends Record<string, 
               };
             } catch (error) {
               if (options.fallbackToStaleOnError && channelData) {
-                logger.warn(new ErrorWithMetadata("Cached request handler using stale cache after fetch failure", {
-                  channel: channel.name,
-                  queryKey: options.queryKey,
-                }, { cause: error }));
+                logger.warn(
+                  new ErrorWithMetadata(
+                    "Cached request handler using stale cache after fetch failure",
+                    {
+                      channel: channel.name,
+                      queryKey: options.queryKey,
+                    },
+                    { cause: error },
+                  ),
+                );
                 return channelData;
               }
 
--- a/packages/request-handler/src/lib/cached-request-handler.ts
+++ b/packages/request-handler/src/lib/cached-request-handler.ts
@@ -15,6 +15,7 @@ interface Options<TData, TInput extends Record<string, unknown>> {
     options: Options<TData, TInput>,
   ) => ReturnType<typeof createChannelWithLatestAndEvents<TData>>;
   cacheDuration: Duration;
+  fallbackToStaleOnError?: boolean;
 }
 
 export const createCachedRequestHandler = <TData, TInput extends Record<string, unknown>>(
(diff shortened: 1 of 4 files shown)


Meierschlumpf:
This one seems of (119 files for a small bugfix should not be necessary). It seems like you have unrelated changes in this pr

ajnart:
> This one seems of (119 files for a small bugfix should not be necessary). It seems like you have unrelated changes in this pr

I'm very confused how it got this bad, I'll fix it asap

packages/request-handler/src/lib/cached-request-handler.ts
```diff
@@ -26,13 +27,39 @@ export const createCachedRequestHandler = <TData, TInput extends Record<string,
...
+              const staleFallbackHandlers: Record<string, () => { data: TData; timestamp: Date } | null | undefined> = {
+                enabled: () => channelData,
+                disabled: () => null,
+              };
+              const staleFallbackKeyMap: Record<string, string> = {
+                true: "enabled",
+                false: "disabled",
+              };
+              const staleFallbackKey = staleFallbackKeyMap[String(Boolean(options.fallbackToStaleOnError))];
+              const staleData = staleFallbackHandlers[staleFallbackKey]();
```
Meierschlumpf: Is there a reason we need to make it this complicated? Can't we just check for it and if enabled return the stale data?

Meierschlumpf (dismissed):

packages/request-handler/src/lib/cached-request-handler.ts
ajnart: you're right ill just push the simplified flow

packages/request-handler/src/lib/cached-request-handler.ts
```diff
@@ -26,13 +27,28 @@ export const createCachedRequestHandler = <TData, TInput extends Record<string,
...
+                data,
+                timestamp: new Date(),
+              };
+            } catch (error) {
+              if (options.fallbackToStaleOnError && channelData) {
+                logger.warn("Cached request handler using stale cache after fetch failure", {
+                  channel: channel.name,
+                  queryKey: options.queryKey,
+                  cause: error instanceof Error ? error.message : String(error),
+                });
```
Meierschlumpf:
```suggestion
                logger.warn(new ErrorWithMetadata("Cached request handler using stale cache after fetch failure", {
                  channel: channel.name,
                  queryKey: options.queryKey,
                }, { cause: error });
```

Meierschlumpf: Don't forget the import

Meierschlumpf (changes requested):
After this change it's good to go!

Meierschlumpf (dismissed):

PR #5809, opened 2026-05-28
feat(widgets): add in-place app editing in widget edit modal

ajnart:
<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

## Summary

Adds the ability to edit app properties (name, icon, URL, description, ping URL) directly from the dashboard's "Edit item - App" modal, removing the need to navigate to `/manage/apps/edit/[id]`.

- New `EmbeddedAppEditForm` component that fetches the app by ID and renders the shared `AppForm` in headless mode (no buttons, no `<form>` wrapper)
- Calls `clientApi.app.update` on save, gated by `form.isDirty()` so untouched apps are never mutated
- Permission-gated: the "Edit app" tab only appears for users with `app-modify-all`
- Informational alert explains that changes propagate across all dashboards

## Screenshots

| Edit widget tab | Edit app tab |
|---|---|
| CleanShot 2026-05-28 at 14 48 51 |  CleanShot 2026-05-28 at 14 48 56  |

## Changes

- `AppForm`: added `hideButtons` and `formRef` props for external submission control
- `WidgetEditModal`: Mantine Tabs wrapping "Edit widget" / "Edit app" panels
- `EmbeddedAppEditForm`: new component for in-place app editing
- `BoardItemMenu`: passes `appId` from widget options to the modal
- Translation keys for tab labels and propagation notice


packages/widgets/src/modals/widget-edit-modal.tsx
```diff
@@ -77,78 +82,111 @@ export const WidgetEditModal = createModal<ModalProps<WidgetKind>>(({ actions, i
...
+          label={t("item.edit.field.integrations.label")}
+          data={innerProps.integrationData}
+          {...form.getInputProps("integrationIds")}
+        />
+      )}
+      {Object.entries(options).map(([key, value]) => {
+        const Input = getInputForType(value.type);
+
+        if (
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
manuel-rw: Is this really needed?

manuel-rw (dismissed):

packages/widgets/src/modals/widget-edit-modal.tsx
ajnart: You're right it's not, I put it there before pulling dev that switched to oxlint, I'll get rid of it

apps/nextjs/src/components/board/items/item-menu.tsx
```diff
@@ -85,6 +85,7 @@ const BoardItemMenuInner = ({ offset, item, resetErrorBoundary }: BoardItemMenuP
         ),
         integrationSupport: "supportedIntegrations" in currentDefinition,
         settings,
+        appId: item.kind === "app" ? (item.options.appId as string | undefined) : undefined,
```
Meierschlumpf, line 88: How do we deal with bookmarks?

packages/widgets/src/modals/embedded-app-edit-form.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+interface EmbeddedAppEditFormProps {
+  appId: string;
+  handleRef: React.Ref<EmbeddedAppEditFormHandle>;
+}
+
+export const EmbeddedAppEditForm = ({ appId, handleRef }: EmbeddedAppEditFormProps) => {
+  const t = useI18n();
+  const appFormRef = useRef<AppFormHandle>(null);
+  const { data: app, isPending: isLoadingApp } = clientApi.app.byId.useQuery({ id: appId });
```
Meierschlumpf: What happens if the app does no longer exist?

Meierschlumpf (dismissed):
Not sure how you guys feel about this, but IMO it's bad UX that the app is saved when we close the modal (see also issue from copilot regarding validation etc.)

IMO it would be best to have an edit button within the app input and open a second modal to edit the app, similar to the existing create modal

packages/widgets/src/modals/embedded-app-edit-form.tsx
ajnart: I haven't considered this edge case, in what way do you think we can edit an app that doesn't exist anymore ?

apps/nextjs/src/components/board/items/item-menu.tsx
ajnart: I haven't thought about this, I guess we don't for now ?

ajnart:
> Not sure how you guys feel about this, but IMO it's bad UX that the app is saved when we close the modal (see also issue from copilot regarding validation etc.)
> 
> IMO it would be best to have an edit button within the app input and open a second modal to edit the app, similar to the existing create modal

I believe comment is more about error handling, as soon as the form is dirty then the save will trigger when pressing the save button and directly close the modal, so it's bypassing the form validation for the app change. The user could be in a not-authorized app modification state, press save expecting this change to propagate but saving would error silently. 

I wouldn't want to save any changes without the user pressing save manually either

PR #5811, opened 2026-05-28
feat(ui): add global query cache leveraging Redis 

ajnart:
## Summary

- Adds a global **query refresh indicator** to the board header (left of the edit button) that monitors all widget-related TanStack Query cache entries (`widget.*`, `docker.*`, `app.*`, `integration.*`)
- Indicator shows an **orange pulsing dot** while any widget query is fetching, **green flash** on completion, and **red** if any query is in error state
- Hover card displays per-query status rows with colored dots, labels, and relative timestamps (`fromNow`)
- Removes the per-widget loading spinner from the **calendar widget**, which now relies on the global indicator instead of a local `<Loader>`

Continues the work from #5768 with a broader scope.

## Test plan

- [ ] Verify the indicator appears in the board header when widget queries are active
- [ ] Confirm orange dot pulses during fetch, flashes green on completion
- [ ] Hover over the indicator to see per-query status, staleness, and error counts
- [ ] Verify beep plays on fetch-to-idle transition (requires prior user interaction for AudioContext)
- [ ] Check that calendar widget renders immediately without a local spinner
- [ ] Test with Docker, bookmarks, and app widgets to confirm they appear in the indicator


ajnart:
i believe this change will be nice, I'd much rather see the calendar and then 500-800ms later see the entries popping up rather than the loading circle, that's probably the case for a bunch of other integrations that could have fallback states in place of the loading spinner. The issue is I'm not sure where to indicate the freshness of the queries

PR #5861, opened 2026-06-03
fix(integrations): support truenas json-rpc api, certificate trust, and api key auth

### LINKED ISSUE #4206: bug: TrueNAS Integration doesn't work with 25.10
### Describe the bug

I was trying to integrate TrueNAS 25.10-RC1 but the documentation isn't correct anymore since it asks to make a user belonging to the 'auxiliary_administrator' group which has been removed since. I tried creating other kinds of users but I can't get a connection either way. I believe the UI backend has somewhat changed with different API calls. 

### Steps to reproduce

Following documentation steps from: https://homarr.dev/docs/integrations/truenas/

### Impact

Just an important part of the integration since Homarr is running on TrueNAS and it's important to be able to monitor my TrueNAS systems from Homarr. 

### Additional information

_No response_

### Version

1.39.0

### Installation method

TrueNAS Apps

### Browser

Chrome

afonsojramos:
## Summary

The TrueNAS integration could not connect to current TrueNAS installs. Two root causes:

1. **Self-signed certificates were never trusted.** The integration opened a raw `WebSocket` that performed default TLS verification with no access to Homarr's trusted certificates, so the handshake failed against the default TrueNAS self-signed certificate (close `1006`). The underlying `WebSocket` also swallowed the TLS error, so the existing "trust this certificate" prompt never fired.
2. **Only the legacy DDP `/websocket` protocol was supported.** TrueNAS 24.10+ moved to a JSON-RPC 2.0 API (`/api/current`) and is removing the legacy protocol, so newer versions could not be reached.

## Changes

- Connect with the `ws` client so the socket carries Homarr's trusted CA and custom `checkServerIdentity`, exactly like the HTTP integrations. TLS failures now map to a certificate request error, so the existing "trust certificate" flow works for TrueNAS.
- Speak both the JSON-RPC 2.0 API (`/api/current`, TrueNAS 24.10+) and the legacy `/websocket` API, auto-detecting per server (JSON-RPC first, legacy fallback). Certificate, connection, DNS, and timeout errors are endpoint-independent and are never retried against the other API.
- Add API key authentication (`auth.login_with_api_key`) as an alternative to username/password — the method recommended for recent TrueNAS versions and one that does not depend on the removed `auxiliary_administrator` group.
- Harden the connection lifecycle: in-flight connection de-duplication so concurrent callers share one socket, eviction on close, sockets cached only after successful authentication, and per-request close/error handling so a dropped socket fails fast instead of waiting out the timeout.

## Testing

- Unit tests covering protocol envelope/response matching (JSON-RPC and legacy), auto-detection and fallback, auth-method selection, and certificate-error mapping.
- Verified live against a real **TrueNAS 25.04** server with **both username/password and an API key** - each authenticates over auto-detected JSON-RPC 2.0 with the self-signed certificate trusted, and returns valid pool, CPU, memory, and network data.
- `tsc --noEmit`, `oxlint`, and `oxfmt` are clean on the touched packages.

> Verified on 25.04; 25.10 was not available to test, but JSON-RPC is the API 25.10 requires. The integration documentation was also updated to recommend API key authentication in https://github.com/homarr-labs/documentation/pull/559

---

Closes #4206

### DIFF
--- a/packages/integrations/src/truenas/truenas-integration.ts
+++ b/packages/integrations/src/truenas/truenas-integration.ts
@@ -309,7 +309,7 @@ export class TrueNasIntegration extends Integration implements ISystemHealthMoni
 
     return {
       cpuUtilization: cpuData.reduce((acc, item) => acc + (item > 100 ? 0 : item), 0) / cpuData.length,
-      cpuTemp: Math.max(...cpuTempData.filter((_item, i) => i > 0)),
+      cpuTemp: Math.max(...cpuTempData.filter((_item, index) => index > 0)),
       memAvailableInBytes: systemInformation.physmem,
       memUsedInBytes: memoryData[1] ?? 0, // Index 0 is UNIX timestamp, Index 1 is free space in bytes
       fileSystem: datasets.map((dataset) => ({
--- a/packages/integrations/src/truenas/truenas-integration.ts
+++ b/packages/integrations/src/truenas/truenas-integration.ts
@@ -1,15 +1,6 @@
 import dayjs from "dayjs";
-import type { RawData } from "ws";
-import { WebSocket } from "ws";
 import z from "zod";
 
-import { createId } from "@homarr/common";
-import { matchErrorCode, RequestError, ResponseError } from "@homarr/common/server";
-import {
-  getAllTrustedCertificatesAsync,
-  getTrustedCertificateHostnamesAsync,
-} from "@homarr/core/infrastructure/certificates";
-import { createCustomCheckServerIdentity } from "@homarr/core/infrastructure/http";
 import { createLogger } from "@homarr/core/infrastructure/logs";
 
 import type { IntegrationTestingInput } from "../base/integration";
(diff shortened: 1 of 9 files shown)


manuel-rw:
Thank you very much for working on this. It'll review it as soon as possible

packages/integrations/src/truenas/truenas-integration.ts
```diff
@@ -1,121 +1,168 @@
...
+  // the in-flight connection promise so concurrent callers reuse a single socket instead of racing.
+  private static readonly connectionMap = new Map<string, Promise<TrueNasSocket>>();
+  private static readonly apiMap = new Map<string, TrueNasApi>();
 
-  private wsUrl() {
-    const url = super.url("/websocket");
+  private wsUrl(path: `/${string}`) {
+    const url = super.url(path);
     url.protocol = url.protocol.replace("http", "ws");
     return url;
```
manuel-rw: Not your change, but maybe we should add this to the base?

packages/integrations/src/truenas/truenas-integration.ts
```diff
@@ -301,46 +338,112 @@ export class TrueNasIntegration extends Integration implements ISystemHealthMoni
...
   }
 
+  private extractNetworkTrafficData = (data: z.infer<typeof reportingNetDataSchema>, index: 1 | 2) => {
+    return data.reduce((acc, current) => acc + (current.data.at(-1)?.at(index) ?? 0), 0);
+  };
+
+  private extractLatestReportingData(data: ReportingItem[], key: ReportingItem["identifier"]) {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const dataObject = data.find((item) => item.identifier === key)!;
+    // TODO: check why the below sorting is done, because right now it compares number[] with number[]?
```
manuel-rw: TODO?

packages/integrations/src/truenas/truenas-integration.ts
manuel-rw, line 1: I know that this is not a problem introduced in your PR, but this class is getting increasingly long.
It does multiple things. Could you try splitting it?

manuel-rw (changes requested):

packages/integrations/src/truenas/truenas-integration.ts
afonsojramos: Done! Extracted a TrueNasClient (`truenas-client.ts`) that handles the websocket transport, connection lifecycle, protocol detection, and auth. TrueNasIntegration now only fetches and maps the health data.

packages/integrations/src/truenas/truenas-integration.ts
afonsojramos: Whoops 🙊 Replaced it with data.at(-1) (reporting rows are time-ascending). Validated that the reported values are unchanged against my 25.04 server 👌

afonsojramos:
@manuel-rw applied all changes!

manuel-rw (approved):

PR #5862, opened 2026-06-03
feat: add PeaNUT integration and UPS monitoring widget

afonsojramos:
## Summary

- add a new `peaNut` integration in the `ups` category, mapping PeaNUT NUT variables into `UpsSummary`
- add mock UPS summary data so the feature can be exercised locally without a live PeaNUT server
- add cached request-handler and widget API routes for UPS summaries and subscriptions
- add a new UPS widget with battery, load, voltage, runtime, and status displays
- add adaptive widget layouts (full, compact, and mini) so it stays useful from a full panel down to a 1x1 tile

## Verification

- `pnpm build`
- `turbo typecheck`
- `oxlint` on the changed files
- unit tests for PeaNUT mapping and Basic Auth handling
- local dev verification with the mock UPS integration and widget; and against a real PeaNUT server

> image
On the left is the mocked UPS and on the right my actual UPS

### Updated: Now with Card:
image

- [x] Documentation is up to date. PR available here: https://github.com/homarr-labs/documentation/pull/561


packages/definitions/src/integration.ts
```diff
@@ -378,6 +378,13 @@ export const integrationDefs = {
     defaultUrl: "https://api.umami.is/v1",
     documentationUrl: createDocumentationLink("/docs/integrations/umami"),
   },
+  peaNut: {
```
manuel-rw, line 381: Is this compatible with https://networkupstools.org/index.html ?
Or do I specifically need to setup PeaNUT?

packages/integrations/src/interfaces/ups-summary/ups-summary-types.ts
```diff
@@ -0,0 +1,24 @@
+export interface UpsSummary {
+  /** Stable identifier of the UPS device on the NUT server */
+  id: string;
+  /** Human readable name, derived from the manufacturer and model */
+  name: string;
+  manufacturer: string | null;
+  model: string | null;
+  serial: string | null;
+  /** Raw NUT `ups.status` value, a space separated set of flags (e.g. "OL CHRG") */
+  status: string;
```
manuel-rw: Would it be possible to set this to a literal / enum instead and parse the PeaNUT status?
This makes it easier to add new integrations and allows you to combine states into one.

packages/request-handler/src/ups.ts
```diff
@@ -0,0 +1,20 @@
...
+export const upsSummariesRequestHandler = createCachedIntegrationRequestHandler<
+  UpsSummary[],
+  IntegrationKindByCategory<"ups">,
+  Record<string, never>
+>({
+  async requestAsync(integration, _input) {
+    const integrationInstance = await createIntegrationAsync(integration);
+    return await integrationInstance.getUpsSummariesAsync();
+  },
+  cacheDuration: dayjs.duration(30, "seconds"),
```
manuel-rw: This seems excessive, maybe reduce to 1 minutes?

packages/widgets/src/ups/component.tsx
```diff
@@ -0,0 +1,204 @@
...
+export default function UpsWidget({ options, integrationIds, isEditMode }: WidgetComponentProps<"ups">) {
+  const t = useScopedI18n("widget.ups");
+
+  if (integrationIds.length === 0) {
+    return (
+      <Stack align="center" justify="center" h="100%">
+        <Text c="dimmed">{t("error.noIntegration")}</Text>
+      </Stack>
+    );
+  }
```
manuel-rw, line 28: Please check the other widgets, the new and preferred way is throwing an exception.

packages/widgets/src/ups/component.tsx
```diff
@@ -0,0 +1,204 @@
...
+    instance.summaries.map((summary) => ({ key: `${instance.integrationId}:${summary.id}`, summary })),
+  );
+
+  if (devices.length === 0) {
+    return (
+      <Stack align="center" justify="center" h="100%">
+        <Text c="dimmed">{t("noData")}</Text>
+      </Stack>
+    );
+  }
```
manuel-rw, line 68: Also consider throwing?

packages/widgets/src/ups/component.tsx
```diff
@@ -0,0 +1,204 @@
...
+  summary: UpsSummary;
+  options: WidgetComponentProps<"ups">["options"];
+  t: ScopedTranslationFunction<"widget.ups">;
+}
+
+function UpsDeviceCard({ summary, options, t }: UpsDeviceCardProps) {
+  const status = getStatusDisplay(summary.status, t);
+
+  return (
+    <Box p="xs" style={{ borderRadius: "var(--mantine-radius-md)", backgroundColor: "var(--mantine-color-default)" }}>
```
manuel-rw: Maybe a Card would be more appropriate here?

packages/widgets/src/ups/component.tsx
```diff
@@ -0,0 +1,204 @@
...
+        {label}
+      </Text>
+      <Text size="xs" fw={500}>
+        {value}
+      </Text>
+    </Group>
+  );
+}
+
+function getStatusDisplay(
```
manuel-rw: This can be removed / simplified once you address my comment about status resolution / parsing

packages/widgets/src/ups/component.tsx
```diff
@@ -0,0 +1,204 @@
...
+function formatRuntime(seconds: number): string {
+  if (seconds <= 0) return "0m";
+
+  const hours = Math.floor(seconds / 3600);
+  const minutes = Math.floor((seconds % 3600) / 60);
+
+  if (hours > 0) return `${hours}h ${minutes}m`;
+  if (minutes > 0) return `${minutes}m`;
+  return `${seconds}s`;
+}
```
manuel-rw: 1. Either we already have a function for this
2. or you can use dayjs for this.

manuel-rw (changes requested):

packages/widgets/src/ups/component.tsx
afonsojramos: Updated!

packages/widgets/src/ups/component.tsx
afonsojramos: Now using `formatDuration` 👍

packages/integrations/src/interfaces/ups-summary/ups-summary-types.ts
afonsojramos: Nice suggestion! `UpsSummary.status` is now a typed union and the raw NUT flags are parsed into it inside the integration (`parseStatus`), so the widget just maps `status`→`colour`/`label`

packages/definitions/src/integration.ts
afonsojramos: Hmm... This integration talks to PeaNUT’s HTTP API (GET `/api/v1/devices`), so you need PeaNUT (or another service exposing that API) in front of NUT; it doesn’t speak the raw NUT protocol on TCP 3493, so a bare `nut-server` won’t work on its own.

I kept the ups category and IUpsSummaryIntegration interface provider-agnostic, so a native NUT integration could be added later if there’s demand. I’d leave that as a possible follow-up rather than expanding this PR, since it’d be the first integration speaking a raw TCP protocol rather than HTTP.

ajnart:
Thanks for the PR ! 
The UI quite simplistic but it does not need to be fancy. Could you make sure you had some kind of adaptive versions for smaller widget sizes ? Like 1x1 with the info still shown, and make sure that for example if it's 1x3 and you have two entries then there's a scroll bar (which should be handled by our widget container's default behavior I believe) 
Sidenote is that in order to merge a PR for a new integration / widget it needs to be accompanied by a PR to the docs, documenting that feature ;) 

@manuel-rw I was thinking, we should introduce/encourage some kind of CODEOWNERS requirements so that contributors can be mentioned / flagged directly when someone is modifying a feature they introduced

afonsojramos:
> Sidenote is that in order to merge a PR for a new integration / widget it needs to be accompanied by a PR to the docs, documenting that feature ;)

The PR is already open here: https://github.com/homarr-labs/documentation/pull/561

afonsojramos:
> The UI quite simplistic but it does not need to be fancy. Could you make sure you had some kind of adaptive versions for smaller widget sizes ? Like 1x1 with the info still shown, and make sure that for example if it's 1x3 and you have two entries then there's a scroll bar (which should be handled by our widget container's default behavior I believe)

Sure, maybe a follow up PR though?

manuel-rw:
> > The UI quite simplistic but it does not need to be fancy. Could you make sure you had some kind of adaptive versions for smaller widget sizes ? Like 1x1 with the info still shown, and make sure that for example if it's 1x3 and you have two entries then there's a scroll bar (which should be handled by our widget container's default behavior I believe)
> 
> Sure, maybe a follow up PR though?

I think this should be part of this PR, since otherwise the widget may not be us useful for small sizes.

afonsojramos:
Alright, will update. Regarding CODEOWNERS, I think it is a good thing, but if I am not mistaken, users needs to have at least write access to the repo to be able to be included there, no?

ajnart:
> Alright, will update. Regarding CODEOWNERS, I think it is a good thing, but if I am not mistaken, users needs to have at least write access to the repo to be able to be included there, no?

You're right I just read the docs about it. It makes sense because otherwise your PR's approval / change request wouldn't have much weight. I think we'll continue not using it for now

afonsojramos:
> You're right I just read the docs about it. It makes sense because otherwise your PR's approval / change request wouldn't have much weight. I think we'll continue not using it for now

@ajnart to be honest, not sure. Obviously, I'm speaking for myself, but I think it should be acceptable for anyone that contributes as much as an integration to be added as a contributor. Contributors still have to go through the PR process, so the only change is not having to fork and having the benefit of having approvals count to something, ie, there is no "danger" in doing this afaik.

ajnart (approved):

PR #5892, opened 2026-06-07
feat(lang): add Finnish language support

ajnart:
## Summary

Adds Finnish (fi) as a supported language in Homarr.

### Changes
- Added fi locale configuration to config.ts with native name Suomi, flag icon fi, MRT localization MRT_Localization_FI, and Day.js locale fi
- Created fi.json translation file based on en.json (translations to be populated via Crowdin)

### Notes
- Both dayjs and mantine-react-table have built-in Finnish locale support
- Follows the same pattern as previous language additions (de-CH, en-gb)


packages/translation/src/lang/fi.json
Meierschlumpf, line 1: Not sure if it overrides these values or just adds missing / modified values. But let's just try it out

Meierschlumpf (dismissed):

manuel-rw (dismissed):

PR #5982, opened 2026-06-17
fix(definitions): resolve doc-slug gaps breaking typecheck

mairas:
## What

When rebasing homarr-labs/homarr#5595 and hatlabs/homarr#4, I noticed that the
`gotify` integration, and the `archiveTeamWarrior` integration and widget, were
added without matching entries in the doc-slug maps. This left the
`Record<IntegrationKind, …>` / `Record<WidgetKind, …>` maps incomplete and made
`gotify` reference a `/docs/integrations/gotify` page that exists neither in the
repo nor the published sitemap — so it never entered the generated
`HomarrDocumentationPath` type.

Result: `@homarr/definitions` fails `typecheck`, which cascades to every
dependent package.

```
src/docs/integration-doc-slugs.ts(3,14): error TS2739: ... is missing the following properties ...: gotify, archiveTeamWarrior
src/docs/widget-doc-slugs.ts(3,14): error TS2741: Property 'archiveTeamWarrior' is missing
src/integration.ts(363,47): error TS2345: Argument of type '"/docs/integrations/gotify"' is not assignable to parameter of type 'HomarrDocumentationPath'
```

`dev`'s `[Quality] Code Analysis` run is currently red for exactly this reason.

## How

`gotify` has no documentation page, so its `documentationUrl` is set to `null`
and both new keys are mapped to `null`, matching `archiveTeamWarrior`'s existing
treatment. Since both slugs are `null`, no path enters the generated sitemap and
`homarr-docs-sitemap.ts` is unchanged.

- `integration.ts` — gotify `documentationUrl` → `null`
- `integration-doc-slugs.ts` — add `gotify: null`, `archiveTeamWarrior: null`
- `widget-doc-slugs.ts` — add `archiveTeamWarrior: null`

If a gotify docs page is planned, the follow-up is to add the page and flip its
slug to `"gotify"` (restoring the `documentationUrl`).

## Verification

- `pnpm build` — passes
- `pnpm typecheck` — 37/37 packages pass (previously halted at `@homarr/definitions`)
- `pnpm lint` / `pnpm format` for `@homarr/definitions` — pass

---


ajnart:
You're right, I have a PR opened that would fix these, I'm porting over the docs from the old repo, so the values would be non-null for these integrations.

PR #6016, opened 2026-06-20
fix(widgets): correct TrueNAS memory/pool display and certificate trust button

afonsojramos:
## Summary

Fixes several display bugs in the health monitoring widget when used with TrueNAS (and a related certificate-flow bug), found while setting up a TrueNAS integration.

### TrueNAS memory usage was inverted
The `reporting.get_data` `"memory"` graph reports **available (free)** memory (legend is `["time", "available"]`), but the code assigned that free value to `memUsedInBytes` and the physical total to `memAvailableInBytes`. Memory therefore showed ~10% used on a box that was actually ~88% used (ZFS ARC). Now `memAvailableInBytes` = free and `memUsedInBytes` = `physmem - free`.

### Pool "Available" showed total size, not free space
The integration set `fileSystem.available` to `dataset.size` (the pool's total size) rather than `dataset.free`, so the widget's "Available" label/value displayed the total. Fixed to use `dataset.free`. Unraid had the same latent bug (`available` = total disk size) and is fixed to report free space too.

### Health monitoring widget polish
- **Hide temperature for pools/disks without a reading.** Pools have no SMART temperature, but the widget defaulted `null` to `0` and rendered "0°C". Temperature now stays `null` and the element is only shown when a real reading exists.
- **Show used/available sizes as readable text.** Sizes were hidden in hover tooltips, only "available" was formatted (the "used" tooltip showed raw bytes). They are now formatted consistently below the bar via a helper that handles both raw-byte and pre-formatted integrations.
- **Fix the cramped progress-bar labels** that truncated to "Us…"/"Avail…" by moving them out of the bar segments.
- **Remove the "Last status update: in a few seconds" line**, which was noise given the 5s subscription refresh.

### Certificate "Trust" button closed the modal
On the integration form, the "Trust"/"Upload" certificate buttons had no `type`, so inside the `<form>` they defaulted to `type="submit"` and submitted the form on click (tearing down the confirm modal). Added `type="button"` (the "Retry" button intentionally stays `type="submit"`).

## Verification
- Verified the corrected memory and pool values against a live TrueNAS server over the JSON-RPC API.
- TrueNAS integration tests updated and passing.
- `oxfmt` and `oxlint` clean on the changed files.

## Checklist
- [x] Commits follow the conventional commits guideline
- [x] No shorthand variable names are used
- [x] `oxfmt`/`oxlint` pass on changed files; tests pass

### DIFF
--- a/apps/nextjs/src/app/[locale]/manage/integrations/_components/test-connection/test-connection-certificate.tsx
+++ b/apps/nextjs/src/app/[locale]/manage/integrations/_components/test-connection/test-connection-certificate.tsx
@@ -131,6 +131,7 @@ export const CertificateErrorDetails = ({ error, url }: CertificateErrorDetailsP
       {(error.data.reason === "untrusted" && rootCertificate.isSelfSigned) ||
       error.data.reason === "hostnameMismatch" ? (
         <Button
+          type="button"
           variant="default"
           fullWidth
           onClick={error.data.reason === "hostnameMismatch" ? handleTrustHostname : handleTrustSelfSigned}
@@ -140,6 +141,7 @@ export const CertificateErrorDetails = ({ error, url }: CertificateErrorDetailsP
       ) : null}
       {error.data.reason === "untrusted" && !rootCertificate.isSelfSigned ? (
         <Button
+          type="button"
           variant="default"
           fullWidth
           onClick={() =>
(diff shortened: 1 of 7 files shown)


afonsojramos:
Split into focused, independently reviewable PRs:
- #6017 fix(truenas): report actual memory usage and pool free space
- #6018 fix(unraid): report disk free space instead of total size
- #6019 fix(widgets): clean up health monitoring disk and size display
- #6020 fix(integrations): prevent certificate trust button from submitting the form
- #6021 fix(widgets): hide health monitoring default tab option for non-proxmox integrations
- #6022 fix(widgets): hide system disk usage value when it overflows and show it in a tooltip

PR #6021, opened 2026-06-20
fix(widgets): hide health monitoring default tab option for non-proxmox integrations

afonsojramos:
## Summary

The health monitoring widget's **"Default tab"** option only affects the System/Cluster tabs, which render only for Proxmox. For every other integration (TrueNAS, Unraid, Glances, …) it was a dead setting shown in the widget options. Added a `shouldHide` guard so it only appears when a Proxmox integration is selected, consistent with the sibling cluster-only options (`visibleClusterSections`, `showUptime`, `sectionIndicatorRequirement`).

image

## Checklist
- [x] Targets `dev`
- [x] Conventional commits


manuel-rw (approved):

ajnart:
Thank you!

PR #6083, opened 2026-06-26
fix(plex): use series title for media releases instead of season label

### LINKED ISSUE #5755: bug: Homarr displays Plex Media Releases names incorrectly
### Describe the bug

When using the Media Releases item for Plex, instead of displaying the name of the TV show title, it uses the season name (eg.

Image

 Season 1, Season 2) instead of grouping the two seasons, and or displaying the actual title of the series (unsure if it is the same on Jellyfin/Ember)

### Steps to reproduce

Use the Media Releases Item with Plex and with a TV Show added to Plex.

### Impact

The impact isnt big, just something I've noticed.

### Additional information

_No response_

### Version

1.61.0

### Installation method

Docker Run

### Browser

Chrome

mvanhorn:
## Summary

The Plex "recently added" media-releases widget showed the wrong title for episodes and seasons. It used Plex's `title` field directly, which for an episode is the episode name and for a season is a label like "Season 1" rather than the series it belongs to. The result was a releases list where TV items were labelled "Season 1" / "Episode 3" with no indication of which show they came from.

This maps the displayed title to the series for episodes (`grandparentTitle`) and to the series for seasons (`parentTitle`), falling back to the item's own `title` when those are absent (movies, albums, anything without a parent). The more specific name is preserved as the subtitle so no information is lost.

## Why this matters

Issue #5755 reports that media releases display the season label instead of the series title, which makes the widget hard to use: a list of "Season 1", "Season 2" entries doesn't tell you what was actually added. Plex already carries the series name in `grandparentTitle` / `parentTitle`; this just uses it.

## Changes

`plex-integration.ts` selects the release title by item type before building the media object, and the recently-added Zod schema is extended to parse the `parentTitle` and `grandparentTitle` fields it now reads. The subtitle falls back to the original specific title when the main title is promoted to the series name, so the episode/season detail is still shown.

## Testing

Added `plex-integration.spec.ts` covering season, episode, movie, show, and missing-parent cases. Run with the repo's expected env: `CI=true NODE_ENV=development pnpm vitest run --exclude e2e packages/integrations/src/plex/test/plex-integration.spec.ts` — 1 file, all cases pass. The spec is pinned to the node environment (`// @vitest-environment node`) because the integration imports server-only env access.

Fixes #5755

### DIFF
--- a/packages/integrations/src/plex/plex-integration.ts
+++ b/packages/integrations/src/plex/plex-integration.ts
@@ -127,11 +127,18 @@ export class PlexIntegration extends Integration implements IMediaServerIntegrat
 
     const media =
       data.MediaContainer.Metadata?.filter((item) => item.Image).map((item) => {
+        const title =
+          item.type === "episode"
+            ? (item.grandparentTitle ?? item.title)
+            : item.type === "season"
+              ? (item.parentTitle ?? item.title)
+              : item.title;
+
         return {
           id: item.Media?.at(0)?.id.toString() ?? item.key,
           type: mapType(item.type),
-          title: item.title,
-          subtitle: item.tagline,
+          title,
+          subtitle: title === item.title ? item.tagline : item.title,
           description: item.summary,
           releaseDate: item.originallyAvailableAt
             ? new Date(item.originallyAvailableAt)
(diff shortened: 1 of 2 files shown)


packages/integrations/src/plex/plex-integration.ts
```diff
@@ -127,11 +127,18 @@ export class PlexIntegration extends Integration implements IMediaServerIntegrat
 
     const media =
       data.MediaContainer.Metadata?.filter((item) => item.Image).map((item) => {
+        const title =
+          item.type === "episode"
+            ? (item.grandparentTitle ?? item.title)
+            : item.type === "season"
+              ? (item.parentTitle ?? item.title)
+              : item.title;
+
```
ajnart, line 136: I really don't like double ternaries

ajnart:
Thank you for the PR and the new test 👍🏼

PR #6089, opened 2026-06-26
feat: Refactor widget-only integrations

ajnart:
## Summary

Removes widget-only services from the global Integration system and stores their configuration on the widgets that use them.

- Moves release provider, Search.ch, and ArchiveTeam Warrior request logic out of `@homarr/integrations`.
- Stores Releases provider, identifier, version filter, and provider URL directly in widget options.
- Adds public GHCR OCI support for full refs such as `ghcr.io/homarr-labs/homarr:latest`.
- Adds migration coverage to move old widget-only integration config into widget options and unlink stale integration rows.
- Updates docs, translations, and tests for the new widget-only configuration model.

## Why

These providers are not shared integrations. Keeping them as global integrations adds setup and database state that only one widget can use, and it made GHCR full image refs fail through the old release-provider path.

## Validation

- `pnpm turbo typecheck` passed.
- `pnpm lint` passed with existing warnings only.
- `pnpm vitest run packages/request-handler/src/test/release-providers.spec.ts packages/definitions/src/test/release-provider.spec.ts packages/definitions/src/test/docker-integration-match.spec.ts` passed.
- `pnpm vitest run packages/widgets/src/test/translation.spec.ts` passed.
- Live GHCR smoke passed for `ghcr.io/muchobien/pocketbase:latest` and `ghcr.io/homarr-labs/homarr:latest`.

## Note

`pnpm test packages/integrations/test/pi-hole.spec.ts` passed earlier during this work, but failed on rebased reruns because the Pi-hole container reported `domainsBeingBlocked` as `-2`/`0` instead of `> 1`. That test failure appears unrelated to this refactor.

## Summary by CodeRabbit

* **New Features**
  * Releases and Timetable widgets are now provider-based (including provider URLs).
  * Releases supports per-item provider token secrets (and a new secret management flow).
  * Timetable adds a configurable base URL (default: Search.ch).
  * Immich Album can randomize slideshow photo order.
* **Bug Fixes**
  * Improved handling of full registry/image identifiers (including GHCR).
* **Documentation**
  * Refreshed onboarding and updated Releases/Timetable docs and UI labels; removed outdated integration pages.
* **Breaking Changes**
  * Release providers and Search.ch were removed from the global integrations system and are configured within widgets instead.


ajnart:
@coderabbitai Remember that in PRs we don't need to modify `packages/translation/src/lang/*` files other than EN because these will be handled by Crowdin, it will automatically update the keys

ajnart:
@coderabbitai learn the following about the homarr codebase:

1. **Integration middleware provides the URL, not the caller.**
   When a tRPC procedure uses `createOneIntegrationMiddleware("query", "<kind>")`, the integration's URL is loaded from the database (configured by an admin). The procedure does NOT accept a caller-controlled URL. Do not flag these as SSRF vulnerabilities — the URL is pre-configured, not user-supplied at call time.

2. **Widget options with baseUrl are intentional, not SSRF.**
   Some widgets (timetable, releases) store a `baseUrl` in their widget options to support self-hosted service instances (e.g., self-hosted search.ch or GitLab). These URLs are configured by an admin in the widget editor, not by untrusted users. The trust boundary is the authenticated admin session, not the individual API call. Do not suggest restricting these to a hardcoded allow-list — that would break self-hosted setups.

3. **Migrations are applied once and are append-only.**
   Custom migrations in `packages/db/migrations/custom/` run once on existing databases. Correctness fixes are still valid (they help fresh installs), but do not frame them as critical/blocking since existing users have already run the migration. Focus the severity accordingly.

4. **The `publicProcedure` in widget routers is by design.**
   Widget data endpoints (releases, timetable, etc.) use `publicProcedure` because they serve unauthenticated dashboard viewers. The security model relies on the widget options being configured by an admin, not on per-request authentication of data fetches.
   
   
   Also, remember to use context7 to check that the comments you are making is relevant with the version of the package we are using or when there is an alternative better way or better method of implementing a certain feature by leveraging a package we are already using. Or when the PR's creator forget to use a pattern that they could benefit from. For example you made a comment about zod v3. 
   **`z.url()` and `z.string().url()` are NOT interchangeable in Zod v4.**
   `z.url()` returns a `URL` object. `z.string().url()` returns a validated `string`. When downstream code expects a string (passed to fetch, stored in DB, used in URL construction), `z.string().url()` is correct. Do not suggest migrating to `z.url()` without checking consumer types.

manuel-rw:
Interesting proposal, I think I generally agree with this.
It is (was) weird before, that they needed to be integrations.
Before I review this fully, @Meierschlumpf and @SeDemal can you confirm that you also agree with this proposal?
I think it is the right way to go. It removes potential for user confusion and removes complexity.
@ajnart please keep it open, I'll review properly after we discuss :)

SeDemal:
Well I can comment as someone that actually didn't know about this until recently. 
Having Integrations preregistered for me that I didn't want nor intended to use was weird. 
I remembered when we queried external services API like the weather, that was directly in the widget too. 
For me this makes much more sense too.

Meierschlumpf:
If you think it is better, feel free to adjust it. The reason I used integrations for it was, that
- We can support multiple services
- It has the same setup process as other connected services

~~Also don't forget that some of the integrations have secrets which should be encrypted in the db~~

Aandree5:
I've been a bit away because had a few changes that took all my time away! I'm now trying to get back into helping out.

@ajnart i will try to give my opinion soon. After I'm able test it and go through the changes.

Would this change then mean that if I have two `releases` widget i would need to configure the providers twice?

I remember at the beginning discussing this on discord and some of the points that @Meierschlumpf mentioned were raised. I understand it sounds a bit too overkill for a single widget, but there were some nice features about being able to have multiple even of the same type and they seamlessly work like all the other integrations, which they do allow Homarr to integrate with their respective system.
Another point was that we could potentially use them elsewhere. Like when homarr checks for version, instead of using no credentials we could actually use authenticated calls and not run into rate limiting issues (I'm not sure if any work was done on this or if something changed on that front that would render this point mute). Or even an issues or PR widget, deployments checker, GH actions monitor.

Just trying to think if this change would limit the app in any way.
If everyone agrees otherwise, I'm all up for making things easier.

Aandree5:
I was able to test the app but haven't fully gone through the code yet.
Here are my initial findings, don't want to look like I don't appreciate the work (because I do, when I first saw the token boxes I thought "that actually looks neat", but some of the features that will be lost I believe to be quite important for the people that use the widget), just want whichever decision the team ends up taking to be as informed as possible.

### Maybe just a bug
* I've added multiple Docker Hub providers, one with a different URL and there's only one box for adding a token - this could be an issue if you want to look at different location of the same provider (i.e. locally hosted and an online/work version).
* Adding new repositories when adding the widget for the first time will not show the token boxes, had to close and reopen the widget settings.
* DB migrations missing `widget_secret` (unless I missed something, sorry if so, I'm still in the "getting back to things" mode 😄)
```
Runtime TRPCClientError
no such table: widget_secret
```

### User experience changed
* Having multiple releases widgets will required retyping the tokens and any custom URLs.
* When tokens need updating, all the widgets will need updating. 
* Having multiple custom (or non-standard) repository provider URLs will require the user to type it against each tracked repository.
* Having the widget in different dashboards will required reconfiguring the providers, both tokens and custom URLs.
* Duplicating the widget will not carry the tokens across.
* This will no longer support different types of authentication, granted that tokens are probably the most used, but things like GitHub app authentication would require a different auth information (not a single token)

### Alternatives
* Declutter the integrations list, which I believe we could achieve by somehow separating the integrations by type for example.
* Stop creating integrations by default, I personally see nothing wrong with this, but have heard people mentioning in the past, we could do this by adding a tick box in the onboarding screen, something like "Initialise default integrations" and do it at that stage instead of via migrations. 
* Being able to use full image paths including version (code after ":"), this should be achievable via improving the version filter.
* The integrations having only a single widget, these are not the only integrations that are only used in a single widget.  
  
image
image

ajnart:
Thanks for the feedback on this PR @Aandree5 ! I think several points you mention are valid but might not be worth the effort to be addressed. 

- the fact that you would have to re-configure the widget when re-using it is a valid concern but it's a one-time pain I don't think anyone would complain about that 
- same thing with having multiple tokens, it's a good observation I had not thought about it. at the same time how many people would want to monitor releases for multiple private repositories they have access to that can't make a token that has access to both ? 

Thanks for the screenshots btw 

I view these tokens more as a way to prevent rate-limiting but they might not even be necessary, it could be over-engineering to have them in the first place (as widget secrets) because it adds friction when we could've put a more aggressive caching mechanism instead


Overall my goal with this PR is to drive adoption of this widget more than enhance its features, it's to make it easier to use without having to look at the readme 


Regarding db migration 
there's a new table for widget-scoped secrets that only used for the release api keys for now. - I guess in the future we can use that for the secret inside of a custom widget for example 

Let me know if you have ui/ux suggestions or feedback the other widgets I've done changes to

PR #6099, opened 2026-06-27
fix: remove conflicting nodeLinker=hoisted with enableGlobalVirtualStore

ajnart:
Both `enableGlobalVirtualStore` and `nodeLinker: hoisted` cannot be set simultaneously — GVS expects pnpm's isolated virtual store layout, but hoisted creates a flat `node_modules` with no virtual store at all. This caused Docker builds to fail with ENOTEMPTY errors during pnpm's linking phase.

Replaces `nodeLinker: hoisted` with `enableGlobalVirtualStore: true`.

See #6098 for context.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency in how values are converted to stable encoded strings, reducing unexpected differences for equivalent data.
  * Updated package workspace settings to use a different dependency linking approach, which may improve local install behavior and workspace reliability.
  * Adjusted logging infrastructure to keep transport behavior compatible with the current runtime setup.

### DIFF
--- a/packages/common/src/object.ts
+++ b/packages/common/src/object.ts
@@ -1,5 +1,3 @@
-import { hashKey } from "@tanstack/query-core";
-
 export function objectKeys<O extends object>(obj: O): (keyof O)[] {
   return Object.keys(obj) as (keyof O)[];
 }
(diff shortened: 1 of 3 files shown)


apps/docs/docs/getting-started/installation/helm.md
manuel-rw, line 1: revert

manuel-rw (changes requested):
Will this fix that my pnpm is deleting my whole src?

Image

ajnart:
> Will this fix that my pnpm is deleting my whole src?
> 
> Image

In my tests the new PNPM v11 option for dynamic linking worked on Linux & MacOS but the docs said explicitly it's not compatible with hoisted modules so the conflict might be the issue

manuel-rw:
Can you revert the unrelated changes? After, you can merge

ajnart:
Let's see if build passes now

PR #6122, opened 2026-06-29
fix(downloads-widget): make name column wider

### LINKED ISSUE #4157: feat: Download Client Item - Job Name column should be wider than the rest.

As per title. To avoid wasted space, job name should be wider than the other columns, or have a way to adjust columns. Not possible to see full Job Name without the other columns being so wide.

### Additional information

Image

alstondsouza1:
Closes #4157.

Made the download client job name column a bit wider so longer names are easier to read.

Testing:
- Ran git diff --check
- Could not run pnpm turbo typecheck --filter=@homarr/widgets because Corepack failed while verifying/downloading pnpm 11.9.0

## Summary by CodeRabbit

* **UI Improvements**
  * Improved the downloads table layout with responsive column sizing that adapts header and column spacing as the widget width changes.
* **New Features**
  * Added a reusable helper to compute responsive table column widths from base sizes and available width, with optional per-column maximum limits.
* **Tests**
  * Added unit tests covering narrow vs. wide table behavior and enforcing per-column max size constraints.

### DIFF
--- a/packages/widgets/src/downloads/component.tsx
+++ b/packages/widgets/src/downloads/component.tsx
@@ -48,6 +48,7 @@ import { getIconUrl, getIntegrationKindsByCategory } from "@homarr/definitions";
 import type { ExtendedClientStatus, ExtendedDownloadClientItem } from "@homarr/integrations";
 import { useScopedI18n } from "@homarr/translation/client";
 
+import { getResponsiveTableColumnSizes } from "../common/table-column-sizing";
 import type { WidgetComponentProps } from "../definition";
 
 dayjs.extend(relativeTime);
@@ -58,7 +59,7 @@ interface QuickFilter {
 }
 
 //Ratio table for relative width between columns
-const columnsRatios: Record<keyof ExtendedDownloadClientItem, number> = {
+const baseColumnsRatios: Record<keyof ExtendedDownloadClientItem, number> = {
   actions: 2,
   added: 4,
   category: 1,
@@ -83,6 +84,7 @@ export default function DownloadClientsWidget({
   integrationIds,
   options,
   setOptions,
+  width,
 }: WidgetComponentProps<"downloads">) {
   const integrationsWithInteractions = useIntegrationsWithInteractAccess().flatMap(({ id }) =>
     integrationIds.includes(id) ? [id] : [],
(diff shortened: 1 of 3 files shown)


manuel-rw (changes requested):
This is a reoccurring issue: #5713 
We should implement a more suitable solution.
e.g. change width depending on widget width.
@homarr-labs/maintainers what do you think?

ajnart:
> This is a reoccurring issue: #5713 
> We should implement a more suitable solution.
> e.g. change width depending on widget width.
> @homarr-labs/maintainers what do you think?

I 100% agree with this, generally speaking we have a lot of tables in the project that should have the exact same base logic and adapt 

Also we could swap to the more advanced mantine data tables library and save all of the settings (vol width, order, sort, filters, ...) as JSON in a hidden option for that widget. I've recently added an hidden option (I forgot which widget) + the ability to modify that setting inside of the widget itself even outside of edit mode.

alstondsouza1:
Updated this based on the review.

Instead of only bumping the name column to one fixed value, it now uses the widget width to scale the name column ratio. Small widgets keep the old baseline, and wider widgets give more room to the job name column.

Also rebased on latest dev.

Testing:
- Ran git diff --check
- Tried pnpm turbo typecheck --filter=@homarr/widgets, but Corepack still fails locally while verifying pnpm

packages/widgets/src/downloads/component.tsx
```diff
@@ -78,11 +78,17 @@ const columnsRatios: Record<keyof ExtendedDownloadClientItem, number> = {
   upSpeed: 3,
 };
 
+const getColumnsRatios = (width: number): Record<keyof ExtendedDownloadClientItem, number> => ({
+  ...baseColumnsRatios,
+  name: Math.min(32, Math.max(baseColumnsRatios.name, Math.round(width / 32))),
+});
```
manuel-rw: Could be extracted to be reused in other tables

manuel-rw (changes requested):

alstondsouza1:
Updated this again based on the latest review.

I moved the responsive column sizing into a reusable helper at packages/widgets/src/common/table-column-sizing.ts, then used it in the downloads widget for the name column.

Also rebased on latest dev.

Testing:
- Ran git diff --check
- Tried pnpm turbo typecheck --filter=@homarr/widgets again, but Corepack still fails locally while verifying pnpm

alstondsouza1:
Fixed the CodeRabbit nitpick.

The responsive table sizing helper now treats maxSize as a growth cap only, so it cannot shrink a column below its base size even if a smaller maxSize is passed.

Testing:
- Ran git diff --check

alstondsouza1:
Added a focused unit test for the responsive table column sizing helper.

The new spec covers:
- narrow tables keeping base sizes
- wider tables growing the configured name column
- maxSize capping growth
- maxSize not shrinking below the base size

Testing:
- Ran git diff --check
- Tried pnpm vitest run packages/widgets/src/common/table-column-sizing.spec.ts, but Corepack still fails locally before Vitest starts while verifying pnpm

PR #6124, opened 2026-06-29
fix: make production source maps opt-in to restore v1.67.0 memory baseline

### LINKED ISSUE #6123: bug: about 50% increased memory requirments in 1.68.0
### Describe the bug

Memory baseline from ~600MB to ~1000MB when viewing dashboard after updating to 1.68.0. Downgrading to 1.67.0 restores the usage.

v1.67.0
Image

v1.68.0
Image

### Steps to reproduce

Open dashboard and look at memory usage.

### Impact

Out of memory issues because I had my deploy limit set to 1024M.

### Additional information

Image

### Version

1.68.0

### Installation method

Docker Compose

### Browser

Firefox

ajnart:
## Problem

v1.68.0 introduced a **~50% memory regression** (600MB → 1000MB, reported in #6123). After investigation, the root cause is:

**[`--enable-source-maps`](https://nodejs.org/api/cli.html#--enable-source-maps)** — introduced in #6031 — causes Node.js to load **all** server chunk `.map` files into memory and keep them resident. When `serverSourceMaps: true` generates maps for every Next.js chunk and `productionBrowserSourceMaps: true` generates them for client bundles, this adds hundreds of MB of resident memory.

## Before / After

| | v1.67.0 | v1.68.0 (regression) | This PR (fix) |
|---|---|---|---|
| `productionBrowserSourceMaps` | `false` (default) | `true` (hardcoded) | removed (defaults to `false`) |
| `serverSourceMaps` | not set | `true` (hardcoded) | removed (defaults to `false`) |
| `NODE_OPTIONS` | not set | `--enable-source-maps` | removed |
| `.map` files shipped | no | yes (all server chunks) | no |
| Idle RSS | ~600MB | ~1000MB (+67%) | ~600MB (restored) |
| Production stack traces | minified | readable | minified |

### Evidence

| Source | Finding |
|---|---|
| [Node.js docs](https://nodejs.org/api/cli.html) | "Enabling source maps can introduce latency… caching of Source Maps" |
| [nodejs/node#46140](https://github.com/nodejs/node/issues/46140) | User measured **631MB → 999MB** after errors with `--enable-source-maps` |
| [nodejs/node#41541](https://github.com/nodejs/node/issues/41541) | "unnecessarily slow with large source files" — impacts CPU + memory |
| [nodejs/node#43812](https://github.com/nodejs/node/issues/43812) | Performance warning was added to Node docs for this flag |
| [Next.js Memory Guide](https://nextjs.org/docs/app/guides/memory-usage) | "Disable source maps" to reduce memory |

### Investigation Notes

AI-assisted investigation via OpenCode. The agent:
1. Traced the full diff between v1.67.0 and v1.68.0 (60 commits, 130 files)
2. Identified `89380f81d` (source maps in production) as the primary suspect
3. Cross-referenced against Node.js upstream memory issues (#46140, #41541, #43812)
4. Verified that none of the other changes (new VPN widget, Beszel cache changes, dependency bumps) could account for a 400MB server-side increase
5. Confirmed the fix by benchmarking Node.js source-map memory behavior from upstream reports

## Fix

Revert the three source map changes from #6031.

### Changes

| File | Change |
|---|---|
| `apps/nextjs/next.config.ts` | Remove `productionBrowserSourceMaps: true` and `experimental.serverSourceMaps: true` |
| `Dockerfile` | Remove `.map` copy step and `ENV NODE_OPTIONS=--enable-source-maps` |
| `.github/workflows/code-quality.yml` | New `memory-regression` job: builds Docker image, starts container, asserts RSS < 700MB |

## Memory Regression Guard

A new CI job `memory-regression` builds the production Docker image, starts it, warms up, and asserts memory stays under 700MB (the v1.67.0 baseline). This catches memory regressions before they ship.

## Verification

- [x] oxlint: no new issues
- [x] oxfmt: format OK
- [x] bash -n: run.sh syntax valid
- [x] YAML: CI workflow parses correctly
- [x] CodeRabbit: both comments addressed (ARG placement + health check fail-fast)

Closes #6123

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -188,13 +188,20 @@ jobs:
         run: |
           docker run -d --name homarr-mem-test -p 17575:7575 homarr-memory-test
           # Wait for container to be healthy (up to 60s)
+          healthy=false
           for i in $(seq 1 30); do
             if curl -sf http://localhost:17575/api/health 2>/dev/null; then
               echo "Container healthy after $((i * 2))s"
+              healthy=true
               break
             fi
             sleep 2
           done
+          if [ "$healthy" != "true" ]; then
+            echo "ERROR: Container never became healthy"
+            docker logs homarr-mem-test --tail 50
+            exit 1
+          fi
           # Warm up: hit the dashboard page a few times
           for i in $(seq 1 5); do
             curl -sf http://localhost:17575/ > /dev/null 2>&1 || true
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -163,3 +163,58 @@ jobs:
         run: cp .env.example .env
       - name: Build
         run: pnpm turbo build --affected
+
+  memory-regression:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v6
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+      - name: Build Docker image
+        uses: docker/build-push-action@v7
+        with:
+          platforms: linux/amd64
+          context: .
+          push: false
+          load: true
+          tags: homarr-memory-test
+          network: host
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+        env:
+          SKIP_ENV_VALIDATION: true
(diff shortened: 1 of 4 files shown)


ajnart:
@coderabbitai review

PR #6125, opened 2026-06-29
feat(boards): add board-wide open all apps in new tabs action 

### LINKED ISSUE #3771: feat(boards): open all apps

I love Homarr, and use the open all in each category, everyday. It would be great to have a single dropdown at the top of Homarr that would open ALL of the apps in ALL of the categories in new tabs; rather than going into each cat and clicking open all.

Thank you for the great work!

### Additional information

Originally posted https://github.com/ajnart/homarr/issues/1928

hirrrooo:
## Description
This PR adds a board-wide "open all apps in new tabs" action button to the board content header, allowing users to open every app on the current board layout in new tabs with a single click — rather than having to go into each category individually.

Closes #3771 

## Changes

- **New board header button** (\`_header-actions.tsx\`): Adds an external-link icon button that collects all app IDs on the current layout and opens them in new tabs. Visible for both editors (in view mode) and read-only users.
- **Shared helper** (\`open-apps-in-new-tabs.ts\`): Extracted \`openAppsInNewTabs(appIds, { t, openConfirmModal })\` so the board-level and category-level actions share the same logic. Tests included.
- **Category action** (\`category-menu-actions.tsx\`): Now uses the shared helper (also picks up dedup + security fixes for free).
- **Documentation** (\`after-the-installation.mdx\`): Added a note about the board-level open all action.

## Implementation details

- App IDs are de-duplicated before fetching to avoid duplicate tabs and redundant network/DB work.
- Tabs are opened with \`noopener,noreferrer\` to prevent reverse tabnabbing.
- Popup-blocked tabs trigger the existing confirm modal.

## Checklist

- [x] Builds without warnings or errors (\`pnpm build\`)
- [x] Formatted with \`pnpm format:fix\`
- [x] Pull request targets \`dev\` branch
- [x] Commits follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) guideline
- [x] No shorthand variable names
- [x] Documentation is up to date

# Screenshot
image
image

## Summary by CodeRabbit

* **New Features**
  * Added an “open all in new tabs” action from board headers and category menus to launch multiple apps at once.
  * Allows opening apps for the current board layout in new tabs.
* **Bug Fixes**
  * De-duplicates apps and improves handling when browsers block pop-ups, including a clearer confirmation flow.
* **Documentation**
  * Updated the “Using categories” instructions to reflect the new tab-opening actions.
* **Tests**
  * Added coverage for opening multiple apps in new tabs.


ajnart:
Can you add screenshots ?

hirrrooo:
Hi @ajnart, I apologize about the late reply, I've added the screenshots. 

Looking at it now, should I perhaps make the button present whilst in edit mode but make it appear as disabled/grayed out? Could possibly use a tooltip here too, denoting that the button is unavailable while in edit mode.

I feel like on a note of UI consistency, completely removing the button causes the edit icon to shift right. When toggling back and forth, the layout shifting can be a bit jarring and disrupt muscle memory.

# Screenshot

Made a draft of what it would look like.

image
image

ajnart:
I'd rather not add this feature in this shape/ui to be honest, I think the population of users who have this problem is much much lower than the population of users who would be disturbed by such button. My advice would be to just use the group feature that allows to do that, that way you can filter the entries you want to open directly in that way. 
Closing for now.

PR #6179, opened 2026-07-04
fix: demo/mock widgets crashing — namespace request-handler cache and complete SuperJSON cache fix

ajnart:
## Problem

Since #6075, all mock-integration widgets on the demo fail with TypeErrors (`e.systems.map is not a function`, `t.time is undefined`, `e.imageUrls is undefined`, `a.name.localeCompare`, …). #6172 was supposed to fix it but didn't fully.

Three root causes:

1. **Cross-handler cache collision (main cause).** The in-memory request-handler cache introduced in #6075 keys entries by `JSON.stringify(input)` in a single `Map` shared by **all** handlers, where input is only `{ integrationId, options }`. Any two handlers querying the same integration with equal options (commonly `{}`) collide and read each other's data. With the mock integration serving every widget kind, all demo widgets received whichever handler's payload was cached first — e.g. `widget.beszel.getSystems` returned the health-monitoring payload (`systems: { cpuModelName: "Mock CPU", … }` instead of an array). `updateCheckerRequestHandler` and `duckDuckGoBangsRequestHandler` collided the same way (both key on `{}`).

2. **Half-merged SuperJSON fix.** Commit `161e8a937` (branch `feat/simplify-caching-tanstack-persister`) fixed cache serialization in three places, but only the client persister landed via #6172. The server-side `QueryCacheHydration` still used `JSON.parse`, which can't unwrap the superjson blobs the client now writes — so Redis-backed hydration silently returned `null` on every load, and legacy JSON blobs hydrated with mangled Dates. The defensive null guards across ~30 widget components never landed either.

3. **Demo mode persist spam.** `queryCache.setItem` is a mutation and demo mode rejects all mutations, so the persister retried a request that can never succeed every throttle tick.

## Changes

- **Namespace the request-handler cache per handler definition** (`createRequestHandler` gets a stable auto-generated `cacheKey`; `createIntegrationRequestHandler` passes its own stable key since it re-creates the inner handler per call). No call-site changes needed.
- **Cherry-pick the lost half of `161e8a937`**: `SuperJSON.parse` in `QueryCacheHydration` (Dates/Maps survive the Redis round-trip; legacy JSON blobs fail the `clientState` guard and are discarded, so no Redis flush is needed) + null guards across ~30 widget components.
- **Stop persister retries in demo mode**: after the first `FORBIDDEN` response the persister disables itself for the page load.

## Verification

- Fresh DB seeded with `DEMO_MODE=true UNSAFE_ENABLE_MOCK_INTEGRATION=true`, logged in as demo/demo: all 36 widgets render correct mock data, zero console/page TypeErrors, exactly one 403 persist probe per page load. (Pre-fix, the reported errors reproduced exactly, with the beszel query cache verifiably containing health-monitoring data.)
- `DEMO_MODE=false`: persister writes a superjson blob to `qc:{userId}:{boardId}` (24h TTL); after reload the hydrated `notification.time` / `releaseDate` are real `Date` instances.
- `tsc` clean on `@homarr/widgets`, `nextjs`, `@homarr/request-handler`; request-handler (32) and query-cache (14) test suites pass.

Fixes the demo regression from #6075; completes #6172.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability across multiple widgets when certain data fields are missing, reducing empty-screen and runtime-error issues.
  * Fixed cache persistence behavior so read-only server rejections no longer trigger repeated failed retries.
  * Updated data loading and sorting in dashboards to handle timestamps and serialized values more consistently.
* **Refactor**
  * Made request handling cache behavior more stable, helping reuse cached results more predictably across repeated handler creation.

### DIFF
--- a/apps/nextjs/src/app/[locale]/_client-providers/query-cache-persister.ts
+++ b/apps/nextjs/src/app/[locale]/_client-providers/query-cache-persister.ts
@@ -1,22 +1,39 @@
 import { createAsyncStoragePersister } from "@tanstack/query-async-storage-persister";
+import { TRPCClientError } from "@trpc/client";
 import superjson from "superjson";
 
 import { fetchApi } from "@homarr/api/client";
 import { getActiveQueryCacheBoardId, queryCacheStoragePrefix } from "@homarr/api/query-cache";
 
+// Set once the server rejects a persist mutation as read-only (demo mode),
+// so we stop retrying a mutation that can never succeed.
+let persistDisabled = false;
+
+const isReadOnlyError = (error: unknown) => error instanceof TRPCClientError && error.data?.code === "FORBIDDEN";
+
 const queryCacheStorage = {
   getItem: (_key: string) => null as string | null,
   async setItem(_key: string, value: string) {
+    if (persistDisabled) return;
     const boardId = getActiveQueryCacheBoardId();
     if (!boardId) return;
     await fetchApi.queryCache.setItem.mutate({ boardId, value }).catch((error) => {
+      if (isReadOnlyError(error)) {
+        persistDisabled = true;
+        return;
+      }
       console.warn("[query-cache] persist failed", error);
     });
   },
   async removeItem(_key: string) {
+    if (persistDisabled) return;
     const boardId = getActiveQueryCacheBoardId();
     if (!boardId) return;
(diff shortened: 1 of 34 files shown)


ajnart:
Closing this (exploration PR by claude) just in-case we still have issues later.

PR #6249, opened 2026-07-07
fix(docker): report current CPU usage instead of the lifetime average

### LINKED ISSUE #5544: bug: Docker widget cpu calculation is container lifetime average instead of "current"
### Describe the bug

The docker containers widget reports each container's lifetime-average CPU usage instead of its current CPU usage. The widget claims "updated a minute ago" — which is true of the sample time, but the value sampled is cumulative, not instantaneous.

I believe this was working correctly before PR #4266. That broke cpu stats and PR #4311 changed it to lifetime CPU rather than instantaneous.

### Steps to reproduce

Start a container and create a high CPU load. Stop the load and watch the widget. The CPU usage will stay high.

### Impact

Makes the CPU usage in the docker container less useful.

### Additional information

_No response_

### Version

1.59.2

### Installation method

Docker Compose

### Browser

Chrome

junkerderprovinz:
Closes #5544

## Problem

The Docker containers widget reports each container's **lifetime-average** CPU usage instead of its **current** usage — a value stays high long after a load spike has ended (as described in the issue, regressed by #4266 / #4311).

## Cause

The stats are requested in one-shot mode:

```ts
.stats({ stream: false, "one-shot": true })
```

One-shot mode makes the daemon return immediately **without** `precpu_stats` (the previous reading). `calculateCpuUsage` was therefore left dividing the cumulative `total_usage` by the cumulative `system_cpu_usage`, which is exactly the lifetime average.

## Fix

- Request the stats **without** one-shot so the daemon includes `precpu_stats`.
- Compute the usage from the **delta** between the current and previous readings, which is [Docker's own CPU-percentage formula](https://github.com/docker/cli/blob/master/cli/command/container/stats_helpers.go).

When `precpu_stats` is absent (e.g. Podman, or the very first reading) the deltas fall back to the cumulative totals, so the previous behaviour is preserved for those runtimes.

## Tests

Extended `docker.spec.ts`: one test proving the delta (a large lifetime `total_usage` but a small recent delta yields the real current percentage) and one proving a container that was busy over its lifetime but is idle now reports `0`. All existing `calculateCpuUsage` tests still pass.

## Checklist

- [x] Builds without errors (`tsc --noEmit` clean on `@homarr/request-handler`)
- [x] `oxfmt` formatted, `oxlint` clean
- [x] Targets `dev`; conventional commit; no shorthand variable names

Note: dropping one-shot means the stats call now takes the usual ~1s Docker sampling interval; the per-container calls already run in parallel, so this only adds that interval once per refresh.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container CPU usage reporting so values now reflect recent activity more accurately instead of relying on lifetime totals.
  * Fixed cases where CPU usage could appear inflated or incorrect when a container had little or no change since the last reading.
  * Added coverage to ensure CPU usage shows `0` when there is no measurable CPU change.

### DIFF
--- a/packages/request-handler/src/docker.ts
+++ b/packages/request-handler/src/docker.ts
@@ -300,7 +300,10 @@ async function getContainersWithStatsAsync() {
 
     const stats = await instance
       .getContainer(container.Id)
-      .stats({ stream: false, "one-shot": true })
+      // Not in one-shot mode: the daemon then includes precpu_stats (the
+      // previous reading), which is required to compute the current CPU usage
+      // as a delta. One-shot omits it and would only expose lifetime totals.
+      .stats({ stream: false })
       .catch(
         () =>
           ({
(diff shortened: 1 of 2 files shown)


ajnart:
Sounds good, thank you for this!

PR #6250, opened 2026-07-07
fix(docker): report container CPU as a total-system percentage

### LINKED ISSUE #6082: bug: Unraid integration CPU calculation is incorrectly reporting CPU load
### Describe the bug

Looks like there may be a bug in the calculation for CPU when trying to show system resources. My overall CPU load recently was around 42-46% in Unraid. However, in Homarr it was incorrectly saying it was 93%. (At times it was even going above 100% - such as 105-108%)

In Unraid, when I enter "uptime" from the terminal, the load average returned at the time was: 5.12, 5.35, and 5.36. I have 12 cores. So (5.12 / 12) x 100 = 43% which is what aligns with the 42-46% load I was seeing in Unraid. 

Image

### Steps to reproduce

See above in bug description.

### Impact

CPU load in homarr from Unraid Integration is unusable/can't be relied on.

### Additional information

_No response_

### Version

1.67.0

### Installation method

Unraid Apps

### Browser

_No response_

junkerderprovinz:
This branch started as the Unraid CPU fix for #6082. That landed separately in #6510 on 2026-08-22, so the Unraid hunk is gone and what is left is a different fix on a different file. Retitling rather than opening a new PR, so the review history stays attached.

## Problem

`calculateCpuUsage` in `packages/request-handler/src/docker.ts` divides the container's cumulative `total_usage` by the cumulative `system_cpu_usage`. Both counters run from container start, so the result is the container's lifetime average, not what it is doing now. A container that was busy an hour ago keeps reporting a high number long after the load ended.

It then multiplies by `online_cpus`, following docker stats' per-core convention, where 134% means 1.34 busy cores. Every other CPU reading in Homarr is a 0-100% share of the machine, so the same dashboard shows two different scales.

## Fix

Compare the latest reading against the previous one, which is what `precpu_stats` is for, and report the plain ratio. `system_cpu_usage` already accumulates across all cores, so the ratio is the whole-machine share and cannot exceed 100%.

`precpu_stats` is absent when stats are fetched in one-shot mode, Podman being the case that matters here, so the deltas fall back to the cumulative totals and behave as before rather than dividing by zero.

## The part worth a second opinion

Dropping the `online_cpus` multiplier changes what the number means: a container using 1.34 cores on an 8 core host reported 134% before and reports about 17% now. I think matching the rest of Homarr's 0-100% scale is the right call, but it is a visible change and it is yours to make. Say so if you would rather keep docker stats' convention and I will restore the multiplier while keeping the delta fix.

## Tests

`docker.spec.ts` covers the delta calculation, the fallback when `precpu_stats` is missing, and that several fully busy cores stay at or below 100%. The two existing expectations that encoded the old per-core scale were updated with the arithmetic spelled out.

### DIFF
--- a/packages/integrations/src/unraid/unraid-integration.ts
+++ b/packages/integrations/src/unraid/unraid-integration.ts
@@ -36,9 +36,6 @@ export class UnraidIntegration extends Integration implements ISystemHealthMonit
   public async getSystemInfoAsync(): Promise<SystemHealthMonitoring> {
     const systemInfo = await this.getSystemInformationAsync();
 
-    const cpuUtilization = systemInfo.metrics.cpu.cpus.reduce((acc, val) => acc + val.percentTotal, 0);
-    const cpuCount = systemInfo.info.cpu.cores;
-
     // We use "info" object instead of the stats since this is the exact amount the kernel sees, which is what Unraid displays.
     const totalMemory = systemInfo.info.memory.layout.reduce((acc, layout) => layout.size + acc, 0);
     const usedMemory = totalMemory * (systemInfo.metrics.memory.percentTotal / 100);
@@ -47,7 +44,12 @@ export class UnraidIntegration extends Integration implements ISystemHealthMonit
     return {
       version: systemInfo.info.os.release,
       cpuModelName: systemInfo.info.cpu.brand,
-      cpuUtilization: cpuUtilization / cpuCount,
+      // Use the overall CPU utilization the Unraid API already reports (0-100),
+      // which is the value Unraid itself displays. Previously the per-thread
+      // percentages were summed and divided by the number of physical cores,
+      // which double-counted on hyper-threaded systems (e.g. 24 threads over
+      // 12 cores) and could even exceed 100%.
+      cpuUtilization: systemInfo.metrics.cpu.percentTotal,
       memUsedInBytes: usedMemory,
       memAvailableInBytes: totalMemory - usedMemory,
       uptime: dayjs().diff(uptime, "seconds"),
--- a/packages/request-handler/src/docker.ts
+++ b/packages/request-handler/src/docker.ts
@@ -335,13 +335,25 @@ export function calculateCpuUsage(stats: ContainerStats): number {
     return 0;
   }
 
-  const numberOfCpus = stats.cpu_stats.online_cpus;
-  const usage = stats.cpu_stats.system_cpu_usage;
-  if (!usage || usage === 0) {
+  // Instantaneous CPU usage: compare the latest reading against the previous one
+  // (precpu_stats). Dividing the cumulative total_usage by the cumulative
+  // system_cpu_usage instead would report the container's lifetime average, which
+  // stays high long after a load spike has ended. precpu_stats is only present
+  // when the stats are NOT fetched in one-shot mode; when it is absent (e.g.
+  // Podman) the deltas fall back to the cumulative totals.
+  const cpuDelta = stats.cpu_stats.cpu_usage.total_usage - (stats.precpu_stats?.cpu_usage?.total_usage ?? 0);
+  const systemDelta = (stats.cpu_stats.system_cpu_usage ?? 0) - (stats.precpu_stats?.system_cpu_usage ?? 0);
+
+  if (systemDelta <= 0 || cpuDelta < 0) {
     return 0;
   }
 
-  return (stats.cpu_stats.cpu_usage.total_usage / usage) * numberOfCpus * 100;
+  // Report usage as a share of the WHOLE machine (0-100%), not docker-stats'
+  // per-core convention (which multiplies by online_cpus and can exceed 100%,
+  // e.g. "134%" for 1.34 busy cores). system_cpu_usage already accumulates
+  // across all cores, so the plain ratio is the total-system share. This keeps
+  // every CPU reading in homarr on the same intuitive 0-100% scale.
+  return (cpuDelta / systemDelta) * 100;
 }
 
 export function calculateMemoryUsage(stats: ContainerStats): number {
--- a/packages/request-handler/src/test/docker.spec.ts
+++ b/packages/request-handler/src/test/docker.spec.ts
@@ -64,16 +64,45 @@ describe("calculateCpuUsage", () => {
     const stats = createStats({
       cpu_stats: { online_cpus: 4, cpu_usage: { total_usage: 2000 }, system_cpu_usage: 10000 },
     });
-    // (2000 / 10000) * 4 * 100 = 80
-    expect(calculateCpuUsage(stats)).toBe(80);
+    // Total-system share: (2000 / 10000) * 100 = 20 (system_cpu_usage spans all cores)
+    expect(calculateCpuUsage(stats)).toBe(20);
   });
 
   test("should handle fractional CPU usage", () => {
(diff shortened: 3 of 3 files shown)


ajnart:
Somewhat related (although I understand the issue you're solving here), your previous PR with Docker (https://github.com/homarr-labs/homarr/pull/6249) made it so that the CPU can go over 100%. Users are complaining that they are not expecting this kind of reading going over 100%. 
In Portainer, the CPU usage by a container logic is the same as the current Homarr, but in Beszel for example, it shows the total CPU usage (not on a per-core basis) 

I think it might be beneficial to define and think about how we want to show CPU usage across the whole homarr to normalize it. I am of the opinion that seeing the strain on the total system `3.5%` makes more sense that a value like `134%` 
What's your opinion on this @junkerderprovinz ?

junkerderprovinz:
@ajnart Really good point, and I agree this is worth settling consistently across homarr rather than per-widget.

The two readings just answer different questions: per-core (can go over 100%) is "how hard is this container working the cores it has", like docker stats or Portainer, so 134% means ~1.34 cores busy, while total-system % (0-100%) is "how much of the whole machine is it using", like Beszel, which is more intuitive at a glance and never surprises anyone.

For a dashboard I think you're right that total-system % is the better default: 3.5% is immediately meaningful and nobody has to wonder why a number went past 100%. My suggestion would be to normalize to total-system % everywhere by default (divide by the core count), and keep the per-core detail as an option or a tooltip if we still want it.

This PR is really just making the per-core number correct on hyper-threaded Unraid hosts in the first place, so it slots in under either choice. I'm happy to adapt it (and revisit #6249) to emit the normalized total-system % so everything lines up. Just say the word and I'll push the change. Thanks for thinking this through with me!

ajnart:
@junkerderprovinz feel free to push the change here as well, then we'll merge both inside of this PR

junkerderprovinz:
@ajnart Done! I rebased onto main and pushed the Docker normalization into this PR (1dc4fa02).

The change itself is tiny: `system_cpu_usage` already accumulates across all cores, so dropping the `* online_cpus` from the delta ratio directly yields the share of the whole machine. Container CPU is now 0-100% total-system, matching what the Unraid fix in this PR reports, so every CPU reading in homarr sits on the same intuitive scale - no more 134%.

Tests updated accordingly, plus a new case asserting that multiple fully busy cores can never push the reading past 100% (18/18 passing locally). Thanks for pushing for consistency here, I think the dashboard is better for it!

junkerderprovinz:
Rebased onto the latest `dev` — the `calculateCpuUsage` conflict is resolved (kept the instantaneous delta-vs-`precpu_stats` calc; the existing no-`precpu` tests still pass because the deltas fall back to the cumulative totals). `MERGEABLE` again and green from my side. Ready whenever you are, @ajnart.

junkerderprovinz:
Hi @ajnart — this is rebased onto the latest `dev`, the `calculateCpuUsage` conflict is resolved, and checks are green. Ready whenever you'd like to merge both fixes here as you suggested. Thanks!

junkerderprovinz:
Hi @ajnart, following up after a couple of weeks. This is still rebased onto `dev`, conflict-free and green, and the Docker over-100% CPU normalization is folded in as you asked. Happy to adjust anything if it helps, otherwise it's ready whenever you have a moment. Thanks!

junkerderprovinz:
@ajnart heads up, #6510 (opened 2026-08-03) fixes the same issue #6082 in the same file, dividing the summed per-thread total by cpu count rather than reading percentTotal like this PR does. Both should produce the same number, just worth knowing this one exists before merging either, in case you want to compare or just pick whichever's simpler to review. This one's still CI-green and rebased if it's useful.

ThatTallGuy21:
Any updates on this fix?

junkerderprovinz:
Nothing has changed here since July, and that turns out to be the problem: this branch does not do what its description says. calculateCpuUsage subtracts precpu_stats, but the call site above still reads with one-shot: true, and the daemon sends no pre-read in that mode, so both deltas fall back to the lifetime counters. My first version dropped one-shot for exactly that reason and a rebase on 16 July quietly put it back.

What does work as described is the scale: the reading is no longer multiplied by online_cpus, so it is a share of the whole machine rather than of a single core.

Dropping one-shot again is the fix, but it is not free, because the daemon then holds the request while it collects a second sample, once per container. That trade is yours to make rather than mine, so I would rather ask than change it quietly. Say the word and I will push it.

ThatTallGuy21:
Thanks for the response! Do you know whether those container stats requests are made in parallel or sequentially?

ajnart:
I think this PR / issue is kind of cursed these two ways of measuring cpu usage both work but some people complain one and some people complain about the other, I merged a PR that did a similar change and people complained about it so I reverted it, I think ideally we'll want this kind of CPU indication to be a personal preference across the app just like I made the KiB / Kb preference in v2. 

As of now I don't think it's relevant to work on this as we have more pressing changes with v2 ongoing

junkerderprovinz:
Fair enough, and the preference route makes sense to me. Two notes before this goes on the shelf.

@ThatTallGuy21 on your question: parallel, and uncapped. The stats calls are all started inside `containers.map(async ...)` and only collected afterwards by `Promise.all`, with no limiter in that path and no custom agent on dockerode, so nothing queues on Homarr's side. That also means the wait I mentioned on the 13th should be paid once per refresh, roughly one to two seconds, not once per container.

The other note is not about scale, and it applies even if this PR stays parked. #6782 ported this calculation to `release/v2` on 6 September, and it took the same one-shot problem along: the call site there still reads with `"one-shot": true` ([docker.ts:426](https://github.com/homarr-labs/homarr/blob/4305dd3b7dbe30032e93cf6cc1691b89349babca/packages/request-handler/src/docker.ts#L426)). In that mode the precpu values come back as zero, so both deltas collapse to the raw counters and the widget reports a container's total CPU time divided by the host's cumulative CPU time rather than a current reading. On Docker, a container started an hour ago on a host up for a month reads near zero whatever it is doing. That is wrong under either scale, so whichever way the preference goes, v2 needs a real delta underneath it. I am happy to open an issue so it does not get lost, or a PR against `release/v2` if you would rather have the fix.

ThatTallGuy21:
@junkerderprovinz - Thanks for clarifying, that answers my question about the performance impact. I'm perfectly fine waiting for v2. I do agree the one-shot issue sounds like a separate correctness bug, independent of the CPU scale/preference discussion. If you're willing, I'd appreciate a PR against release/v2 for the real-delta fix so the underlying CPU calculation is correct when v2 lands.

junkerderprovinz:
@ThatTallGuy21 happy to. @ajnart one question before I open it, since it touches a choice you made in #6782.

The fix is ready on a branch and it drops `one-shot` again, which #6782 kept on purpose to avoid the wait for a second sample. On a Docker 29 host with 36 running containers, the stats calls in a refresh that misses the cache then take about 2 seconds instead of a few milliseconds, inside the existing 5 second timeout per call. It does not help Podman, which only fills `precpu_stats` when streaming.

The other way keeps `one-shot` and computes the delta against the counters from the previous refresh. That adds no wait and works for Podman too, but it is more code: the first read has nothing to compare against, and the previous values have to live in Redis so every replica sees them.

Which would you take, or would you rather leave it for now?

PR #6301, opened 2026-07-11
chore: upgrade TypeScript from v6 to v7 (tsgo native port)

ajnart:
## Summary

- Bumps `typescript` from `^6.0.3` to `^7.0.2` in the pnpm catalog
- TypeScript 7 is a native Go port (tsgo) delivering **8-12x faster builds**, shared memory multithreading, a rebuilt `--watch` mode, and a new LSP-based language server
- Zero tsconfig changes needed — the codebase was already compatible with all TS7 defaults and deprecation-to-error promotions (`esModuleInterop` always on, `strict` default true, no `baseUrl`, no `moduleResolution: node`, etc.)
- No TS API consumers in the codebase (oxlint, not typescript-eslint), so no need for the `@typescript/typescript6` side-by-side setup

### Key TS7 changes (all already satisfied):
- `strict: true` by default — already set
- `module: esnext` by default — already set
- `types: []` by default — already explicitly `["node"]`
- `moduleResolution: Bundler` — still supported (only `node/node10` removed)
- `esModuleInterop` forced true — was already true
- `noUncheckedSideEffectImports` true by default — no issues found

Reference: https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/

## Test plan

- [x] `pnpm install` succeeds
- [x] `npx tsc --version` → 7.0.2
- [x] `pnpm turbo typecheck` — all 37 packages pass with zero errors
- [ ] CI passes

## Summary by CodeRabbit

* **Chores**
  * Updated the TypeScript workspace dependency to version 7.0.2 or later.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -517,8 +517,8 @@ catalogs:
       specifier: ^2.10.2
       version: 2.10.2
     typescript:
-      specifier: ^6.0.3
-      version: 6.0.3
+      specifier: ^7.0.2
+      version: 7.0.2
     undici:
       specifier: 7.28.0
       version: 7.28.0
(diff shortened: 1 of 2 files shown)


arturict:
Thanks for this major dependency upgrade. Could you document any tooling behavior changes in local build scripts for TS upgrade in description?

PR #6307, opened 2026-07-12
feat(docker-widget): replace footer timestamp text with compact refresh button

Bergasha:
Replaces the always-visible "Updated a few seconds ago" text in the Docker widget's footer with a compact icon-only refresh button. The previous text took up too much horizontal space in the footer, competing with the CPU/Memory totals — especially on narrower widget widths.

Screenshot 2026-07-12 143405
Screenshot 2026-07-12 145532
Screenshot 2026-07-12 145547

- [ x Documentation is up to date. Create a pull request [here](https://github.com/homarr-labs/documentation/).

## Summary by CodeRabbit

* **New Features**
  * Added a manual refresh action to the Docker Containers widget.
  * Displayed “Last refresh {when}” using a relative time label with a tooltip.
  * Added a loading/spinner indicator on the refresh button while container data is updating.

### DIFF
--- a/packages/translation/src/lang/en.json
+++ b/packages/translation/src/lang/en.json
@@ -5204,7 +5204,6 @@
     "table": {
       "updated": "Updated {when}",
       "refresh": {
-        "action": "Press to refresh",
         "lastUpdated": "Last refresh {when}"
       },
       "search": "Seach {count} containers",
--- a/packages/translation/src/lang/en.json
+++ b/packages/translation/src/lang/en.json
@@ -5203,6 +5203,10 @@
     "title": "Containers",
     "table": {
       "updated": "Updated {when}",
+      "refresh": {
+        "action": "Press to refresh",
+        "lastUpdated": "Last refresh {when}"
+      },
       "search": "Seach {count} containers",
       "selected": "{selectCount} of {totalCount} containers selected",
       "footer": "Total {count} containers",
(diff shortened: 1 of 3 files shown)


ajnart:
Sounds good, I agree it took too much space. Why not have it in the right-click context menu as well ? I think it would be nice to make it be there (could be only there or both)

Bergasha:
> Sounds good, I agree it took too much space. Why not have it in the right-click context menu as well ? I think it would be nice to make it be there (could be only there or both)

Yeah good idea. Maybe both for now as this gives users time to get used to right click context menu.

packages/translation/src/lang/en.json
```diff
@@ -5203,6 +5203,10 @@
     "title": "Containers",
     "table": {
       "updated": "Updated {when}",
+      "refresh": {
+        "action": "Press to refresh",
+        "lastUpdated": "Last refresh {when}"
+      },
```
ajnart, line 5208: I think you can leverage the translations we have for the context menu here. Or just don't put one for press to refresh, it's pretty self-explanatory

Bergasha: I did think that but you never know haha, Ill remove it now. Looks like the refresh is already in the context menu but does nothing. Ill address that now to.

Bergasha: Done and done, thanks mate should be ready to go.

ajnart:
Thanks @Bergasha !

PR #6312, opened 2026-07-12
feat(media-requests): filter requests list by status and time window

### LINKED ISSUE #4561: feat: Ability to filter seerr requests
### Seerr integration controls

When using the overseerr or Jellyseerr integrations, the app shows recent and pending requests. It would be much more usable in my opinion to allow a filter option to allow showing either only certain statuses eg pending + processing, as well as set the definition of recently to last <#> <Unit> so you can choose the number of either days or weeks to count as recent

### Additional information

_No response_

ajnart:
## Description

Closes #4561

Adds three new options to the **Media Requests List** widget so users can narrow the Overseerr/Jellyseerr requests list to specific request statuses and a configurable recent time window.

### New widget options

| Option | Type | Default | Description |
| --- | --- | --- | --- |
| `statusFilter` | multi-select | all 5 statuses | Only requests with these statuses are shown |
| `recentDays` | number (0-365) | `0` | Time-window length. `0` disables the filter |
| `recentUnit` | select (`days` / `weeks`) | `days` | Unit used by `recentDays` |

### Behavior

- The new options default to the previous behavior (all statuses, no time filter) so existing widgets keep working without reconfiguration.
- Filtering is applied server-side in the request handler / tRPC `getLatestRequests` procedure.
- The status list reuses the existing five homarr request statuses (`pending`, `approved`, `declined`, `failed`, `completed`).

## Changes

- `packages/widgets/src/media-requests/list/index.ts` — new widget options
- `packages/widgets/src/media-requests/list/component.tsx` — pass new options to the tRPC query
- `packages/request-handler/src/media-request-list.ts` — input schema + server-side filter
- `packages/api/src/router/widgets/media-requests.ts` — accept input on `getLatestRequests`
- `packages/old-import/src/widgets/options.ts` — migration mapping (`undefined` for new options)
- `packages/translation/src/lang/en.json` — translation keys for the new options
- `apps/docs/docs/widgets/media-request-list/{index.mdx,index.ts}` — docs

## Verification

- `pnpm turbo typecheck` — passes for the 6 affected packages
- `pnpm lint` — 0 errors
- `pnpm oxfmt --check` — all touched files formatted

### DIFF
--- a/packages/widgets/src/media-requests/list/component.tsx
+++ b/packages/widgets/src/media-requests/list/component.tsx
@@ -38,6 +38,11 @@ export default function MediaServerWidget({
 }: WidgetComponentProps<"mediaRequests-requestList">) {
   const { data: mediaRequests } = clientApi.widget.mediaRequests.getLatestRequests.useQuery({
     integrationIds,
+    statuses:
+      options.statusFilter.length > 0
+        ? options.statusFilter
+        : ["pending", "approved", "declined", "failed", "completed"],
+    recentDays: options.recentDays > 0 ? options.recentDays * (options.recentUnit === "weeks" ? 7 : 1) : 0,
   });
 
   if (!mediaRequests) return <WidgetEmptyState />;
--- a/apps/docs/docs/widgets/media-request-list/index.mdx
+++ b/apps/docs/docs/widgets/media-request-list/index.mdx
@@ -1,41 +1,44 @@
 ---
-title: 'Media Request List'
-description: 'See a list of all media requests from your integration'
+title: "Media Request List"
+description: "See a list of all media requests from your integration"
 hide_title: true
 ---
 
-import { WidgetHeader } from '@site/src/components/widgets/header';
-import { AddingWidget } from '@site/src/components/widgets/adding';
-import { WidgetConfig } from '@site/src/components/widgets/configuration';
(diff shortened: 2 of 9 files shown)


packages/widgets/src/media-requests/list/component.tsx
```diff
@@ -38,6 +38,11 @@ export default function MediaServerWidget({
 }: WidgetComponentProps<"mediaRequests-requestList">) {
   const { data: mediaRequests } = clientApi.widget.mediaRequests.getLatestRequests.useQuery({
     integrationIds,
+    statuses:
+      options.statusFilter.length > 0
+        ? options.statusFilter
+        : ["pending", "approved", "declined", "failed", "completed"],
+    recentDays: options.recentDays > 0 ? options.recentDays * (options.recentUnit === "weeks" ? 7 : 1) : 0,
```
arturict, line 45: The weeks conversion can produce values that the request schema rejects. `recentDays` is valid up to 365 in the widget option (and the PR docs advertise 0–365 combined with either unit), but selecting 53 weeks sends 371 here while `mediaRequestListInputSchema` caps the normalized value at 365. The widget then fails its query for every weeks value above 52. Please either validate weeks against 52 in the option/UI, or let the handler receive the amount and unit and apply a consistent limit after conversion. A boundary test for 52/53 weeks would catch this.

PR #6330, opened 2026-07-13
docs(integrations): define the CPU utilization convention (total-system %, 0-100)

junkerderprovinz:
Follow-up to the discussion in #6250, where @ajnart suggested defining how CPU usage should be shown across the whole of homarr.

This PR writes that convention down and aligns the code with it:

**The convention:** `cpuUtilization` is the share of the WHOLE machine, in percent (0-100). A machine with 3 of 4 cores fully busy reports 75, never 300. Readings like `134%` (docker-stats per-core style) are not allowed to reach widgets.

**Changes:**

- Documented the convention as JSDoc on `SystemHealthMonitoring.cpuUtilization`, so every current and future integration author sees it right at the source of truth.
- Audited all integrations that report `cpuUtilization` today: glances (`cpu.total`), truenas (per-core average), synology, openmediavault and unraid (after #6250) already follow the convention. dashdot also does, but its internal field was misleadingly named `sumLoad` while actually holding the per-core *average* - renamed to `averageLoad` so nobody "fixes" it into a real sum later.
- Clamped the health-monitoring CPU ring geometry at 100 as a defensive guard against misreporting sources (the text label keeps showing the raw value, so a genuinely broken source stays visible).
- Container CPU on the Docker side is aligned to the same convention in #6250.

No behavioural change for any integration that already followed the convention.

### DIFF
--- a/packages/integrations/src/dashdot/dashdot-integration.ts
+++ b/packages/integrations/src/dashdot/dashdot-integration.ts
@@ -39,7 +39,7 @@ export class DashDotIntegration extends Integration implements ISystemHealthMoni
     const history = await channel.getSliceUntilTimeAsync(dayjs().subtract(15, "minutes").toDate());
 
     return {
-      cpuUtilization: cpuLoad.sumLoad,
+      cpuUtilization: cpuLoad.averageLoad,
       memUsedInBytes: memoryLoad.loadInBytes,
       memAvailableInBytes: info.maxAvailableMemoryBytes - memoryLoad.loadInBytes,
       network: networkLoad,
@@ -97,7 +97,7 @@ export class DashDotIntegration extends Integration implements ISystemHealthMoni
     // we convert it to text as the response is either valid json or empty if cpu widget is disabled.
     if (result.length === 0) {
       return {
-        sumLoad: 0,
+        averageLoad: 0,
         averageTemperature: 0,
       };
     }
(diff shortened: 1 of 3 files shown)


ajnart:
This PR doesn't seem to do anything ?
How about instead we just revert the previous one and focus on a more simplistic approach later / just open an issue for the CPU concern ? 

Then I can tackle it later

junkerderprovinz:
Fair points, and thanks for taking a look!

You're right that #6330 is deliberately light. It's really just docs (the JSDoc convention) plus two tiny guards: renaming dashdot's sumLoad to averageLoad so nobody later "fixes" that average into a real sum, and clamping the ring at 100 while the label still shows the raw value. No behaviour change at all. If that's more than you want right now, I'm totally happy to close it and open an issue instead that just captures the convention (cpuUtilization = share of the whole machine, 0 to 100) for whenever you feel like tackling it properly.

The one thing I'd gently flag is reverting the actual fixes. #6249 (already merged) and #6250 each fix a real, user-reported wrong reading: Docker was showing the lifetime average instead of the current value, and Unraid was double-counting hyper-threaded cores (#6082, where a genuine 43% showed up as 93%). A revert would just hand those bugs back to users. Could we keep those two bugfixes and let you take the broader convention and cleanup whenever it suits you? Happy to close #6330 in the meantime.

Either way, I'm glad to go with whatever keeps this simplest on your side. Thanks again!

PR #6360, opened 2026-07-15
fix(qbittorrent): remove progress sort to include seeding torrents in limited fetches

ajnart:
## Problem

PR #6297 added `sort: "progress", reverse: false` to fix the case where users with many completed torrents couldn't see active downloads. However, this created the inverse bug: users with many torrents (e.g. 970) now only see the 50 least-complete torrents, and **uploading/seeding torrents never appear** because they're at progress=1 — always at the end of the sorted list, outside the limit cutoff.

## Root Cause

`sort: "progress", reverse: false` systematically excludes completed torrents from the limited result set. Client-side filtering cannot show what it never receives from the server.

## Fix

Removed the `sort` and `reverse` params from the qBittorrent `listTorrents` call. qBittorrent returns torrents in its default order (by hash/add order), giving both downloading and seeding torrents a fair chance to appear in the first `limit` results. Client-side filtering (showCompletedTorrent toggle, category filter, threshold) handles visibility as designed.

## Files Changed

- `packages/integrations/src/download-client/qbittorrent/qbittorrent-integration.ts` — 1 line

## Summary by CodeRabbit

* **Improvements**
  * qBittorrent downloads can now follow the ordering selected via new widget options, instead of using a fixed progress-based sort.
  * Backend and UI wiring updated so the selected qBittorrent sorting key and optional reverse-order preference are applied when fetching torrent jobs.
* **UI/Refinement**
  * Media server table rendering for remaining time and transcoding status was reformatted without changing displayed results.
* **Translations**
  * Added labels for the new qBittorrent sorting settings.

### DIFF
--- a/packages/api/src/router/widgets/downloads.ts
+++ b/packages/api/src/router/widgets/downloads.ts
@@ -22,10 +22,20 @@ export const downloadsRouter = createTRPCRouter({
       },
     })
     .concat(createDownloadClientIntegrationMiddleware("query"))
-    .input(z.object({ limitPerIntegration: z.number().default(50) }))
+    .input(
+      z.object({
+        limitPerIntegration: z.number().default(50),
+        qbittorrentSort: z.string().optional(),
+        qbittorrentSortReverse: z.boolean().optional(),
+      }),
+    )
     .query(async ({ ctx, input }) => {
       return await settleIntegrationQueries(ctx.integrations, async (integration) => {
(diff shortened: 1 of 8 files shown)


ajnart:
Superseded by #6430 — the one-line fix (2000e1f) has been cherry-picked into the downloads widget rework. The configurable server-side sort option is unnecessary since the widget handles sorting client-side.

PR #6393, opened 2026-07-18
feat(boards): add keyboard shortcuts for board switching

### LINKED ISSUE #6392: feat: Switch Board via keyboard shortcut

I'd like the option to switch boards using a keyboard shortcut.  The ability to assign a shortcut on a per board basis would allow for quicker user navigation.  For example `control + shift + 1`, `control + shift + 2` etc.  At a minimum  you should be able to cycle through all boards using `control + shift + >` (next board) and `control + shift + <` (previous board)

### Additional information

_No response_

ajnart:
## Description

Adds keyboard shortcuts to quickly switch between boards without using the mouse.

### Shortcuts
- **`mod+shift+ArrowRight`** — switch to next board
- **`mod+shift+ArrowLeft`** — switch to previous board
- **`mod+shift+1`** through **`mod+shift+9`** — jump to board by index (1-based)

`mod` = Ctrl on Windows/Linux, Cmd on macOS.

### Implementation
- Uses Mantine's `useHotkeys` hook (already used in the same file for edit mode toggle)
- Shortcuts are ignored when typing in inputs (Mantine default behavior)
- Extracted into a clean `BoardNavigationHotkeys` component
- Fixed a small logic issue: the board switcher menu is now always visible, even for users without change access
- Arrow shortcuts are guarded so they only register when the current board is found in the cache (`currentIndex !== -1`), preventing an unexpected jump before the boards list has loaded
- Extracted the wraparound index calculation into a pure `getNextBoardIndex` helper with unit tests

Closes #6392

### Out of scope
The Windows-style board switcher **overlay** (Alt-Tab style UI with board cards) discussed in the comments is tracked separately in [#6513](https://github.com/homarr-labs/homarr/issues/6513), so this PR stays focused on the keyboard shortcuts themselves.

### DIFF
--- a/apps/nextjs/src/app/[locale]/boards/(content)/_board-navigation.spec.ts
+++ b/apps/nextjs/src/app/[locale]/boards/(content)/_board-navigation.spec.ts
@@ -1,0 +1,23 @@
+import { describe, expect, test } from "vitest";
+
+import { getNextBoardIndex } from "./_board-navigation";
+
+describe("get next board index", () => {
+  test.each([
+    [3, 0, 1, 1],
+    [3, 2, 1, 0],
+    [3, 2, -1, 1],
+    [3, 0, -1, 2],
+  ] as const)("moves through boards with wraparound", (boardCount, currentIndex, direction, expected) => {
+    expect(getNextBoardIndex(boardCount, currentIndex, direction)).toBe(expected);
+  });
+
+  test.each([
+    [0, 0],
+    [1, 0],
+    [3, -1],
+    [3, 3],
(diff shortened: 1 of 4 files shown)


Meierschlumpf:
Just had the idea that we could also make a menu like on windows to switch between boards (but of course no need to implement this)

ajnart:
> Just had the idea that we could also make a menu like on windows to switch between boards (but of course no need to implement this)

Like what kind of menu ? The one to switch between spaces ? I think someone opened an issue about it recently

Meierschlumpf:
image
Something like this, just for the boards

ajnart:
@Meierschlumpf its a good idea

PR #6396, opened 2026-07-18
fix(immich): use thumbnail endpoint instead of original for album carousel

### LINKED ISSUE #6388: bug: immich `getSystemHealthStatus` Query Fails (TRPCClientError) Due to HEIC Image Incompatibility
### Describe the bug

When the frontend loads the `immich album` widget and triggers the `widget.healthMonitoring.getSystemHealthStatus` query, a `TRPCClientError` is thrown in the console, and the corresponding image resource fails to render.

Upon inspecting the **Network tab**, the image request returns a `Content-Type` of **`application/octet-stream`** instead of the expected `image/*` format.

### Steps to reproduce

1. Upload a HEIC image
2. Select an album it have HEIC image

### Impact

HEIC image can't show

1. **HEIC Compatibility Issue:** The image currently being used is in **HEIC** format (the default photo format for iOS devices). Most web browsers (e.g., Chrome, Edge, Firefox) do not natively support or render HEIC images. As a result, the server/browser falls back to treating it as a generic binary stream (`application/octet-stream`).
2. **Loading Original Full-Res Images:** The frontend is currently pulling and rendering the **original uncompressed image**. HEIC files from mobile devices are typically very large, causing high bandwidth consumption, slow response times, and an increased likelihood of decoding errors.

### Additional information

I recommend switching from serving the original image to thumbnails(jpeg or webp).

### Version

1.70.0

### Installation method

Docker Run

### Browser

Firefox

ajnart:
## Description

Switches the Immich album carousel from fetching the original asset to using the thumbnail endpoint.

### Before
`GET /api/assets/:id/original` — returns the raw file (HEIC from iPhones = broken in browsers, huge file sizes)

### After
`GET /api/assets/:id/thumbnail` — returns a JPEG thumbnail (browser-compatible, ~50-100KB)

### Fixes
- **HEIC compatibility**: Thumbnails are always JPEG, so iPhone photos render correctly in all browsers
- **Bandwidth**: Dramatically smaller images (~50-100KB vs multi-megabyte originals)
- **Speed**: Faster loading for album carousels
- **Permissions**: thumbnail uses `AssetView` permission (subset of `AssetDownload` required for original)

Closes #6388

## Summary by CodeRabbit

* **Bug Fixes**
  * Immich album asset links now use thumbnail previews instead of original media, improving performance and reducing bandwidth.
  * The image proxy now returns clearer HTTP status codes and consistent error signaling when upstream requests fail or the image can’t be found.
* **New Features**
  * Image proxy responses now provide accurate `Content-Type` on success and apply consistent caching headers.
* **Documentation**
  * Updated Immich integration docs to specify the required “Asset View” permission for loading thumbnail carousel images (otherwise thumbnail requests may return `403`).


ajnart:
Confirmed during manual verification: the proxy sends the API key correctly. Immich returned 403 until the key was granted **Asset View** permission; granting that permission fixed thumbnail loading. The integration documentation now states this requirement.

PR #6399, opened 2026-07-19
fix(widgets): refetch after render errors

### LINKED ISSUE #6398: bug: Media Server Streams crashes after v1.71 upgrade when hydrating a v1.70 query cache
After upgrading Homarr from v1.70.0 to v1.71.0, the **Current Media Server Streams** widget can fail with:

```text
TypeError: Cannot read properties of undefined (reading 'state')
```

The media-server integration remains healthy, and a fresh `widget.mediaServer.getCurrentStreams` response contains the new `currentlyPlaying.playback` object. The crash occurs when Homarr hydrates a Media Server query cached by v1.70, where `currentlyPlaying` exists but `playback` does not.

v1.70 and v1.71 both use the query-cache buster `v2-superjson`, and widget queries are persisted for up to 24 hours. In v1.71, `packages/widgets/src/media-server/component.tsx` accesses:

```ts
currentlyPlaying.playback.state
```

without guarding against a hydrated pre-v1.71 session object.

An already-open client can also persist the old query state back to the server after the Redis key is removed, so clicking **Try again** or hard-refreshing does not necessarily recover the widget.

## Steps to reproduce

1. Run Homarr v1.70.0 with an Emby integration and a Current Media Server Streams widget.
2. Allow the widget to persist a successful query while a session is playing. The cached v1.70 session contains `currentlyPlaying` but no `currentlyPlaying.playback`.
3. Upgrade Homarr to v1.71.0 without clearing the persisted widget query cache.
4. Open or reload the board.
5. Observe that the Streams widget fails with:

   ```text
   TypeError: Cannot read properties of undefined (reading 'state')
   ```

Minimal reproduction of the incompatible object shape against the v1.71 access pattern:

```js
const session = {
  currentlyPlaying: {
    type: "tv",
    name: "Example channel",
    metadata: null,
  },
};

session.currentlyPlaying.playback.state;
// TypeError: Cannot read properties of undefined (reading 'state')
```

## Expected behavior

Homarr should invalidate incompatible Media Server query data during the upgrade, migrate it, or tolerate a missing `playback` object until the fresh v1.71 query replaces it. The widget should not crash on persisted data produced by the immediately preceding release.

Possible remedies include changing the query-cache buster for this schema change or defensively handling `currentlyPlaying.playback` when rendering.

## Impact

The Streams widget remains unusable even though the media server and Homarr integration are healthy. Media playback itself is unaffected.

Workarounds are to fully remove the stale query state after closing all clients/site state, or temporarily downgrade to v1.70.0. A normal hard refresh may not work because an open client can write the stale state back into Homarr's server-side query cache.

## Additional information

- Homarr: v1.71.0
- Installation: Docker Compose
- Host: TrueNAS SCALE/Linux
- Browser: Edge (Chromium)
- Integration: Emby
- The exact behavior was reproduced using the v1.70 session shape and the v1.71 widget access pattern.
- A fresh v1.71 API response was verified to contain `playback` for every active session, isolating the failure to hydrated persisted query data rather than the Emby configuration.
- The current development branch still uses `queryCacheBuster = "v2-superjson"` and accesses `currentlyPlaying.playback.state` without a compatibility guard.
- Related change: https://github.com/homarr-labs/homarr/pull/6315

A screenshot of the widget error can be attached separately.

Clarkar00:
## Summary

- remove persisted board-query data before resetting a crashed widget's error boundary
- reuse Homarr's existing persisted-query predicate so widget, app, Docker, and integration query families recover consistently
- preserve unrelated non-persisted queries while allowing widgets to fetch fresh data

## Root cause

The widget error boundary currently invalidates widget queries when **Try again** is clicked. Invalidating a successful persisted query leaves its stale data available while the refetch starts. The error boundary can therefore render the same incompatible value immediately, crash again, and never stay mounted long enough to consume fresh data.

Removing persisted widget data before resetting the boundary leaves no stale value to render. When the widget remounts, React Query performs fresh requests, and the cache's normal persistence subscription records the recovered state. This recovery is scoped to the user who clicks **Try again** rather than globally invalidating every user's cache.

## Validation

- regression coverage was run RED first for the missing recovery path and for widget kinds whose query path differs from their kind
- `pnpm exec vitest run apps/nextjs/src/components/board/items/widget-query-recovery.spec.ts packages/api/src/test/query-cache.spec.ts --coverage.enabled=false` (19 tests passed)
- `pnpm --filter @homarr/nextjs typecheck`
- `pnpm exec oxfmt --check` on changed files
- `pnpm exec oxlint` on changed files (0 errors; one pre-existing warning in `item-content.tsx`)
- `git diff --check`

Closes #6398

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery from widget rendering errors by clearing stale persisted widget-related cached data before retrying.
  * Ensured only widget-specific persisted queries are removed, leaving unrelated application data unaffected.

* **Tests**
  * Added Vitest coverage for persisted widget query cleanup across multiple recovery scenarios to confirm correct scoping and preservation of unrelated queries.


ajnart:
@Clarkar00 fair enough but won't you just need to wait 48 hours for the cache keys to be invalidated for everyone for it to naturally fix itself ? 
I think generally you raise a fair point, its that a failing query can't load the widget and therefore can re-query itself, going into a bad cycle where it can't re-trigger the request and keeps getting handed the broken cache values

Clarkar00:
Good point — if the key is left to expire naturally, the issue should eventually resolve. My concern was the loop you described, where the widget fails before it can recover normally.

While reproducing it, I found the affected Redis key still had nearly the full 24-hour TTL remaining and contained the older payload without `playback`. Since persisting the query cache resets that TTL, I wondered whether an actively loaded affected board could keep refreshing the incompatible value rather than allowing it to expire.

That was the reasoning behind using both the cache-buster and the defensive fallback. But I’m happy to narrow the PR to only the fallback if you think that better fits the project.

ajnart:
> Good point — if the key is left to expire naturally, the issue should eventually resolve. My concern was the loop you described, where the widget fails before it can recover normally.
> 
> While reproducing it, I found the affected Redis key still had nearly the full 24-hour TTL remaining and contained the older payload without `playback`. Since persisting the query cache resets that TTL, I wondered whether an actively loaded affected board could keep refreshing the incompatible value rather than allowing it to expire.
> 
> That was the reasoning behind using both the cache-buster and the defensive fallback. But I’m happy to narrow the PR to only the fallback if you think that better fits the project.

I believe narrowing the PR would be better. Ideally you can make it so that a widget that crashes can still trigger a re-fetch and doesn't get stuck in a stale-data loop

Clarkar00:
Thanks — I narrowed the PR and changed the recovery path as suggested. **Try again** now removes persisted widget-query data before resetting the error boundary, so the remounted widget fetches fresh data instead of immediately rendering the same stale value. I also removed the global cache-buster and the Media Server-specific fallback, and added regression coverage for widget, app, Docker, and mismatched widget/query namespaces.

PR #6412, opened 2026-07-20
fix(integrations): stop auto-seeding ArchiveTeam Warrior on first install

ajnart:
Removes defaultUrl from the ArchiveTeam Warrior integration definition so it is no longer auto-created as a default integration on first install.

The seeding logic auto-creates any integration that has both a `defaultUrl` and a no-auth option. ArchiveTeam Warrior had both, causing it to be seeded on every new install even though it's a niche service. Users who want it can still add it manually.

- Removed `defaultUrl: "http://localhost:8001"` from `archiveTeamWarrior` definition in `packages/definitions/src/integration.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed an outdated fallback URL from the TeamWarrior integration configuration.


ajnart:
Closing in favor of a clearer fix

PR #6434, opened 2026-07-21
fix(widgets): prevent stale cache from serving errored data

### LINKED ISSUE #6398: bug: Media Server Streams crashes after v1.71 upgrade when hydrating a v1.70 query cache

ajnart:
## Summary

Add two layers on top of PR #6399's error boundary recovery to comprehensively fix the global issue of React Query's cache serving stale/errored data instead of allowing fresh refetches.

### Layer 1 — QueryCache subscriber (`query-cache-error-cleanup.ts`)
Subscribe to `QueryCache` events. When a persistable query transitions from `success` to `error`, `resetQueries` clears stale data so widgets show error/loading UI instead of rendering incompatible cached values. Uses `resetQueries` (not `removeQueries`) so active `useQuery` observers get a refetch triggered rather than being orphaned with a destroyed query reference.

### Layer 2 — Rehydration invalidation (`trpc.tsx`)
Add `onSuccess` callback to `PersistQueryClientProvider` that invalidates all persistable queries after rehydration. Combined with `staleTime: 0`, this ensures rehydrated persisted data is immediately marked stale and refetches on mount rather than being served as fresh.

### Layer 3 — Error boundary reset (from PR #6399)
`removePersistedWidgetQueries` in the error boundary's `onReset` wipes all persistable queries before the widget remounts. Handles render crashes where the query is still in `success` status but the data structure is incompatible.

## Root cause

React Query v5 has three mechanisms interacting badly:

1. **Rehydration**: Persisted data is `setQueryData`'d as **fresh**. `useQuery` serves it immediately on first render. If a background refetch fails, `{ data: stale, isError: true }` is returned — widgets show old data with no error indication.
2. **Error state**: `useQuery` returns `data` even when `status === "error"`. Components render stale data instead of the error, masking the failure.
3. **Error boundary reset** (fixed by PR #6399): `invalidateQueries` kept data in cache. Widget remounts, renders same stale data, crashes again → infinite loop.

## Validation

- `pnpm --filter @homarr/nextjs typecheck`
- `pnpm exec vitest run apps/nextjs/src/components/board/items/widget-query-recovery.spec.ts --coverage.enabled=false` (8 tests passed)
- `pnpm exec oxlint` on changed files (0 new errors)
- `pnpm exec oxfmt --check` on changed files
- `git diff --check`

Supersedes #6399

Closes #6398

### DIFF
--- a/apps/nextjs/src/app/[locale]/_client-providers/query-cache-error-cleanup.ts
+++ b/apps/nextjs/src/app/[locale]/_client-providers/query-cache-error-cleanup.ts
@@ -1,0 +1,29 @@
+import type { QueryClient } from "@tanstack/react-query";
+
+import { isPersistableWidgetQueryKey } from "@homarr/api/query-cache";
+
+export const createQueryCacheErrorCleanup = (queryClient: QueryClient): (() => void) => {
+  const wasSuccessful = new Set<string>();
+
+  return queryClient.getQueryCache().subscribe((event) => {
+    const query = event.query;
+    const key = query.queryHash;
+
+    if (event.type === "removed") {
+      wasSuccessful.delete(key);
+      return;
+    }
+
+    if (event.type !== "updated") return;
+
+    if (query.state.status === "success" && isPersistableWidgetQueryKey(query.queryKey)) {
+      wasSuccessful.add(key);
+      return;
+    }
+
+    if (query.state.status === "error" && wasSuccessful.has(key)) {
+      queryClient.resetQueries({ queryKey: query.queryKey, exact: true });
+      wasSuccessful.delete(key);
+    }
+  });
+};
--- a/apps/nextjs/src/app/[locale]/_client-providers/trpc.tsx
+++ b/apps/nextjs/src/app/[locale]/_client-providers/trpc.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import type { PropsWithChildren } from "react";
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import type { QueryKey } from "@tanstack/react-query";
 import { QueryClient } from "@tanstack/react-query";
 import { ReactQueryDevtools } from "@tanstack/react-query-devtools";
@@ -36,6 +36,7 @@ import { env } from "@homarr/common/env";
 import { showWarningNotification } from "@homarr/notifications";
 import { widgetImports } from "@homarr/widgets";
 
+import { createQueryCacheErrorCleanup } from "./query-cache-error-cleanup";
 import { createWidgetQueryPersister } from "./query-cache-persister";
 
 const getWebSocketProtocol = () => {
@@ -102,6 +103,10 @@ export function TRPCReactProvider(props: PropsWithChildren) {
     return client;
   });
 
+  useEffect(() => {
+    return createQueryCacheErrorCleanup(queryClient);
+  }, [queryClient]);
+
   const [trpcClient] = useState(() => {
     return clientApi.createClient({
       links: [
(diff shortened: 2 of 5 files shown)


ajnart:
Closing in favor of #6484, which has already been merged.

The two PRs address related cache behavior but have conflicting failure policies: #6484 preserves the last successful widget data when all selected integrations temporarily fail, while this PR's global success-to-error cache reset would clear that data.

The safe render-crash recovery remains separately scoped in #6399: incompatible persisted data is removed only after the user clicks **Try again**. The generic query-error reset should not be merged because it would also react to ordinary network and upstream integration failures.

This decision was prepared with assistance from Codex, an LLM.

PR #6442, opened 2026-07-22
feat(widgets): add live tennis widget

bensynapse:
Adds a Live Tennis widget showing live, upcoming or completed matches from the ATP, WTA, Challenger, ITF and junior tours. Homarr currently has no sports widget of any kind.

### Why a dedicated widget rather than `customApi`

`customApi` genuinely can render this, and I don't want to pretend otherwise: it fetches server-side with encrypted credentials, and its JSX template whitelists `Table`/`Badge`/`Group` with `.map()` and ternaries via `SAFE_BINDINGS`. It even whitelists `PaginatedList` and `TabsContainer`, so interactivity is available too.

The case for a dedicated widget is **domain rendering**, not saved configuration effort. The API returns per-set game counts as `games[0]`/`games[1]`, which have to be split across two players and aligned with `sets`, plus a `server` index and current game points. Every user would reconstruct that in a JSX template by hand — unversioned, untranslated, untested. This ships it typed, translated and covered.

I'll be straight that this is a moderate justification rather than a slam dunk. **If you'd rather it stayed a `customApi` recipe in the docs, that's a reasonable call and I'll close this.**

### API key handling

The key is read server-side from `LIVE_TENNIS_API_KEY` and is deliberately **not** a widget option.

`getBoardByName` in `packages/api/src/router/board.ts` is a `publicProcedure`, and it returns item options unredacted:

```ts
options: superjson.parse<Record<string, unknown>>(item.options),
```

So a credential stored in a widget option would be readable by anonymous visitors on a public board. That also matches the existing code: no widget stores a credential in an option today, and `widgetSecrets` is scoped to one kind (`if (!item || item.kind !== "releases") throw`). If you'd prefer this go through `widgetSecrets` instead, I'm happy to rework it.

Free tier, no card. Only free-tier endpoints are used — `/analysis` and `/events` return 403 on the free plan, so nothing is built against them and they aren't advertised.

### Manual smoke test

Verified against the live API for live, upcoming and completed matches and per-tour filtering, including doubles (null country) and upcoming matches (null score).

Offline and error paths: a missing key or a 401 renders "No valid Live Tennis API key configured" with a retry button; 5xx and DNS failure hit the standard widget error boundary; `NO_EXTERNAL_CONNECTION` short-circuits without making a request. 13 unit tests cover the mapping and every error branch.

### Checks

`pnpm build` passes (4/4 tasks). `pnpm lint` shows no new findings — per-package warning counts are byte-identical to `dev`, and no lint output mentions tennis.

One heads-up: `@homarr/db` lint is **already failing on `dev`** (`no-shadow`, 5 errors), and `db` depends on `definitions`, which any new widget must touch — so `--affected` may surface that here through no fault of this PR. I reproduced it on a pristine checkout of `cf0e469` to confirm.

### Docs

In-tree at `apps/docs/docs/widgets/tennis/`. The PR template still points at `homarr-labs/documentation`, but that repo is archived, so I followed `AGENTS.md` and kept it to one PR. English-only translations; no other locales touched. No temp files.

### AI assistance

This PR was written with AI assistance (Claude). All build, lint, test and live-API output was executed and verified rather than assumed, and both the `customApi` trade-off and the `publicProcedure` credential finding were checked against the source before being claimed here.

## Summary by CodeRabbit

* **New Features**
  * Added a Live Tennis dashboard widget with live, upcoming, and completed match views.
  * Added widget configuration for tour, match status, match count (1–20), and optional tournament name and player ranking display.
  * Added a new documentation page and a Widgets environment variable section for `LIVE_TENNIS_API_KEY`.
* **Bug Fixes**
  * Improved API-key error handling with a tennis-specific unauthorized state and no retries for auth failures.
  * Added correct behavior when external connections are disabled, plus robust response parsing and validation.

### DIFF
--- a/apps/docs/docs/advanced/environment-variables/index.mdx
+++ b/apps/docs/docs/advanced/environment-variables/index.mdx
@@ -33,6 +33,12 @@ Using the `PUID` and `PGID` will require you to set the correct permissions on t
 | ``NO_EXTERNAL_CONNECTION`` | Disables some requests that need internet connection | `true` or `false` | `false` |
 | ``ENABLE_DNS_CACHING`` | Enables dns caching. Enabled by default, but if you experience issues with IPv6 or static IPs, see [#4006](https://github.com/homarr-labs/homarr/issues/4006) or disable this flag | `true` or `false` | `true` |
 
+## Widgets
+
+| Environment Variable     | Description | Possible values | Default |
+| ------------------------ | ----------- | --------------- | ------- |
+| ``LIVE_TENNIS_API_KEY`` | API key used by the [Live Tennis](/docs/widgets/tennis) widget. A free key is available at [livetennisapi.com](https://livetennisapi.com/subscribe/free) | Any valid Live Tennis API key | *empty* |
+
 
 
 ## Authentication
--- a/apps/docs/docs/widgets/tennis/index.mdx
+++ b/apps/docs/docs/widgets/tennis/index.mdx
@@ -1,0 +1,53 @@
+---
+title: "Live Tennis"
+description: "Displays live, upcoming or completed tennis matches."
+hide_title: true
+---
+
+import { AddingWidget } from "@site/src/components/widgets/adding";
+import { WidgetConfig } from "@site/src/components/widgets/configuration";
+import { WidgetHeader } from "@site/src/components/widgets/header";
+import { tennisWidget } from ".";
+
+<WidgetHeader widget={tennisWidget} categories={["Sports"]} />
+
+The Live Tennis widget shows tennis matches on your dashboard, including live set and game scores,
+the player currently serving and the current game points.
+
+### Adding the widget
+
+<AddingWidget />
+
+### Configuration
+
+<WidgetConfig configuration={tennisWidget.configuration} />
+
+### API key
+
+This widget uses the [Live Tennis API](https://livetennisapi.com), which requires an API key.
+A free tier is available without a credit card at https://livetennisapi.com/subscribe/free.
+
+The key is read **server side only** from the `LIVE_TENNIS_API_KEY` environment variable and is
+never sent to the browser:
+
+```yaml
+services:
+  homarr:
+    environment:
+      - LIVE_TENNIS_API_KEY=twjp_your_key_here
+```
+
+:::warning
+Do not put the API key into a widget option. Board items are readable by everyone who can view the
+board, including anonymous visitors on public boards.
+:::
+
+If the variable is not set, or the key is rejected by the API, the widget shows a
+"No valid Live Tennis API key configured" message instead of the scoreboard.
+
+### Data source
+
+Match data is fetched server side from `https://api.livetennisapi.com`.
+If you need to whitelist the API, please use the following URL: `https://api.livetennisapi.com`
+
+The widget respects `NO_EXTERNAL_CONNECTION`; when that is enabled no request is made.
--- a/apps/docs/docs/widgets/tennis/index.tsx
+++ b/apps/docs/docs/widgets/tennis/index.tsx
@@ -1,0 +1,50 @@
+import { WidgetDefinition } from "@site/src/types";
+import { IconBallTennis } from "@tabler/icons-react";
+
+export const tennisWidget: WidgetDefinition = {
+  icon: IconBallTennis,
+  name: "Live Tennis",
+  description:
+    "Displays live, upcoming or completed tennis matches from the ATP, WTA, Challenger, ITF and junior tours.",
+  path: "../../widgets/tennis",
+  configuration: {
+    items: [
+      {
+        name: "Tour",
+        description: "Restrict the shown matches to a single tour",
+        values: {
+          type: "select",
+          options: ["All tours", "ATP", "WTA", "Challenger", "ITF", "Juniors"],
+        },
+        defaultValue: "All tours",
+      },
+      {
+        name: "Match status",
+        description: "Whether to show live, upcoming or completed matches",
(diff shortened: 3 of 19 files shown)


bensynapse:
Thanks @coderabbitai — one of these caught a real error in my own PR description, so I want to correct that explicitly rather than quietly fix it.

### Correction: the API key error state did not work

My PR body said a missing or rejected key shows a dedicated configuration message. **That was not true.** The router raised `UNAUTHORIZED` and the definition declared the error entry, but the component destructured only `data`, and react-query defaults `throwOnError` to `false` — this repo sets no global override — so the error never reached `WidgetError`. Users got the generic empty state, and the message documented in `apps/docs/docs/widgets/tennis/index.mdx` was unreachable.

I verified the router's behaviour and reported it as the rendered result. That was my mistake and the review was right to flag it.

Fixed in `fa6e790`, along with a second issue found while fixing it: the query client sets `retry: 3`, so a rejected key was retried three times with backoff before showing anything. A bad key is not transient, so that path no longer retries.

I also added a render test over `WidgetError`. The existing unit tests would have passed against the original bug, which is exactly why they weren't enough. The new test asserts rendered output, and I confirmed it fails when the error entry is removed — in that case the widget claims *"You don't have permission to access this widget"*, which is the wrong diagnosis for a misconfigured key.

### Also addressed
- `matchCount` now requires an integer, in both the router schema and the slider validator.
- `getMatches` now declares `mcp` metadata, matching `patchmon` and `health-monitoring`.

### Two I'd push back on, with evidence

**Juniors is not a dead option.** I probed the API before removing it. `?tour=juniors` resolves correctly and returns real matches — the fixtures endpoint currently returns Grand Slam junior events (e.g. French Open Boys). It reads as empty today only because junior draws run at Grand Slams and none is in progress. Removing it would delete a working filter that repopulates several times a year, so I've kept it.

**"server side" is correct unhyphenated here.** Both uses are adverbial — "read server side only", "fetched server side" — where the unhyphenated form is standard; the hyphen belongs to the attributive use ("a server-side call"). It also matches the existing prose in `apps/docs/docs/management/settings/index.mdx`, and the docs contain no instance of "server-side". Happy to change it if you'd prefer the hyphen for consistency.

Note for maintainers: `pnpm lint` reports 5 pre-existing errors in `@homarr/db` (`collection.ts`, `test/schema.spec.ts`) that are present on a clean checkout of this branch's base and unrelated to this PR.

ajnart:
Yeah @bensynapse this is too much of a claude PR with a super small range of users who will be interested by this since it's so outside of the scope of what homarr usually does. I would be much happier if you spent you Claude credits on helping me with the v2 branch https://github.com/homarr-labs/homarr/pull/6356 in order to bring the custom widgets feature to a state where it can support that live tennis widget you did. 

I genuinely think that this could be great, but not as a "core" widget, as a custom one. On v2 there is a widget store where you can share custom widgets you make, they can all have options, requests, actions, ...

If you could instead try to prompt Claude to do "Checkout v2, understand the PR and the changes it brings and setup the homarr mcp, then iterate with it until you can make an exact replica of https://github.com/homarr-labs/homarr/pull/6442 inside of the custom widgets feature. If not possible, describe all the necessary steps and constraints we must overcome to make it so" 

This would help me and the community much more, because then other people than you will be able to make use of the new custom widgets directly in one-shot without Claude code ;)

bensynapse:
@ajnart took you up on this — checked out \`v2\` and worked through the custom-widget path (schema, the source/request executor, the MCP + skill).

Short version: it's not just possible, it's a better home for this than core was — and specifically it fixes the thing that actually sank the core version. The blocker on #6442 was the API key: on a public board a key in a widget option is readable by anyone through the public \`getBoardByName\`, which is why I was pushed into a server-side \`LIVE_TENNIS_API_KEY\` env var. The v2 \`apiKeyHeader\` source auth solves that natively — the user enters their key at install, it's stored encrypted and injected server-side, and the JSX template never sees it. The options (tour/status/count/switches) and the 401 "no key configured" branch map straight onto custom-widget primitives.

Honest tradeoffs, since they're the reason I'd argued for core: it loses compile-time types, the unit tests, and i18n (options are single-label). And the per-set score + serving-dot derivation that lived in a server-side TS mapper has to be re-expressed in the JSX expression sandbox — that's the one non-trivial piece.

Happy to author it as a manifest and publish to Workshop so it's the one-shot install you described. Do you want me to go at it against \`v2\` as it stands now, or wait until the branch settles a bit — the docs still flag the format as likely to change, and I'd rather not ship something that breaks under you.

ajnart:
@bensynapse 
> it loses compile-time types, the unit tests, and i18n (options are single-label). And the per-set score + serving-dot derivation that lived in a server-side TS mapper has to be re-expressed in the JSX expression sandbox

These are not relevant in that context, it's not that deep. 

I think you should continue iterating on v2's capabilities until you are convinced that the outcome widget is close to what you initially wanted. Then you can pinpoint implement flaws that would counter against it. 

For that you have to actually use the custom widget and play around with it to spot its flaws, instead of letting Claude answer your GitHub replies

bensynapse:
Closing this one out — you were clear that this belongs in the v2 custom-widget world with real hands-on iteration rather than as a core widget, and that is fair. Thanks for engaging with it seriously and for the pointers on v2; if we build the custom-widget version properly down the line we will do it the way you described. Appreciate the work you all put into homarr.

ajnart:
Thanks for your efforts @bensynapse ! I would love to see this widget in the custom widgets workshop. Let me know in the dedicated issue for v2 if you're encountering some problems with it

PR #6458, opened 2026-07-24
feat(boards): allow addItem to target a specific section

angusmaul:
## Description

The `board.addItem` endpoint (OpenAPI `POST /api/boards/items` and MCP tool `board_addItem`) always places new items in the first **empty** section of a board. For boards organised into named **category** sections there is no API-reachable way to add an item to a specific category: `addItem` can't target one, and `saveBoard` (which can reposition items) is not exposed via OpenAPI or MCP. The only workaround is editing the database directly.

This PR adds:

- an optional `sectionId` to `addItemToBoardSchema`. When provided, the item is placed at the next free grid position of that section (the existing placement algorithm, unchanged — just pointed at the given section). Only `empty` and `category` sections are allowed; a `dynamic` section id is rejected with `BAD_REQUEST`, and an unknown id with `NOT_FOUND`. When omitted, behaviour is exactly as before (first empty section), so existing OpenAPI/MCP clients are unaffected.
- the sections of each board (`id`, `kind`, `name`, position) in the `getAllBoards` response, so API/MCP clients can discover the `sectionId` to pass without an additional tool (per review feedback, instead of a separate `getSections` procedure).
- tests (default placement, section-targeted placement, unknown section, dynamic section, empty section id, section listing in `getAllBoards`).

Motivation: automation and agent clients (e.g. via the built-in MCP server) managing a categorised dashboard — "add an app tile for service X in the *Local Apps* category".

Verified locally: `pnpm typecheck` (all 37 packages) and the board router test suite (51 tests, including 6 new) pass; changed files formatted with `oxfmt`.

## Checklist

### DIFF
--- a/packages/api/src/router/board.ts
+++ b/packages/api/src/router/board.ts
@@ -35,6 +35,7 @@ import {
   everyoneGroup,
   getPermissionsWithChildren,
   getPermissionsWithParents,
+  sectionKinds,
   widgetDefaultSizes,
   widgetKinds,
 } from "@homarr/definitions";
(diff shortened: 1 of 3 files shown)


packages/api/src/router/board.ts
```diff
@@ -1466,12 +1506,27 @@ export const boardRouter = createTRPCRouter({
...
+              .toSorted((sectionA, sectionB) => (sectionA.yOffset ?? 0) - (sectionB.yOffset ?? 0))
+              .at(0);
+
+      if (!targetSection) {
+        throw new TRPCError(
+          input.sectionId !== undefined
+            ? { code: "NOT_FOUND", message: "Section not found on this board" }
+            : { code: "BAD_REQUEST", message: "Board has no empty section to place items in" },
+        );
+      }
```
ajnart: This code looks really difficult to read with the ternaries, I'd recommend refactoring it

packages/api/src/router/board.ts
```diff
@@ -1427,13 +1428,52 @@ export const boardRouter = createTRPCRouter({
       const oldmarr = oldmarrConfigSchema.parse(JSON.parse(content));
       await importOldmarrAsync(ctx.db, oldmarr, input.configuration);
     }),
+  getSections: protectedProcedure
```
ajnart: I don't see the point of adding a whole procedure for this, instead consider making it a part of the endpoint to get a dashboard ? 

I think overall my MCP implementation is not the best because we have many tools instead of having a few really powerful ones. In that case I believe it's beneficial to merge this behavior into the main dashboard related query

ajnart (changes requested):
Thanks for the PR, I understand its goal I'm just not sure that this is how I would've gone about it. Please check my other comments

angusmaul:
Thanks for the review @ajnart — both points addressed:

- Dropped the getSections procedure; the sections (id, kind, name, position) are now part of the getAllBoards response, so sectionId is discoverable without an extra tool.
- Extracted the section lookup into a findTargetSectionOrThrow helper, so the ternaries are gone.

Locally pnpm typecheck passes across all 37 packages and the board router suite is green (51 tests).

PR #6490, opened 2026-07-30
feat(widgets): persist Beszel and Docker table layouts

ajnart:
## Summary

- match the Beszel and Docker tables to the Downloads widget's draggable, resizable, persistent column UX
- add Docker container action menus on the actions button and row right-click, including logs, lifecycle actions, guarded removal, and Add to Homarr
- preserve existing visibility, filtering, sorting, modal, footer, and edit-mode behavior
- document the interactions and handle malformed saved layouts safely

## Validation

- `pnpm exec vitest run packages/widgets/src/beszel-system-table/component.spec.ts packages/widgets/src/docker/component.spec.ts packages/widgets/src/test/translation.spec.ts --coverage.enabled=false` (558 tests passed)
- `pnpm turbo typecheck --filter=@homarr/widgets --filter=@homarr/docs`
- `pnpm turbo build --filter=@homarr/docs`
- `pnpm --filter @homarr/widgets lint`
- `pnpm --filter @homarr/docs lint`

## Summary by CodeRabbit

- **New Features**
  - Customize visible columns in the Docker containers widget.
  - Drag and resize table columns; layouts are saved automatically per widget.
  - Sort system and Docker tables, with system rows opening detailed statistics.
  - Access Docker actions through row menus or right-click menus.
  - View disk mount details by hovering over system table entries.

- **Improvements**
  - Docker container removal now requires a second confirmation.
  - Tables provide clearer empty states and save-error notifications.
  - Table layouts remain consistent across supported widgets and sessions.

PR #6495, opened 2026-07-31
feat(dev): add Docker development CLI

ajnart:
## Summary

- add a maintained Go developer CLI for local Homarr images, GHCR pull-request images, and running test instances
- support direct, cross-shell CLI usage with `pnpm dev:cli -- ...`, plus optional global installation
- keep development setup explicit: install dependencies, create `.env`, start Redis when needed, run migrations, and expose `pnpm db:seed`
- remove the shell-specific setup script and developer CLI validation from the main CI gate
- document the contributor workflow and optional GitHub CLI/Go requirements

## Validation

- `pnpm typecheck`
- `pnpm build`
- real current-checkout Docker build, launch, readiness probe, provenance discovery, and rebuild
- real PR #6490 temporary checkout/build, provenance discovery, and temporary-directory cleanup
- hosted CI Fast gate and Container/E2E checks on the final branch SHA

The local environment does not provide Go and cannot access the Docker daemon, so this follow-up was checked with static validation; prior CLI tests and the hosted repository checks cover the broader change.

## Checklist

- [x] Targets `dev`
- [x] Conventional commit
- [x] Documentation updated
- [x] No temporary audit images, containers, volumes, or files remain
- [x] AI-assisted code was manually reviewed and exercised end to end


ajnart:
Addressed all CodeRabbit findings in `c8485ed7f` and `ae3e9f779`, including the inline functional issues, the 13 nitpicks, removal of terminal-specific/dead code, and the follow-up encryption-key finding. Development encryption keys are now generated per user, stored with `0600` permissions in the OS config directory, reused across launches, and overrideable with `HOMARR_DEV_SECRET_ENCRYPTION_KEY`. Validation includes `go test -race ./...`, `go vet ./...`, Linux/Windows builds, shell syntax, documentation build, typecheck, production build, and live Docker workflows.

ajnart:
Closed the follow-up first-run race in `b733ed6a7`: key creation now uses an atomic lock directory and same-directory temporary-file rename, so concurrent CLI processes all reuse one complete key. Added a 20-caller concurrency regression test and reran race tests, vet, and Linux/Windows builds.

ajnart:
Improved remote PR reliability and local-build workflow in `c8c5dbc50` through `667bf6239`. The observed `exit status 1` was not rate limiting (the account had 4,345/5,000 GraphQL points remaining and the live query passed at a cost of 3); GitHub stderr was being discarded. Errors now include the actual `gh` message, PR metadata is cached for five minutes with stale fallback, bot toggles/post-build reloads reuse the cache, GHCR image states are reused unless `r` forces a refresh, and failed refreshes retain image states. `R` now builds a selected remote PR locally as `homarr:pr-<number>` from a validated depth-one temporary checkout.

ajnart:
Simplified the primary PR workflow in `d096aecd5` through `ca7e50fb9`: remote rows now include `headRefOid`; local PR images are folded into the matching PR only when their provenance revision exactly equals that head SHA; Enter prefers the exact local image without pulling; stale builds remain separate; `p` still forces GHCR. Exact local rows skip unnecessary registry probes. Local builds continue in the background while navigation and `d` remain usable. Verified against PR #6502, whose GitHub head and local image label both equal `c61695b142ef9a78f2805b21577031a648c64332`.

PR #6515, opened 2026-08-03
feat(sabnzbd): add archived history options to the Downloads widget

Sobe1ac:
## Summary

Add optional SABnzbd archived-history support to Homarr's existing Downloads widget using the widget options architecture.

This change is implemented entirely in Homarr. It does not modify SABnzbd, its queue, Archive behavior, or API.

## User-facing behavior

- Adds **Include archived SABnzbd history** to the Downloads widget when SABnzbd is selected
- Adds an **Archived history window** as a numeric widget option
- Uses positive whole-day values with 1-day increments
- Defaults the history window to **7 days**
- Archived history is **disabled by default**
- The history-window option appears only when archived history is enabled
- Existing **Limit items per integration** still controls how many rows are displayed
- Uses Homarr's existing widget option and translation systems

## Implementation

- Threads the widget options through the existing Downloads request path
- Fetches SABnzbd archived history using `archive`, `start`, and `limit`
- Uses bounded pagination with 100-item pages and a hard 100-page safety cap
- Stops on partial pages or once a full page is entirely older than the selected window
- Merges normal and archived history
- Filters to the selected date window
- Sorts newest-first
- Deduplicates by `nzo_id`

No integration-options persistence layer, database schema changes, migrations, onboarding changes, or integration management-form changes are added.

## Validation

The earlier widget-options candidate passed its focused archived-history tests, affected-workspace typechecks, lint, production Docker build, and live SABnzbd validation.

The current review-response head changes the history-window control from fixed string presets to Homarr's existing numeric option, defaults it to 7 days, and simplifies the archived-history conditional. The guarded compare against the reviewed parent shows only the intended Downloads/SABnzbd code, tests, API/request typing, and documentation files.

Hosted GitHub Actions for the current fork-based head still require maintainer approval before they can run, so fresh CI for this review-response commit is pending.

## Notes

This redesign follows maintainer feedback to keep the feature in the Downloads widget and reuse Homarr's existing option, translation, request-handler, and integration patterns.

### DIFF
--- a/packages/integrations/src/download-client/sabnzbd/sabnzbd-integration.ts
+++ b/packages/integrations/src/download-client/sabnzbd/sabnzbd-integration.ts
@@ -62,7 +62,7 @@ export class SabnzbdIntegration extends Integration implements IDownloadClientIn
     const items = queue.slots
       .map((slot): DownloadClientItem => {
         const state = SabnzbdIntegration.getUsenetQueueState(slot.status);
-        const times = slot.timeleft.split(":").reverse();
+        const times = slot.timeleft.split(":").toReversed();
         const time = dayjs
           .duration({
             seconds: Number(times[0] ?? 0),
--- a/packages/integrations/src/download-client/sabnzbd/sabnzbd-integration.ts
+++ b/packages/integrations/src/download-client/sabnzbd/sabnzbd-integration.ts
@@ -12,6 +12,7 @@ import type { DownloadClientJobsAndStatus } from "../../interfaces/downloads/dow
 import type { IDownloadClientIntegration } from "../../interfaces/downloads/download-client-integration";
 import type { DownloadClientItem } from "../../interfaces/downloads/download-client-items";
 import type { DownloadClientStatus } from "../../interfaces/downloads/download-client-status";
+import { getSabnzbdHistorySlotsAsync } from "./sabnzbd-history";
 import { historySchema, queueSchema } from "./sabnzbd-schema";
 
 dayjs.extend(duration);
@@ -23,7 +24,11 @@ export class SabnzbdIntegration extends Integration implements IDownloadClientIn
     return { success: true };
   }
 
-  public async getClientJobsAndStatusAsync(input: { limit: number }): Promise<DownloadClientJobsAndStatus> {
+  public async getClientJobsAndStatusAsync(input: {
+    limit: number;
+    includeArchivedHistory?: boolean;
+    historyWindowDays?: "10" | "20" | "30";
+  }): Promise<DownloadClientJobsAndStatus> {
     const type = "usenet";
     const { queue } = await queueSchema.parseAsync(
       await this.sabNzbApiCallAsync("queue", { limit: input.limit.toString() }),
@@ -31,6 +36,24 @@ export class SabnzbdIntegration extends Integration implements IDownloadClientIn
     const { history } = await historySchema.parseAsync(
       await this.sabNzbApiCallAsync("history", { limit: input.limit.toString() }),
     );
+    const historySlots = input.includeArchivedHistory
+      ? await getSabnzbdHistorySlotsAsync({
+          activeSlots: history.slots,
+          historyWindowDays: Number(input.historyWindowDays ?? "10"),
+          fetchPageAsync: async ({ start, limit }) => {
+            const { history: archivedHistory } = await historySchema.parseAsync(
+              await this.sabNzbApiCallAsync("history", {
+                archive: "1",
+                start: start.toString(),
+                limit: limit.toString(),
+              }),
+            );
+
+            return archivedHistory.slots;
+          },
+        })
+      : history.slots;
+
     const status: DownloadClientStatus = {
       paused: queue.paused,
       rates: { down: Math.floor(Number(queue.kbpersec) * 1024) }, //Actually rounded kiBps ()
@@ -63,7 +86,7 @@ export class SabnzbdIntegration extends Integration implements IDownloadClientIn
         };
       })
       .concat(
-        history.slots.map((slot, index): DownloadClientItem => {
+        historySlots.map((slot, index): DownloadClientItem => {
           const state = SabnzbdIntegration.getUsenetHistoryState(slot.status);
           return {
             type,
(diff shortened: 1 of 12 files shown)


ajnart (changes requested):
Sorry @Sobe1ac, I thought I'd posted my review.

Overall I'm not against this change, but it reads as over-engineered for a single display toggle. "Include archived history" + a 10/20/30-day window is a **widget option**, not an integration option

- we already have `optionsBuilder` in [packages/widgets/src/downloads/index.ts](https://github.com/homarr-labs/homarr/blob/dev/packages/widgets/src/downloads/index.ts) (same as `showCompletedUsenet` / `limitPerIntegration`), which persists per-widget with no schema change. (So that's already removing the migrations) 

- The `options` DB column + migrations, the schema registry in `definitions`, the serialize/parse helpers in `api`, and the router/middleware/onboarding/request-handler plumbing. Keep sabnzbd-history.ts (the archive fetch + merge/filter/dedupe) and the `history?archive=1` call in [sabnzbd-integration.ts](https://github.com/homarr-labs/homarr/blob/dev/packages/integrations/src/download-client/sabnzbd/sabnzbd-integration.ts), trim the pagination to a simple bounded loop

Then expose the two settings as widget options in `downloads/index.ts` and thread them through `getJobsAndStatuses` like `limitPerIntegration` already is. That's ~100–150 LOC instead of 8.5k

Sobe1ac:
Thank you @ajnart  for taking the time to review this and for the clarification. Your explanation makes sense, and I appreciate you pointing me toward the existing widget options pattern.

My original thinking was to persist these settings at the integration level, but I understand and agree that they fit better as Downloads widget options.

I’ll rework the implementation to align with the existing architecture:

* Move Include archived history and the 10/20/30-day window into the Downloads widget options
* Reuse optionsBuilder in packages/widgets/src/downloads/index.ts, following the same pattern as showCompletedUsenet and limitPerIntegration
* Remove the integration options DB column, migrations, schema registry, API serialization/parsing helpers, and the related router/middleware/onboarding/request-handler plumbing
* Keep sabnzbd-history.ts, including the archive fetch, merge/filter/dedupe behavior, and the history?archive=1 call in sabnzbd-integration.ts
* Simplify the archive pagination to a bounded loop
* Thread the two widget settings through getJobsAndStatuses in the same way as limitPerIntegration

I’ll also reduce the overall scope so the revised change is much smaller and more focused.

Thanks again for the review and for pointing me in the right direction. I appreciate the guidance.

ajnart:
No problem, if you're using AI to code and reply feel free to tell me. In general I think you should tell LLMs to leverage existing systems as much as possible, for common packages, translations, tools, read the docs on the libraries we use, ....

Sobe1ac:
Reopening to address the requested changes.

Sobe1ac:
@coderabbitai review

Sobe1ac:
@coderabbitai review

Sobe1ac:
@ajnart Thanks again for the guidance. The PR has been reworked to use the existing Downloads widget options architecture, and the unrelated formatting change has been removed. When you have a chance, could you please re-review and approve the fork-origin Actions so CI can run?

packages/integrations/src/download-client/sabnzbd/sabnzbd-integration.ts
```diff
@@ -23,14 +24,36 @@ export class SabnzbdIntegration extends Integration implements IDownloadClientIn
...
+                start: start.toString(),
+                limit: limit.toString(),
+              }),
+            );
+
+            return archivedHistory.slots;
+          },
+        })
+      : history.slots;
+
```
ajnart: Nitpick but I think this ternary is confusing

packages/translation/src/lang/en.json
```diff
@@ -3126,6 +3126,19 @@
...
+        },
+        "historyWindowDays": {
+          "label": "Archived history window",
+          "description": "Show archived jobs completed within this period. Increase the item limit to show older results.",
+          "values": {
+            "10Days": "10 days",
+            "20Days": "20 days",
+            "30Days": "30 days"
+          }
+        },
```
ajnart: Don't need 3 keys for the same translation, it's basically "{n} days"

packages/widgets/src/downloads/index.ts
```diff
@@ -73,6 +73,28 @@ export const { definition, componentLoader } = createWidgetDefinition("downloads
...
+              value: "20",
+              label: (t) => t("widget.downloads.option.historyWindowDays.values.20Days"),
+            },
+            {
+              value: "30",
+              label: (t) => t("widget.downloads.option.historyWindowDays.values.30Days"),
+            },
+          ],
+          withDescription: true,
+        }),
```
ajnart: I might be wrong but does it need to be a string, couldn't it be a number ? 

Disregard this if it can't

packages/widgets/src/downloads/index.ts
```diff
@@ -73,6 +73,28 @@ export const { definition, componentLoader } = createWidgetDefinition("downloads
...
+            {
+              value: "20",
+              label: (t) => t("widget.downloads.option.historyWindowDays.values.20Days"),
+            },
+            {
+              value: "30",
+              label: (t) => t("widget.downloads.option.historyWindowDays.values.30Days"),
+            },
+          ],
+          withDescription: true,
```
ajnart: Since you're just passing it to the API it doesn't have to be only these values, people will often reason in terms of weeks I think. (7d, 2wk, 4wk) 

You could make the option a slider for easier use, or we should really have a number input option type if we don't already do.

ajnart (commented):
Much better thank you for the changes, @Sobe1ac i still need to test it but I left some comments about the code, feel free to fix them or not if you think it's not necessary

Sobe1ac:
Thanks @ajnart, I went through the comments and updated this.

- Simplified the archived-history conditional.
- Changed `historyWindowDays` to use the existing numeric option instead of fixed string values.
- Default is now 7 days with 1-day increments, so values like 2, 7, 14, 28, etc. work naturally.
- Removed the obsolete 10/20/30 translation values.

The existing archive pagination safety cap is unchanged.

The latest fork-origin Actions are waiting for approval before CI can run. When you get a chance, could you approve those runs so the current head can be validated? Thanks again for the feedback.

Sobe1ac:
Hi @ajnart — just checking back when you have a chance. The PR is still ready for testing/review, and the fork-origin Actions are still waiting for maintainer approval to run. No rush, and thanks again for the earlier feedback. -- Ahh I just caught up on Discord.. your on Vacation - Enjoy!!

PR #6530, opened 2026-08-07
fix(assistant): mysql message storage, migration metadata and per-page bundle cost

ajnart:
## Summary

Follow-up fixes for #6482. This targets `feat/homarr-assistant` rather than `dev` because it corrects migrations and code that only exist on that branch.

Two of these are functional bugs on MySQL, one is a cross-database ordering bug, and one removes the assistant's cost from every page.

## 1. MySQL cannot store assistant messages with attachments

`assistant_message.content` was `text`, which MySQL caps at 65,535 bytes. The chat route accepts a single image attachment up to 1,000,000 decoded bytes, base64-inlined into the message part, so appending one fails under the default strict mode. SQLite and PostgreSQL are unaffected.

Verified against a real MySQL 8 container:

```
MYSQL sql_mode: ...,STRICT_TRANS_TABLES,...
MYSQL RESULT: FAILED -> ER_DATA_TOO_LONG Data too long for column 'content' at row 1
PG RESULT: inserted, stored length = 200000
```

Now `mediumtext` (16MB), above the route's 2MB request cap. MySQL only; the other two dialects are untouched.

The migration keeps the parenthesised `DEFAULT ('...')` form that every other text default in this folder uses — MySQL rejects a literal default on TEXT columns, and `drizzle-kit` emits the literal form.

## 2. `drizzle-kit generate` has been broken for MySQL since July

`0042_snapshot.json` spells one table's key `checkConstraints` instead of `checkConstraint`. drizzle-kit rejects the folder over it and **exits 0** with only a printed message, so `migration:mysql:generate` has silently produced nothing since #6192 (2026-07-04).

Every MySQL migration authored since was therefore hand-written with no snapshot — including 0044 and 0045 from #6482. Without those, the next generate diffs against 0043 and re-emits the assistant tables.

This is also the class of bug behind the reports of MySQL instances breaking on upgrade: **v1.70.0 shipped `0042_add_widget_secret.sql` on disk but never registered it in `_journal.json`**, so the migrator skipped it and upgraded MySQL instances had no `widget_secret` table.

Adds the two missing snapshots plus `migration-journal.spec.ts`, which asserts for all three dialects: journal ↔ `.sql` agreement both ways, a snapshot per entry, gap-free indexes, and consistent snapshot key spelling. Confirmed it fails on both historical defects.

After this, all three dialects report `No schema changes, nothing to migrate`.

## 3. Message ordering is unstable on MySQL and SQLite

`getThread` ordered purely by `created_at`, which is second-granular on MySQL and SQLite (PostgreSQL has microseconds). A user turn and the assistant reply routinely land in the same second, so their order was whatever the engine returned — and the client rebuilds the tree from `parent_id`, so a child ahead of its parent corrupts the thread.

Adds an id tiebreaker for determinism plus a stable topological pass guaranteeing parents precede children. Dangling parents become roots and cycles are appended, so no message is dropped.

## 4. The assistant shipped on every page, including login

`AssistantProvider` was statically imported by the root layout, so assistant-ui, Lexical, the Markdown renderer and Mermaid were downloaded and parsed on every page. `getAvailability` only skipped *rendering*; the disabled branch still imported the board widget, which pulls the whole conversation surface. Instances with the assistant **off** paid full price.

Measured across production builds:

| Page | Before | After |
|---|---|---|
| `/auth/login` | 42 chunks, 3.81 MB, **820 KB assistant-ui chunk** | 36 chunks, 2.40 MB, **none** |
| board page | 55 chunks, 5.35 MB, **820 KB assistant-ui chunk** | 56 chunks, 4.56 MB, **none** |
| `/manage/about` | — | 39 chunks, 2.48 MB, **none** |

The root layout resolves availability on the server through the existing `api.assistant.getAvailability()` — the same way it already calls `api.user.getById()` — and hands the boolean to a light `AssistantGate`. That server-side answer is what makes it work: `next/dynamic` never references the heavy chunk when the assistant is off, while enabled instances still server-render it, so there is no loading flash and no remount.

No new API surface is added. An earlier revision of this PR introduced a `getFeatureFlagsAsync` helper and a `/api/features` endpoint; both were dropped because they duplicated the rule `getAvailability` already owns. A general feature-flag surface is still worth having, but it belongs at boot time across every optional feature (backups on postgres, docker without a socket, analytics, the assistant) rather than as a per-request database read introduced here.

## Validation

- Real browser against the production build, two seeded databases: enabled → runtime chunks load after login and the assistant works; disabled → `ASSISTANT RUNTIME CHUNKS LOADED: NONE` on the login page and after login
- MySQL migrate + **seed** on a fresh install, and upgrades from **v1.60.0, v1.65.0, v1.70.0, v1.73.0** — all reach 46 migrations, preserve existing rows, and accept a 1.4 MB write
- SQLite and PostgreSQL migrations re-verified unchanged (`No schema changes`)
- All 8 Playwright e2e tests pass in real containers, including `assistant-management` which seeds the assistant *enabled* and so exercises the new gate's enabled path
- Full unit suite: 1870 pass. `pi-hole.spec.ts` is flaky and fails on `dev` too, unrelated to this change
- Typecheck, lint and format clean on every changed file

## Known, not addressed here

`custom_widget_secret`'s foreign key: drizzle derives a 65-character name, one over MySQL's 64-char identifier limit, so `generate` emits an `ALTER` that MySQL rejects with `ERROR 1059` (confirmed against a live server). It is the one remaining spurious MySQL diff. Fixing it properly needs a coordinated rename across all three dialects, which is out of scope here.

## Summary by CodeRabbit

- **New Features**
  - Assistant availability now adapts to instance settings and authentication status.
  - Disabled assistants display localized unavailable states without loading the assistant runtime.
  - Assistant components load more efficiently when needed.
  - Conversation messages now appear in consistent parent-child order.

- **Bug Fixes**
  - Assistant messages support substantially larger content, up to 200,000 characters.
  - Improved handling of missing parents, cycles, and long conversation chains.
  - Preserved message ordering across branching conversations and misplaced replies.


ajnart:
@coderabbitai review

ajnart:
Both findings were real and are fixed in f25be7cf1. I reproduced each against the previous implementation first:

```
cycle case [child->a, a->b, b->a]: child , a , b     <- child ahead of its parent
deep chain 30000: THREW -> RangeError Maximum call stack size exceeded
```

**Recursive traversal.** Replaced with an explicit LIFO stack, pushing siblings in reverse so the incoming order is preserved. Added a 50,000-message chain test.

**Cycle descendants.** The fallback appended unvisited messages without walking their children. Each remaining component is now entered at the earliest member of its cycle in the incoming order and traversed from there, so only the cycle's own back-edge can violate parent-before-child, which is unavoidable by definition. Added the `child -> a, a -> b, b -> a` case asserting `a` precedes `child`.

Both new tests were confirmed to fail against the old implementation and pass against the new one.

ajnart:
@coderabbitai review

ajnart:
@coderabbitai review

ajnart:
Third finding fixed in acc11e8d1 — it was a real UX bug and a good catch.

Collapsing availability and authentication into one boolean meant a signed-out visitor on a **configured** instance was told `unavailable.notConfigured`, which the server cannot actually know for them since `getAvailability` needs a session.

The layout now resolves one of `enabled` / `unconfigured` / `unauthenticated` / `error`, mapping onto the four messages already present in `en.json`. Removing the client-side session check from the branch also eliminates a possible remount once the session settles.

Verified in a browser against production builds with seeded databases:

```
[ENABLED + signed out]  heavy chunks: NONE
[ENABLED + signed in]   heavy chunks: 2dk9chv2qk_gu.js,1gel_77ymk1-w.js | spotlight: "Ask Homarr Assistant"
[DISABLED + signed in]  heavy chunks: NONE | "An administrator must configure and enable the assistant first"
```

Bundle numbers unchanged: login 36 chunks / 2.40MB, board 56 / 4.56MB, about 39 / 2.48MB, no assistant-ui on any.

Also in this round, `drizzle-kit generate` is now clean for **all three** dialects for the first time since 2026-07-04:

```
sqlite:     No schema changes, nothing to migrate
postgresql: No schema changes, nothing to migrate
mysql:      No schema changes, nothing to migrate
```

The remaining MySQL noise was `custom_widget_secret`'s foreign key: drizzle derives a 65-character name, one over MySQL's limit, so every run re-proposed an ALTER that fails with error 1059. The schema now declares the shortened name migration 0040 actually created — verified against a freshly migrated container. Only `schema/mysql.ts` changed; sqlite and postgresql keep their derived names and need no migration. The parity test now matches keys on what they link rather than on their generated name, and a new guard test fails CI if any MySQL identifier exceeds 64 characters.

PR #6538, opened 2026-08-07
fix(docker): Expose TCP port in Dockerfile

svengo:
The tcp port should be exposed in Dockerfile, see [1]. Because the port is now exposed in the docker image, it doesn't need to be specified in the traefik configuration [2].

---
1: https://docs.docker.com/reference/dockerfile/#expose 
2: https://doc.traefik.io/traefik/reference/routing-configuration/other-providers/docker/

<br/>
<div align="center">
  
  <h3>Homarr</h3>
</div>

- [ ] When using AI; No temp files are checked in, the code style follows the rest of the project

- [ ] I confirmed manually that the integration or widget is working with a real system and performed basic smoke
- [ ] Optional but recommended: I confirm that I am willing to be added to the `CODEOWNERS` of this feature and will

## Summary by CodeRabbit

* **Configuration**
  * Updated the container configuration to expose the application on TCP port 7575.
  * Simplified the proxy setup by removing the redundant explicit port label.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -95,5 +95,6 @@ ENV AUTH_PROVIDERS='credentials'
 ENV REDIS_IS_EXTERNAL='false'
 ENV NODE_ENV='production'
 
+EXPOSE 7575/tcp
 ENTRYPOINT [ "/app/entrypoint.sh" ]
 CMD ["sh", "run.sh"]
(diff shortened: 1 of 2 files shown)


ajnart:
What's the purpose of specifically adding /tcp ?

svengo:
> What's the purpose of specifically adding /tcp ?

Thanks for the tip. TCP is the default protocol, so we can omit that.

PR #6546, opened 2026-08-09
feat(api): expand REST coverage for full dashboard automation

### LINKED ISSUE #6435: feat: Expand Homarr API Coverage to Support Full Dashboard Automation
### Summary

Homarr’s current API (https://homarr.dev/docs/management/api/) is too limited to fully automate dashboard creation and configuration. Critical GUI‑only actions—such as creating search engines, adding dashboards, configuring widgets, and managing admin users, not all can be reproduced programmatically. This prevents reliable infrastructure‑as‑code workflows (e.g., Ansible playbooks) and forces users to manually rebuild dashboards through the GUI.

This feature request proposes expanding the API to cover all GUI interactions, enabling complete automation and reproducible deployments.

I know we can make a backup but it contains a sqlite db and a text, nothing that can be use to be easily edited and avoid to do everything in the gui...

###Problem Description

I am building an Ansible playbook to automate Homarr deployment and configuration. The goal is to redeploy or rebuild my dashboard simply by applying a playbook and feeding a text file containing the desired configuration.

However, the current API only supports a small subset of actions. Many essential operations are GUI‑only, including:

    Creating search engines

    Creating additional dashboards

    Adding, removing, or configuring widgets

    Managing dashboard layout

    Managing categories

    Managing services

    Managing admin users

    Exporting/importing dashboard configuration

    Setting global settings (theme, language, etc.)

Additionally, the API currently requires an existing admin user to generate an API key. This creates a bootstrap problem:

    I can automate user creation, but only normal users.

    I cannot automate creation of the initial admin user, because generating the API key requires logging into the GUI as the admin.

    Therefore, full automation is impossible.

### Requested Features

**1. API Endpoint: Create Admin User**

Allow creation of an admin user via API, including:

    Role assignment

    Password definition

    Ability to generate an API key without GUI login

This solves the bootstrap problem and enables fully automated deployments.

**2. API Endpoint: Create Search Engines**

Expose CRUD operations for search engines:

    Create search engine

    Update search engine

    Delete search engine

    List search engines

Fields should match the GUI:

    Name

    URL template

    Icon

    Category

    Default status

**3. API Endpoint: Create & Manage Dashboards**

Expose full dashboard management:

    Create dashboard

    Delete dashboard

    Rename dashboard

    Set default dashboard

    Export dashboard configuration

    Import dashboard configuration

**4. API Endpoint: Manage Widgets**

Expose CRUD operations for widgets:

    Add widget

    Remove widget

    Configure widget settings

    Move widget on the grid

    Resize widget

Widget types should be fully supported:

    Apps

    Search bar

    Weather

    Calendar

    Custom widgets

    Etc.

**5. API Endpoint: Manage Services**

Expose CRUD operations for services:

    Add service

    Remove service

    Configure service fields

    Assign service to dashboard

**6. API Endpoint: Global Settings**

Expose configuration for:

    Theme

    Language

    Layout

    Custom CSS

    Global search settings

    Backup/restore

**7. API Endpoint: Full Configuration Export/Import**

Provide a single JSON/YAML export containing:

    Dashboards

    Widgets

    Search engines

    Services

    Categories

    Global settings

And a matching import endpoint.

This would allow:

    Infrastructure‑as‑code

    Git‑versioned dashboard definitions

    Instant redeployment

Why This Matters

Homarr is positioned as a modern, customizable dashboard. But without a complete API, it cannot be reliably automated or integrated into DevOps workflows.

Adding these endpoints would enable:

    Ansible automation

    Terraform modules (or Open Tofu)

    GitOps workflows

    CI/CD deployment of dashboards

    Backup/restore without GUI

    Multi‑environment reproducibility

This dramatically increases Homarr’s value for power users and self‑hosters.
Proposed Minimal Viable Implementation

If full coverage is too large for one release, the following subset would already unlock automation:

    Create admin user via API

    Generate API key via API

    Create dashboard

    Create search engine

    Export/import dashboard configuration

This alone would allow complete reproducible deployments.

**8.a API Endpoint: Generate a backup**

Homarr’s backup feature is currently GUI‑only and produces a SQLite database file. For automation (e.g., Ansible), an API endpoint is needed to trigger backups without manual interaction.
Requested behavior

Provide an API endpoint that can generate a backup in one or both formats:

    SQLite backup  
    Same as the GUI.
    Endpoint example: POST /backup/sqlite  
    Returns the raw .db file as binary.

    JSON/YAML backup  
    Structured export of all configuration (dashboards, widgets, search engines, services, settings).
    Endpoint example: POST /backup/export  
    Returns a full config object suitable for GitOps and IaC.

Both formats serve different automation needs, and either one would already improve reproducibility.

**8.b API Endpoint: Restore a backup**

To complement the backup endpoint, Homarr should provide a simple restore mechanism through the API. This allows full automation of redeployment without touching the GUI.
Requested behavior

Support restoring from either backup format:

    SQLite restore  
    Endpoint example: POST /restore/sqlite  
    Accepts the same .db file produced by the GUI or /backup/sqlite.

    JSON/YAML restore  
    Endpoint example: POST /restore/import  
    Accepts a structured configuration export and rebuilds dashboards, widgets, search engines, services, and settings.

Both restore methods enable fully automated Homarr provisioning and disaster recovery.

### Conclusion

Homarr’s GUI is excellent, but the lack of API coverage prevents automation. Expanding the API to cover all GUI interactions would make Homarr dramatically more powerful and suitable for modern DevOps workflows.

I would be happy to help test or provide example Ansible playbooks once these endpoints exist.
I had started one but I had to make it much more robust by querying the actual existing dashboard ids, etc... 

### Additional information

_No response_

rfdrew:
Closes #6435.

## Why

Boards could only be created empty through the API. `POST /api/boards/items` took no geometry, so the size always came from `widgetDefaultSizes` and the position was always the next free cell — a dashboard could not be described in code. Search engines, integrations, groups, api keys and global settings were reachable from the interface only.

## Board composition

`POST /api/boards/items` now accepts an **optional** placement. Without it the behaviour is byte for byte what it was before:

```json
{ "boardId": "...", "kind": "clock", "sectionId": "...", "xOffset": 4, "yOffset": 2, "width": 6, "height": 5 }
```

`width`/`height` set the size, `xOffset`/`yOffset` an exact position, `sectionId` the target section. The values apply to every layout and the width is clamped to narrow breakpoints; a `layouts` array gives per breakpoint control instead. Overlaps are rejected with `400`.

Added around it:

| | |
| --- | --- |
| Items | `GET /api/boards/{id}/items`, `PATCH`/`DELETE /api/boards/{boardId}/items/{itemId}` |
| Sections | `GET`/`POST /api/boards/{id}/sections`, `PATCH`/`DELETE .../sections/{sectionId}` |
| Layouts | `GET`/`PUT /api/boards/{id}/layouts` |
| Board | `GET /api/boards/{id}` returns layouts, sections and items at once |
| Access | `GET /api/boards/{id}/permissions`, `PUT .../permissions/users`, `PUT .../permissions/groups` |
| Portability | `GET /api/boards/{id}/export` ⇄ `POST /api/boards/import` with `onConflict: fail｜skip｜replace` |

Ids inside a board document are local references and are remapped on import, so the same document can be applied repeatedly and to other instances.

## Remaining services

Search engines, integrations, groups, api keys and server settings are exposed over OpenAPI and registered as MCP tools. `GET /api/config/export` ⇄ `POST /api/config/import` cover a whole instance.

The configuration document keeps ids as they are — widget options reference other entities by id inside an opaque object, so remapping would silently break boards. Entities are instead matched by natural key when the id is unknown, which is what makes a document from a separately installed instance applicable at all: every instance seeds its own `everyone` group with a random id.

It deliberately contains no users, no group memberships, no secret values, no custom widgets and no media — that is documented, and the backup endpoints remain the way to make a complete copy. Those now accept an `ApiKey` header so a recovery job does not need a browser session.

## Headless bootstrap

`homarr api-keys create -u admin` prints a key without the web interface. Together with the existing `recreate-admin` that removes the last manual step from a first time setup. `api-keys list` and `delete` are there too.

## Safety

- **No database migration.** The schema already stores position and size in `item_layout`.
- **Imports are atomic.** Statements are collected and validated first, then applied in one transaction, so a document with a broken reference never leaves a board half written. `replace` never deletes the board row, so home board settings and per user permissions pointing at it survive.
- **Bounded work.** Grid coordinates and sizes in a document are capped, and the placement scan only tests slot edges and never looks past the last occupied row, so its cost does not depend on the column count.
- **Backwards compatible.** Every new field is optional; existing endpoints keep their behaviour.

## Testing

`pnpm turbo typecheck`, `pnpm lint` and `pnpm format` are clean, docs build passes. 377 tests in the touched packages, including:

- placement: explicit size and position, collisions, clamping to narrow breakpoints, dynamic sections as sub grids, regression on the previous automatic placement
- import/export round trips, `skip`/`replace` idempotency, rollback on an invalid document, cross instance adoption
- authorization: anonymous and unprivileged sessions against every new endpoint, asserting `404` rather than `403` where board existence must not leak

The ten failing files in a full run (`packages/db/test/*-migration`, `packages/integrations/test/*`) are testcontainers based and fail identically on a clean checkout here.

## Notes for review

- Documentation lives in `apps/docs` in this repo, so it is updated in tree rather than in `homarr-labs/documentation`. Happy to move it if that is the wrong place.
- `POST /api/boards/import` keeps the strict board name rule because it creates a board like the interface does; `config/import` accepts any name because it restores. Both are documented.
- The grid cap is 256. It covers oldmarr boards with room to spare, but if wider boards exist in the wild the number should be revisited.

---

- [x] Documentation is up to date

## Summary by CodeRabbit

- **New Features**
  - Added CLI commands to list, create, and delete API keys, displaying new keys once.
  - Added board import/export, placement validation, permissions management, and configuration import/export with conflict handling.
  - Expanded API access for configuration, groups, integrations, search engines, settings, and API keys.
- **Bug Fixes**
  - Improved search engine form handling and validation.
  - Backup export and import now support API-key authentication for automation.
- **Documentation**
  - Added guidance for automated backups and CLI-based API key management.
  - Documented board automation, configuration transfer, service endpoints, and validation behavior.


ajnart:
Thank you for the follow up to your issue @rfdrew, would you mind fixing the comments by coderabbit ? I'll review them. Also, this PR's commits are made by another github user, i think there's an issue with your git config.

ajnart:
@codex review, take into account the coderabbit comments to not say the same thing

apps/docs/docs/advanced/command-line/api-keys/index.mdx
ajnart, line 1: I think it's a good idea to have it like that. So your script will create the container then create the users within that container with the homarr cli, then use the API endpoints directly, correct ?

packages/api/src/open-api.ts
ajnart, line 1: Good reflex to version bump

ajnart (approved):
Reviewed the full automation surface against #6435 and pushed fixes through 6535af85c.

Why these changes were needed:
- multi-table board mutations could partially persist after a later failure
- section removal/cycles and fallback placement could corrupt responsive layouts or reject valid category-only boards
- board import skip could expose a private board ID
- generic settings/config imports accepted malformed data and could bypass public-home-board invariants
- search-engine URL validation accepted non-HTTP schemes
- API-key IDs were only 32-bit and integration creation did not return its ID

How it was hardened:
- atomic DB operation batches for board item/section writes
- per-layout cycle and placement validation before writes
- strict per-group server-setting schemas plus final merged/remapped import validation
- canonical createId() identifiers and explicit automation response contracts
- regression coverage for partial writes, responsive section removal, cycles, private-board access/settings, secret export, and URL schemes

Validation:
- 1,832 tests passed across 144 files (12 skipped)
- all 37 typecheck tasks passed
- repository lint passed with zero errors
- all changed files pass oxfmt
- docs production build passed
- exact-head CodeRabbit and semantic-title checks passed

No remaining material blocker found. Approved; not merged.

apps/docs/docs/advanced/command-line/api-keys/index.mdx
rfdrew: Yes, exactly that flow. Concretely:

```bash
# 1. container is up
docker compose up -d homarr

# 2. create the credentials admin — only needed when none exists yet.
#    Prints a generated password.
docker exec -it <container-id> homarr recreate-admin -u admin

# 3. mint a key for that user. Printed once, in "<id>.<token>" form.
docker exec -it <container-id> homarr api-keys create -u admin

# 4. everything from here is REST
curl -H 'ApiKey: <id>.<token>' -H 'Content-Type: application/json' \
  -d '{"name":"homelab","columnCount":12,"isPublic":true}' \
  http://localhost:7575/api/boards
```

So the CLI is only involved in getting the first credential — steps 2 and 3. Boards, widgets, sections, layouts, integrations, search engines, groups and settings all go over the API afterwards, and `POST /api/config/import` can do the whole instance in one call. No browser at any point.

One thing worth knowing: the key inherits the permissions of the user it was created for, so for unattended provisioning it should belong to an admin.

ajnart:
@rfdrew I haven't had the time to review this yet, it might be better to leave it out before v2 is released, or to target the release/v2 branch as merge target to be honest, there are a bunch of changes there and it might be better to include these new endpoints into the mcp/assistant directly. 
What do you think ?

PR #6561, opened 2026-08-11
feat(integrations): add Komodo integration

Sveeeeeen:
## Summary

- add a native, read-only Komodo integration using API key and API secret authentication
- extend the existing **Containers** widget (`dockerContainers`) with Komodo as an optional data source instead of adding new widgets, as suggested in review
- show Komodo containers in the existing table, with a compact summary of server, stack, deployment, and problem counts above it
- add a **Servers** view to the same widget with CPU, memory, disk, load average, network throughput, core count, and Periphery version
- make the displayed sections and the 1-300 second refresh interval configurable
- document the integration and the extended widget

Leaving the integration selection empty keeps the widget on Homarr's direct Docker connection, unchanged.

## Implementation details

- follows the existing integration → request handler → tRPC router → widget layering
- reads containers via `/read/ListAllContainers`, falling back to `/read/ListAllDockerContainers` on older Komodo versions
- maps Komodo container states onto Homarr's Docker states and falls back to `unknown` for future states
- maps Komodo resource states to healthy, warning, error, and unknown without failing on future states
- excludes templates and keeps malformed individual resources visible as unknown where possible
- normalizes Komodo network deltas to bytes per second using the reported Periphery polling rate
- shares the container icon lookup between the Docker and Komodo paths
- performs only authenticated read requests and does not expose credentials to the client

## Behaviour changes

These are user-visible and affect existing boards:

- the widget is renamed from **Docker stats** to **Containers**, since it is no longer Docker-only. Non-English names follow via Crowdin.
- Docker actions (start, stop, restart, remove, logs, add to Homarr) are hidden when a Komodo integration is selected — the actions column is filtered out and the row context menu is disabled. Komodo is read-only; Homarr never mutates Komodo resources.
- query errors are now surfaced through the widget error boundary instead of being swallowed. On a public board, a viewer without admin rights now sees the "no permission" card rather than an empty table, because `docker.getContainers` requires admin.
- the Komodo-specific switches are marked `skipContextMenu`, so the widget right-click menu is not flooded with options that do not apply to a plain Docker widget.

## Validation

- 935 Vitest tests passed across the affected packages
- `turbo run lint typecheck build` passed across the monorepo, including the Next.js production build
- oxfmt and git diff checks passed
- Docusaurus production build passed
- manually tested with a fork-built Docker image against a live Komodo installation: container table, server table and card layouts, summary at one, two and four columns, running and non-running container states, long-name truncation, and the widget context menu

## Summary by CodeRabbit

- **New Features**
  - Added Komodo integration support with secure API credentials and connection testing.
  - Expanded the Containers widget with read-only Komodo container, server, resource, problem, and metrics views.
  - Added configurable refresh intervals, summaries, server columns, and responsive layouts.
  - Docker actions remain available only for direct Docker data sources.

- **Documentation**
  - Added Komodo setup documentation and updated Containers widget guidance.

- **Bug Fixes**
  - Improved container icon handling and fallback behavior for unavailable data.


ajnart:
I believe you could instead convert this to be added to the container / docker widget to be more general I think it could be made tu support Komodo without adding a new widget, same thing the server numbers, deployments, problems, stacks running are basically "small numbers" info that could be shown inline inside of the docker widget / another table in the same way. 

What do you think ? This way we don't add a new widget but we extend the features of the existing one

Sveeeeeen:
> I believe you could instead convert this to be added to the container / docker widget to be more general I think it could be made tu support Komodo without adding a new widget, same thing the server numbers, deployments, problems, stacks running are basically "small numbers" info that could be shown inline inside of the docker widget / another table in the same way.
> 
> What do you think ? This way we don't add a new widget but we extend the features of the existing one

Yeah, that makes sense to me. My thought would be to let the Docker widget use Komodo as an optional data source. We could show the Komodo containers in the existing table, add the small overview numbers above it, and make the server overview another selectable view or table. The existing direct Docker support would continue to work as it does now.
Is that roughly what you had in mind, or would you prefer the Komodo integration to only add the overview information without listing its containers?

Sveeeeeen:
> I believe you could instead convert this to be added to the container / docker widget to be more general I think it could be made tu support Komodo without adding a new widget, same thing the server numbers, deployments, problems, stacks running are basically "small numbers" info that could be shown inline inside of the docker widget / another table in the same way.
> 
> What do you think ? This way we don't add a new widget but we extend the features of the existing one

Done

PR #6562, opened 2026-08-11
feat(app-form): suggest the icon of the app website in the image picker

junkerderprovinz:
## What this adds

When an app has a URL, the image picker now offers the icon of that website as its first suggestion.

While the picker dropdown is open, the server is asked for the icon of the app URL. If one is found it appears in an "Icon of the website" section above the library results and can be selected with the mouse or the keyboard like any other image. It is a suggestion only, nothing is written into the field for you.

## How it works

`fetchFaviconUrlAsync` in `@homarr/icons` requests the page the app URL points to, reads its `<link rel="icon">` and `<link rel="apple-touch-icon">` declarations and picks the best one. Apple touch icons rank first because they are meant as app artwork, and the declared `sizes` attribute breaks ties within the same relation. Relative hrefs are resolved against the URL that actually answered, so a redirect to a login page still resolves correctly. When the page declares nothing usable, `/favicon.ico` of the origin is probed and only accepted when the response really is an image, because many self hosted apps answer unknown paths with their HTML entry point.

Details:

- The requests go through `fetchWithTrustedCertificatesAsync`, so self signed certificates of local apps are handled the same way as everywhere else in Homarr.
- Timeout of 5 seconds per request, at most 256 KiB of HTML is read, and reading stops at `</head>`.
- Only `http` and `https` are accepted, validated in the input schema and again in the fetcher. Local addresses are allowed on purpose, self hosted apps usually live on the same network as Homarr.
- The procedure requires the `app-create` permission, which every path that opens the app form already requires, and it returns `null` when `NO_EXTERNAL_CONNECTION` is set, like the other outbound procedures.
- Detection only runs while the dropdown is open, so typing a URL does not cause a request per keystroke.
- Everything fails quietly to `null`, an unreachable app simply produces no suggestion.

## Tests

- `packages/icons/src/test/favicon-fetcher.spec.ts` covers relation ranking, size ranking, resolution against the redirect target, the well known fallback, the HTML entry point false positive and rejected protocols.
- `icon-picker.spec.ts` covers that the suggestion is offered and selectable, and that detection stays disabled without a website or while the dropdown is closed.

## Context

This is a focused re-cut of the favicon detection half of #6248 against the picker from #6556. The pagination half of #6248 is obsolete since the rewrite covers it, and the old branch no longer applies. The detection is rebuilt for the new picker rather than ported: it hooks into the picker sections instead of writing the field, and it uses the trusted certificate fetch helper. #6248 can be closed.

---

- [x] Builds without warnings or errors (`tsc --noEmit`, `oxlint` and `oxfmt` are clean for the affected packages, unit tests pass)
- [x] Documentation is up to date (`apps/docs` is updated in this pull request)

## Summary by CodeRabbit

* **New Features**
  * Added website icon suggestions to the app icon picker when a valid website URL is provided.
  * Suggestions prioritize declared site icons and fall back to the website’s standard favicon.
  * Direct image URLs remain unchanged and are not searched automatically.
  * Detection runs only while the picker is open and never replaces a manually selected image.
* **Bug Fixes**
  * Improved handling of unreachable websites, redirects, unsupported URLs, missing icons, and invalid image responses.
* **Documentation**
  * Updated guidance on website icon suggestions and image priority.


ajnart:
Hello, thank you for the PR, I understand what you want to do.
Why not just add a little "ActionButton" with the URL + /favicon that the use can click to change the icon to that one instead?

junkerderprovinz:
Hey @ajnart, thanks for taking a look!

The reason I went with an automatic suggestion instead of a button: it only fetches while the picker dropdown is actually open, which you've already triggered on purpose by clicking to change the icon, so there's no "wasted request per keystroke" problem a button would need to solve. And since it just adds the website's icon as one more option in the same suggestion list the picker already shows (library results etc.), rather than a new control next to the URL field, it stays inside the existing picker UX instead of adding a second way to do a similar thing. Nothing gets written into the field automatically either way, it's purely an extra choice in the list.

That said, if you'd still rather have it be explicit (a button the user clicks on demand instead of it appearing automatically when the picker opens), I'm happy to rework it that way, just let me know which you'd prefer and I'll adjust.

ajnart:
Okay I got it. I think your approach is better, I'm just unsure of when it should be injected because if it is automatic since the form has the icon input before the URL, the user might check if an icon exists (url still sits at default value) -> find one on the web if not -> fill the url. 
In that case of linear form-filling the feature doesn't get used that's what I'm concerned about and why I suggested a specific visible element

What do you think?

junkerderprovinz:
That's a fair concern and I checked the actual form: you're right, IconPicker sits before the href field (packages/forms-collection/src/new-app/_form.tsx), so a top-to-bottom fill really does hit the icon step before there's a URL to suggest from.

Simplest fix that keeps the auto-suggest approach: swap the order so href comes before IconPicker. Then the natural top-to-bottom flow already has a URL by the time someone reaches the icon field, and the suggestion is there without needing a separate control. I don't see a reason for the current name → icon → url order (icon feels more like "the last thing you fine-tune" than a required id field), but it's your form, so wanted to check before just reordering it: good with that, or would you rather keep the current order and I add a visible nudge on the icon field once a URL exists instead?

junkerderprovinz:
Pushed 152f5b3 for the redirect finding: the `/favicon.ico` fallback now uses the origin that actually answered the request instead of the requested one, so a cross-origin redirect resolves against the host that served the page. Added a regression test covering that case.

junkerderprovinz:
Good catch, thank you. Pushed a fix: `fetchPageAsync` now returns the response's resolved url even when there's no usable html to read (non-ok status, non-html content type, or no body), instead of discarding it along with the html. The well known favicon fallback picks that up automatically since it already reads `page.url`. Added a test for the specific case you flagged (redirect target answers with something other than html).

ajnart:
> Good catch, thank you. Pushed a fix: `fetchPageAsync` now returns the response's resolved url even when there's no usable html to read (non-ok status, non-html content type, or no body), instead of discarding it along with the html. The well known favicon fallback picks that up automatically since it already reads `page.url`. Added a test for the specific case you flagged (redirect target answers with something other than html).

I'm confused. Are you an LLM ? 

I'll integrate this when v2 is released probably. You can try to rework the PR to merge into release/v2 branch if you'd like to.

junkerderprovinz:
No, I'm a person, but yes, I use Claude Code as a coding assistant on this one. I direct it and review everything before it gets pushed, and I stand behind the code and the replies. The reply you quoted was aimed at a CodeRabbit review comment, which probably added to the confusion.

On release/v2: I tried it before answering. The five commits cherry-pick onto release/v2 with no conflicts, and the diff comes out the same except for one thing: v2 added an MCP tool review gate (MCP_TOOL_ALLOWLIST) that doesn't exist on dev yet, so detectFavicon needed to be registered there too. Typecheck is clean in icons, validation, api and forms-collection, the full non-e2e suite passes (only pre-existing container-runtime failures unrelated to this, same on a clean release/v2), and oxlint/oxfmt report nothing new.

I've pushed the rebased branch (feat/app-form-favicon-detection-v2) to my fork. Let me know if you'd rather I retarget this PR to release/v2, or open a fresh PR against it and close this one.

Still open from before, in case you want it in the same pass: whether the href field should move above IconPicker, or whether you'd prefer keeping the order and adding a visible nudge on the icon field once a URL exists.

Issue #6600, opened 2026-08-15
🚀 Homarr v2 public beta is here! Join the testing


ajnart:
<p align="center">
  <img src="https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons/svg/homarr-wordmark-light.svg" width="700" />
</p>

<h1 align="center">🚀 Homarr v2 public beta is here!</h1>

We are very excited to introduce and share the biggest Homarr update ever, and we are looking for your help to test and break it before the official release.

> [!WARNING]
> Please **back up your current Homarr database** first and use a **new volume**.

## Main changes

| Feature                | What it brings                                                                                                                                     |
| ---------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Custom Widgets V2**  | A complete rewrite, can be used to build a custom widget for that shiny new tool added to your selfhosted arsenal                                  |
| **Community Workshop** | Create, upload, download, install, update, and share custom widgets and CSS with the community                                                     |
| **Performance**        | ~30% less RAM (hopefully), faster widget rendering, queries and caching, fewer repeated requests, much better browser performance for most widgets |
| **New drag and drop**  | Complete rework, responsive layouts, containers. **Sidebars are back** + no need to create mobile boards anymore                                   |
| **Onboarding**         | Completely re-worked to be much better and sexier                                                                                                  |
| **Advanced widgets**   | New mode, press shift on widgets for more details                                                                                                  |
| **Assistant**          | In the search bar for web searches, Homarr actions, custom CSS, creating widgets and much more                                                     |

> [!NOTE]
> The Assistant ships with a **"homarr provider"** that is free (and will never offer any paid plans). We use only ZDR providers, the provider is open-source, and there is **no prompt or response logging** — we're not selling your data.

## Screenshots

|                                                                                                        |                                                                                                    |                                                                                                   |
| ------------------------------------------------------------------------------------------------------ | -------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
| ![Dashboard overview](https://github.com/user-attachments/assets/d2a52fcd-18c1-4a7b-aa9b-2f64c78e12e2) | ![Custom widgets](https://github.com/user-attachments/assets/54a922b4-9c73-44d1-83cf-5360f2ec5e60) | ![Workshop](https://github.com/user-attachments/assets/9e205a56-e537-4636-ab5f-99b5b8a0ab70)      |
[...]

PR #6605, opened 2026-08-16
fix(media widget): text size fix and added season + Episode info

Bergasha:
This should fix most of the icon sizing issues when setting columns to anything greater than about 8. Text for Media widget is also set to mx not px that seems to fix the clipping issue.

I've also included season and episode info to media widget so it will show for example S04 E12.

Screenshot 2026-08-16 191746

- [x] I confirmed manually that the integration or widget is working with a real system and performed basic smoke
- [x] Optional but recommended: I confirm that I am willing to be added to the `CODEOWNERS` of this feature and will

## Summary by CodeRabbit

* **Style**
  * Standardized icon sizing across widgets and controls using responsive theme-based sizes.
  * Improved visual consistency across alerts, menus, statistics, navigation, and status indicators.
  * Refined media server table spacing, alignment, widths, and text sizing.
* **New Features**
  * Currently playing media entries can now display season and episode details when available.
  * Season and episode values are supported across major media server integrations.


ajnart:
Can you check the CodeRabbit's comments and fix them? Also, I believe there might be an easier way. Perhaps there's a way to make it so that, generally speaking, the icons of the library we use are sized using Mountain's dynamic size. That might be something worth looking into so that we don't need to remember to always do this pattern across the app, and we don't need to put a size, or we can just use S, M, L, XL. That would be useful. 

https://mantine.dev/guides/icons/

Says : 
> rem units in size prop
> The icon size prop is usually converted to width and height props under the hood. If you set size={16} it will be converted to width="16" and height="16" attributes on the svg element.
> You can use rem units in the size prop: size="1rem" will be converted to width="1rem" and height="1rem", but it's not recommended as it's prohibited by the SVG standard – some browsers (Firefox) will show a warning in the console.

Maybe we can do something like this : 
```tsx
export const iconSizes = {
  xs: {
    width: "var(--mantine-font-size-xs)",
    height: "var(--mantine-font-size-xs)",
    flexShrink: 0,
  },
  sm: {
    width: "var(--mantine-font-size-sm)",
    height: "var(--mantine-font-size-sm)",
    flexShrink: 0,
  },
  md: {
    width: "var(--mantine-font-size-md)",
    height: "var(--mantine-font-size-md)",
    flexShrink: 0,
  },
  lg: {
    width: "var(--mantine-font-size-lg)",
    height: "var(--mantine-font-size-lg)",
    flexShrink: 0,
  },
  xl: {
    width: "var(--mantine-font-size-xl)",
    height: "var(--mantine-font-size-xl)",
    flexShrink: 0,
  },
} as const;
```

then 
```
<Icon style={iconSizes.xs} />
<IconAlert style={iconSizes.md} />
```


I haven't tried yet so I can't tell if firefox generates issues for that change. 


Also as a follow up would you mind looking into https://github.com/homarr-labs/homarr/issues/6600#issuecomment-5306325493 you can see the beszel tooltip / edit mode widget hover features aren't correctly sized-up properly either

Bergasha:
@coderabbitai review

Bergasha:
> Can you check the CodeRabbit's comments and fix them? Also, I believe there might be an easier way. Perhaps there's a way to make it so that, generally speaking, the icons of the library we use are sized using Mountain's dynamic size. That might be something worth looking into so that we don't need to remember to always do this pattern across the app, and we don't need to put a size, or we can just use S, M, L, XL. That would be useful.
> 
> https://mantine.dev/guides/icons/
> 
> Says :
> 
> > rem units in size prop
> > The icon size prop is usually converted to width and height props under the hood. If you set size={16} it will be converted to width="16" and height="16" attributes on the svg element.
> > You can use rem units in the size prop: size="1rem" will be converted to width="1rem" and height="1rem", but it's not recommended as it's prohibited by the SVG standard – some browsers (Firefox) will show a warning in the console.
> 
> Maybe we can do something like this :
> 
> ```tsx
> export const iconSizes = {
>   xs: {
>     width: "var(--mantine-font-size-xs)",
>     height: "var(--mantine-font-size-xs)",
>     flexShrink: 0,
>   },
>   sm: {
>     width: "var(--mantine-font-size-sm)",
>     height: "var(--mantine-font-size-sm)",
>     flexShrink: 0,
>   },
>   md: {
>     width: "var(--mantine-font-size-md)",
>     height: "var(--mantine-font-size-md)",
>     flexShrink: 0,
>   },
>   lg: {
>     width: "var(--mantine-font-size-lg)",
>     height: "var(--mantine-font-size-lg)",
>     flexShrink: 0,
>   },
>   xl: {
>     width: "var(--mantine-font-size-xl)",
>     height: "var(--mantine-font-size-xl)",
>     flexShrink: 0,
>   },
> } as const;
> ```
> 
> then
> 
> ```
> <Icon style={iconSizes.xs} />
> <IconAlert style={iconSizes.md} />
> ```
> 
> I haven't tried yet so I can't tell if firefox generates issues for that change.
> 
> Also as a follow up would you mind looking into [#6600 (comment)](https://github.com/homarr-labs/homarr/issues/6600#issuecomment-5306325493) you can see the beszel tooltip / edit mode widget hover features aren't correctly sized-up properly either

@ajnart #6617 should fix this; I've tested in Firefox and all looks good with no errors.

ajnart:
@Bergasha maybe there is a more elegant fix rather than having to modify all of these files, like a global CSS overwrite or something on the base class for the majority of icons

Bergasha:
> @Bergasha maybe there is a more elegant fix rather than having to modify all of these files, like a global CSS overwrite or something on the base class for the majority of icons 

I agree, I'll look into it later today.

PR #6612, opened 2026-08-16
feat: recalculate permission scope & permission matrix UI

### LINKED ISSUE #2126: feat: Disable "apps" menu for users / groups that don't have access to modify boards

I only have 2 access groups set up, one for myself as admin, and another for "everyone".  I expected that since I have all board creating/editting disabled for the "everyone" group, they wouldn't be able to access the "apps" page and see all the apps I have available both publicly and as local IP addresses. Is it possible to have access to the "apps" page as a toggle? For me, the purpose of the seperate groups is so that I can limit the visibility of running apps to those groups.

### Additional information

_No response_

### LINKED ISSUE #4281: feat: Apps and Integration lists are not visible by default to users in the everyone group.

Apart from the applications we want to share (e.g., Jellyfin, Jellyseerr, etc.), there are also administrative applications (e.g., AdGuard, Portainer, etc.). Unfortunately, default users belonging to the everyone group can see the Apps and Integration lists, which include all URLs. To prevent non-admin users from seeing URLs to the administrative apps available in the Apps list, I ended up with two Homarr instances: one for regular users and another for admins. 

Instead of the  feature request #2078 Permission per App / App-Element / Dynamic Group I’d like a small improvement so that the Apps and Integration lists are not visible by default to users in the everyone group. The visibility of the Apps and Integration lists could be a group permission that’s granted optionally.

### Additional information

_No response_

### LINKED ISSUE #5231: bug: [RBAC] Prevent non-admin users from viewing the global "Applications" list (Information Disclosure)
### Describe the bug

In a multi-tenant or enterprise environment (like a WISP/NOC), we use Homarr to provide specific dashboards to different technicians or clients. We restrict their user group to only view their assigned board, with all global app permissions (create, modify, use full access) turned OFF. However, any restricted user can simply click on their profile -> Preferences -> Applications (left menu) and view the entire global pool of applications, including the names and private IP addresses (URLs) of infrastructure they should not have access to. This architectural design makes it a severe information disclosure vulnerability.

### Steps to reproduce

Create an admin user and a restricted standard user.

As admin, create multiple applications with private IP URLs.

Disable all global app permissions (create, modify, full access) for the standard user group.

Log in as the restricted standard user.

Click on Profile -> Preferences -> Applications (on the left menu).

Observe that the user can see the entire global pool of applications and their URLs, despite having no permissions.

### Impact

The impact is severe for multi-tenant, enterprise, or ISP/WISP environments. It creates an Information Disclosure vulnerability where users can see private infrastructure IPs belonging to other clients or restricted nodes. This prevents us from safely deploying Homarr to our clients.

Current Workarounds (and why they are bad):

Injecting Custom CSS (a[href*="/manage/apps"] { display: none !important; }) to hide the menu button. However, this is just "security by obscurity" and easily bypassed by anyone pressing F12.

Using Markdown/Text widgets with HTML links instead of actual Homarr Applications. However, doing this sacrifices the vital Ping/Status indicator functionality which is essential for a NOC.

### Additional information

Add a global Role/Group permission toggle that completely disables access to the "Applications" page in the user settings for non-admins.

Or, make the Applications list only display the apps that are actively used on the boards the user has read access to, hiding the rest of the global pool.

Note: I understand Homarr started as an amazing HomeLab project, but as it grows into enterprise environments, fixing this RBAC limitation will make it the ultimate dashboard. I believe this also relates to the community discussion in #2311.

### Version

1.55.0

### Installation method

Docker Run

### Browser

Edge (Proprietary)

### LINKED ISSUE #5526: feat: Apps and Integration menu items not visible for everyone group

Although I defined specific boards where I limit the apps for my users  all  urls are available in Apps  and Integrations  even for everyone default group. 
I would like that Apps and Integration menu items to be not visible for everyone group. These menu items should be avaialble for administrators and their visibility could be managed by the permissions for everyone group. 

### Additional information

_No response_

### LINKED ISSUE #6507: feat: disable integrations access without proper permissions

Since version 1.72 (https://github.com/homarr-labs/homarr/pull/6425) users without permission for board modifiers don't see apps as a menu item and cannot get the apps with url (.../manage/apps).
These users also don't see menu item with integrations. However these users still can get the intgerations with the url (.../manage/integrations)
Let's disable the integrations access for users without integration modification permission.

### Additional information

_No response_

### LINKED ISSUE #6608: bug: Integration page accessible without board-modify-all permission
### Describe the bug

PR [6509]https://github.com/homarr-labs/homarr/pull/6509) issue

### Steps to reproduce

Get integrations page (.../manage/integrations) without permission board-modify-all .

### Impact

Minor bug

### Additional information

_No response_

### Version

1.74.0

### Installation method

Docker Run

### Browser

Chrome

ajnart:
## Summary

Users without permissions could still reach management pages and see data they should not, and the group permission editor was a flat list of 20 switches that gave no sense of what any of them actually did. This PR fixes the scope server-side and rebuilds the editor as a permission matrix.

Closes #6608
Closes #6507
Closes #5231
Closes #2126
Closes #5526
Closes #4281

## The access rule

There is now exactly one rule, and the navigation, the pages and the management-home statistics all consume the same function (`getAppManagementAccess` / `getIntegrationManagementAccess` in `@homarr/definitions`):

> **A management section is reachable when you can do at least one thing in it, and it only ever lists what you are allowed to see.**

| | `/manage/apps` | `/manage/integrations` |
|---|---|---|
| Manage all | full list, searchable, editable | full list |
| Create only | empty state explaining you may only add, with the create action | empty state explaining you may only add, with the create action |
| Delegated full access on specific integrations | — | only those integrations, above a note that the view is scoped |
| None of the above | not found, with a link back to the management home | not found, with a link back to the management home |

Previously the nav could show a link to a page that returned not-found, and the stat cards were gated on unrelated permissions. They cannot drift now — that was the point of extracting the rule.

## Server-side scope (the actual boundary)

- `app.getPaginated` / `app.all` / `app.search` require `app-modify-all` (they were `protectedProcedure`, i.e. any signed-in user could enumerate every app name and URL — #5231).
- `integration.all` / `allThatSupportSearch` / `allOfGivenCategory` / `search` filter **server-side**: without a global integration permission you only receive integrations you have access to. Delegated per-integration grants still work. `search` now applies its limit *after* filtering, so a scoped user gets a full page of results rather than a truncated one.
- `home.getStats` counts only what the user may see.

## Create is a capability, not an access level

This is the substantive correction to the original design of this PR.

The first version of the matrix modelled every category as a single ordered level (None → View → Modify → Full) and dropped the five create-shaped keys — `board-create`, `app-create`, `integration-create`, `search-engine-create`, `media-upload`. To compensate, `group.savePermissions` force-bundled each create key with a parent. Two consequences:

- **Silent privilege escalation on upgrade.** A group holding `board-create` was rewritten to `board-full-all` — full control over every board, including deletion — on the next save of *any* permission. Same for `integration-create` → `integration-full-all`.
- **A lost capability.** "Members may create their own boards but not see anyone else's" — probably the most common self-hosted configuration — became inexpressible, and applying a preset silently revoked existing create grants.

Creating is orthogonal to how much of everyone else's data you can see, so it is now modelled that way: each row has an ordered access level **and** a checkbox per orthogonal **capability**. When the chosen level already implies a capability the checkbox is ticked and locked with a tooltip saying which level includes it. `savePermissions` stores exactly what was selected.

A ladder may only contain keys that genuinely imply each other. `app-modify-all` implies `app-create` but **not** `app-use-all` — only `app-full-all` grants both — so the app row is `No access | Modify | Full` with *Use all apps* and *Create apps* as capabilities. Stacking use beneath modify would have made picking Modify silently grant the use of every app, which is the same fabrication in the other direction. A test asserts every ladder is a real chain, so this class of mismatch fails CI rather than shipping.

All 20 permission keys are representable and round-trip losslessly; there is a test asserting the matrix covers `groupPermissionKeys` exactly, so adding a permission without placing it in the matrix now fails CI.

## Editor

- Per-category access level, with short labels (View / Use / Interact / Modify / Full) and the selected level's own description underneath, so the meaning is on screen rather than in the docs.
- A separate create checkbox per category.
- Viewer / Editor / Admin presets with tooltips describing what each grants, and a Custom badge when the selection matches none.
- "Also grants: …" hints for implied permissions.
- Choosing Administrator disables the other rows and explains why; setting it back to No access restores the previous selections.
- Collapsible effective-permissions preview (the de-duplicated set including everything granted indirectly).
- Fixes: saving left the unsaved-changes bar on screen (`setInitialValues` moves the values snapshot; only `resetDirty` clears the dirty flags), and a row showing only a create grant described itself as "no access".

## Verification

Exercised against a local instance with a seeded group holding the legacy `board-create` + `media-upload` combination:

- the two legacy keys render as checked, enabled create boxes and can be revoked — on `release/v2` they are invisible;
- raising Boards to Full locks the create box, lowering it back to No access re-enables it still checked, so an explicit grant survives;
- saving `board-create`, `app-create`, `integration-create`, `media-upload` persists exactly those four rows — the previous code also wrote `board-full-all`, `app-modify-all` and `integration-full-all`;
- a user with only `app-create`/`integration-create` sees both nav entries and lands on the create-only empty state instead of a 404;
- a user with no permissions sees no Apps/Integrations nav entries, no stat cards, and gets the not-found page with a back link on both routes;
- the matrix stacks without overflow at 390px.

## Tests

`permissions.spec.ts` covers matrix coverage of every key, create implication per category, standalone create keys, normalisation of redundant create keys, round-trips, and the two access rules. `group.spec.ts` covers no-escalation for each create key, de-duplication, clearing and idempotency. Plus the existing FORBIDDEN cases for the gated app procedures and the integration filtering tests.

## Docs

`management/users`, `management/apps` and `management/integrations` describe the levels, which of them include create, and the access rule per page.

## Review feedback addressed

- **CodeRabbit** — `createPermissionParents` escalating create upward (removed), integrations nav hidden from create-only users (fixed by the shared access rule), `integration.search` limiting before filtering (fixed), the anonymous-session query in `getStats` (already guarded), and the API permission documentation (added).
- **Codex** — the app ladder not matching the permission graph (fixed as described above), and the missing API migration guidance in `apps/docs/docs/management/api/index.mdx` (added: why app list endpoints can now return `FORBIDDEN`, and why integration lists may return fewer records).

## Follow-ups (not in this PR)

- `app.selectable` and `app.byIds` still return app names and URLs to any signed-in user. They back the board widget and bookmark pickers, so gating them needs its own change with its own testing.


ajnart:
Addressed the CodeRabbit review. Summary of the resolution per finding:

**Fixed (pushed in `c369df013`):**
- `integration.search`: access filter now runs before the limit is applied (no more dropped accessible matches); deterministic `name → id` tie-breakers added to both sort paths
- `app.search`: now allows `app-modify-all` **or** `board-modify-all` — board editors using the spotlight app search were broken by the earlier gate (regression), now covered by a new test
- `home.ts getStats`: anonymous sessions skip the per-integration permission queries entirely
- `group.savePermissions`: double type assertion replaced with a type-guarded lookup
- Shared integration access-relations helper extracted (no more duplicated query in `search`)
- New tests: board-modify-all app search, `integration-create` parent bundling, global `integration-full-all` branch, matrix round-trip expansion for a single higher-level key

**Intentional deviations (with reasons):**
- The requested `hasUseAccess === true` assertion for a global `integration-full-all` user cannot hold: `constructIntegrationPermissions` sets `hasUseAccess` only from per-integration rows or the `integration-use-all` key (verified in packages/auth/permissions/integration-permissions.ts:27). The test asserts `hasFullAccess === true` instead, which is the valid expression of the full-access branch.
- `board-create` parent changed from `board-modify-all` to `board-full-all`: per `groupPermissionParents`, only `board-full-all` implies `board-create` — the original value was inconsistent with the hierarchy.

**Not changed (deliberate design):**
- The API-side create→parent expansion in `group.savePermissions` is intentional (Phase A4 of the approved plan): create-only grants are never useful, so saving a create key always persists its implying parent. This is enforced at the API boundary so UI and API behavior stay consistent.
- `getFullAccessIntegrationIdsAsync` stays in-memory: it runs only for authenticated non-global users on the manage home; at homelab scale (a few hundred integrations) the count query would add more complexity than it saves.

Also archived the obsolete `restricted` test user and screenshots remain as committed.

ajnart:
@codex review

PR #6616, opened 2026-08-17
fix(beszel): stop small text from vanishing under board zoom

Bergasha:
This should fix Beszel's widget when having a large columns board, The text became illegible as it was tiny.

- [x] I confirmed manually that the integration or widget is working with a real system and performed basic smoke
- [x] Optional but recommended: I confirm that I am willing to be added to the `CODEOWNERS` of this feature and will

### DIFF
--- a/packages/widgets/src/beszel/_shared/tooltip.tsx
+++ b/packages/widgets/src/beszel/_shared/tooltip.tsx
@@ -11,7 +11,7 @@ const styles: Record<string, CSSProperties> = {
     border: "1px solid var(--mantine-color-dark-4)",
     borderRadius: 6,
     padding: "6px 10px",
-    fontSize: 12,
+    fontSize: "var(--mantine-font-size-xs)",
     lineHeight: 1.5,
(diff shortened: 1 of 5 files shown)


packages/widgets/src/beszel/_shared/tooltip.tsx
```diff
@@ -11,7 +11,7 @@ const styles: Record<string, CSSProperties> = {
     border: "1px solid var(--mantine-color-dark-4)",
     borderRadius: 6,
     padding: "6px 10px",
```
ajnart, line 13: We might want to audit / review all the other uses of px in the future

ajnart (approved):
So basically what's happening is that sometimes I used "px" as the suffix for the values. When it's just size = 12 it will use rem units under the hood for Mantine and that will scale with the page. 

I need to remember to do a 2nd pass to check these as well for consistency / only use predetermined sizes (S/M/L) instead of arbitrary numbers

PR #6623, opened 2026-08-18
fix(widgets): stop Health Monitoring system tab firing NOT_FOUND when Proxmox is mixed in

### LINKED ISSUE #6618: Health Monitoring widget: getSystemHealthStatus throws NOT_FOUND when a Proxmox integration is mixed with other integrations
### Description

The Health Monitoring widget's "system" tab throws a `NOT_FOUND` tRPC error on every board load when a Proxmox integration is attached to it alongside other (e.g. TrueNAS) integrations. The board still renders correctly once the client settles, but the error fires reliably and pollutes server logs continuously.

```
caused by TRPCError: 1 of the specified integrations not found or not of kinds openmediavault,dashDot,truenas,unraid,glances,synology,mock: ([<proxmoxId>,<truenasId1>,<truenasId2>] compared to [truenas:<truenasId1>,truenas:<truenasId2>]) code="NOT_FOUND"
```

### Root cause

`getSystemHealthStatus`'s server-side allowlist in `packages/api/src/router/widgets/health-monitoring.ts` was never updated when Proxmox support was added via the separate `getClusterHealthStatus` procedure:

```ts
const healthMonitoringIntegrationKinds = [
  "openmediavault",
  "dashDot",
  "truenas",
  "unraid",
  "glances",
  "synology",
  "mock",
] as const;
```

Proxmox is a legitimate `healthMonitoring`-category integration (`packages/definitions/src/integration.ts`), and the widget's own `supportedIntegrations` (`packages/widgets/src/health-monitoring/index.ts`) is category-derived and does include it — so the settings UI correctly allows attaching a Proxmox integration to the widget.

The client component (`packages/widgets/src/health-monitoring/component.tsx`) does correctly split integrations by kind — Proxmox goes to the cluster tab, everything else to the system tab:

```ts
const clusterIntegrationId = integrations.find(isClusterIntegration)?.id;

if (!clusterIntegrationId) {
  return <SystemHealthMonitoring {...props} />;
}

const otherIntegrationIds = integrations
  .filter((integration) => integration.kind !== "proxmox")
  .map((integration) => integration.id);
```

But `integrations` comes from `clientApi.integration.byIds.useQuery(props.integrationIds)`, which is empty on first render. Before it resolves, `clusterIntegrationId` is `undefined`, so the early-return branch fires and calls `<SystemHealthMonitoring {...props} />` with the **full, unfiltered** `integrationIds` from the widget's saved config — Proxmox ID included. `SystemHealthMonitoring` immediately fires `getSystemHealthStatus` with that full list, which hits the stale allowlist above and throws.

### Steps to reproduce

1. Add a Health Monitoring widget to a board.
2. Attach at least one Proxmox integration plus at least one integration from the system-tab allowlist (e.g. TrueNAS).
3. Load the board.
4. Check server logs — a `getSystemHealthStatus` `NOT_FOUND` error fires on every load (confirmed recurring roughly hourly/on every board refresh over several days).

### Expected behavior

Either:
- `getSystemHealthStatus`'s allowlist should exclude Proxmox explicitly (since it's meant to always route through `getClusterHealthStatus` instead), so a Proxmox ID in the query is silently filtered rather than triggering a hard `NOT_FOUND`, or
- The widget should defer firing `getSystemHealthStatus` until `integration.byIds` has resolved and the kind-based split is known, instead of falling through to the unfiltered `props.integrationIds` on first render.

### Actual behavior

A `NOT_FOUND` tRPC error fires on every board load for any Health Monitoring widget mixing a Proxmox integration with any other supported kind. The widget still renders correctly once the client-side split settles (both tabs work) — this is log noise rather than a broken feature, but it's persistent and makes it harder to spot real integration errors in the logs.

### Environment

- Homarr version: v1.75.0 (confirmed still present on `main` as of this report — not fixed by upgrading)
- Deployment: Docker (`ghcr.io/homarr-labs/homarr:v1.75.0`)

ajnart:
Fixes #6618

## Problem
When a Health Monitoring widget mixes a Proxmox integration with other supported kinds, loading the board fires a recurring `getSystemHealthStatus` `NOT_FOUND` tRPC error and pollutes server logs.

## Root cause
On first render `integration.byIds` has not resolved, so `clusterIntegrationId` is undefined and the early-return branch renders `<SystemHealthMonitoring {...props} />` with the **full unfiltered** `integrationIds` (Proxmox ID included). `SystemHealthMonitoring` immediately fires `getSystemHealthStatus` with that list, which hits the stale server allowlist that does not include `proxmox` → `NOT_FOUND`.

## Fix
Defer rendering until `integration.byIds` resolves so the kind-based split is known — Proxmox always routes to the cluster tab and `getSystemHealthStatus` only ever receives system-tab integration IDs.

## Alternative considered
Adding `proxmox` to the `healthMonitoringIntegrationKinds` allowlist would change server semantics; the client-side deferral is the scope-appropriate fix (matches the issue's second proposed option).

Closes #6618

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved health monitoring widget behavior while integration data is loading or unavailable.
  * Prevented incomplete integration data from causing display issues.
  * The widget remains hidden when integration information is loading or cannot be retrieved.

### DIFF
--- a/packages/widgets/src/health-monitoring/component.tsx
+++ b/packages/widgets/src/health-monitoring/component.tsx
@@ -18,16 +18,18 @@ const isClusterIntegration = (integration: { kind: IntegrationKind }) =>
   integration.kind === "proxmox" || integration.kind === "mock";
 
 export default function HealthMonitoringWidget(props: WidgetComponentProps<"healthMonitoring">) {
-  const { data: integrations = [] } = clientApi.integration.byIds.useQuery(props.integrationIds);
+  const { data: integrations, isLoading } = clientApi.integration.byIds.useQuery(props.integrationIds);
   const t = useI18n();
 
-  const clusterIntegrationId = integrations.find(isClusterIntegration)?.id;
+  if (isLoading) return null;
+
+  const clusterIntegrationId = integrations?.find(isClusterIntegration)?.id;
 
   if (!clusterIntegrationId) {
     return <SystemHealthMonitoring {...props} />;
   }
 
-  const otherIntegrationIds = integrations
+  const otherIntegrationIds = (integrations ?? [])
     // We want to have the mock integration also in the system tab, so we use it for both
     .filter((integration) => integration.kind !== "proxmox")
     .map((integration) => integration.id);
--- a/packages/widgets/src/health-monitoring/component.tsx
+++ b/packages/widgets/src/health-monitoring/component.tsx
@@ -18,10 +18,10 @@ const isClusterIntegration = (integration: { kind: IntegrationKind }) =>
   integration.kind === "proxmox" || integration.kind === "mock";
 
 export default function HealthMonitoringWidget(props: WidgetComponentProps<"healthMonitoring">) {
-  const { data: integrations, isLoading } = clientApi.integration.byIds.useQuery(props.integrationIds);
+  const { data: integrations, isLoading, isError } = clientApi.integration.byIds.useQuery(props.integrationIds);
   const t = useI18n();
 
-  if (isLoading) return null;
+  if (isLoading || isError) return null;
 
   const clusterIntegrationId = integrations?.find(isClusterIntegration)?.id;
 
--- a/packages/widgets/src/health-monitoring/component.tsx
+++ b/packages/widgets/src/health-monitoring/component.tsx
@@ -18,10 +18,10 @@ const isClusterIntegration = (integration: { kind: IntegrationKind }) =>
   integration.kind === "proxmox" || integration.kind === "mock";
 
 export default function HealthMonitoringWidget(props: WidgetComponentProps<"healthMonitoring">) {
-  const { data: integrations, isLoading, isError } = clientApi.integration.byIds.useQuery(props.integrationIds);
+  const { data: integrations, isPending, isError } = clientApi.integration.byIds.useQuery(props.integrationIds);
   const t = useI18n();
 
-  if (isLoading || isError) return null;
+  if (isPending || isError) return null;
 
   const clusterIntegrationId = integrations?.find(isClusterIntegration)?.id;
 


ajnart:
Hey @angst911! I believe I've found the root cause and applied a fix in #6623.

**Root cause:** On first render, `integration.byIds` hasn't resolved yet, so the widget's early-return branch renders the System tab with the **full, unfiltered** `integrationIds` — Proxmox ID included — which then fires `getSystemHealthStatus` against the stale server allowlist that doesn't include `proxmox` → `NOT_FOUND`. The widget settles correctly afterwards, which is why you only see log noise, not a broken feature.

**Fix:** The widget now defers rendering until `integration.byIds` resolves, so the kind-based split (Proxmox → cluster tab, everything else → system tab) is always known before `getSystemHealthStatus` fires.

Could you test it with the test image below to confirm the error stops appearing in your logs?

```yaml
services:
  homarr:
    image: ghcr.io/homarr-labs/homarr-test:pr-6623
    ...
```

Load your board with the Proxmox + TrueNAS Health Monitoring widget and check the server logs for the recurring `getSystemHealthStatus` `NOT_FOUND` error — it should be gone now.

ajnart:
Good catch — you were right, your first test build still had the bug, and I found the real reason. 🙏

**Why the first fix didn't work:** The widget used `isLoading` from the `integration.byIds` query to defer rendering. In TanStack Query v5, `isLoading` is `isPending && isFetching`. On the **very first synchronous render** after hydration the query is pending but hasn't started its network fetch yet, so `isFetching` is `false` → `isLoading` was `false` → the guard didn't trigger → the widget fell through to the System tab with the **full unfiltered `integrationIds`** (Proxmox ID included) → `getSystemHealthStatus` → `NOT_FOUND`. Exactly the signature you saw.

**The corrected fix (commit `a9960a67e`):** guard on `isPending` instead — it is `true` from the moment the query has no data yet, regardless of fetch state — so the widget can never render the System tab until the Proxmox/system split is known:

```ts
const { data: integrations, isPending, isError } = clientApi.integration.byIds.useQuery(props.integrationIds);
if (isPending || isError) return null;
```

**Please retest with the freshly rebuilt image** (the new commit just pushed, so give CI a few minutes):

```yaml
services:
  homarr:
    image: ghcr.io/homarr-labs/homarr-test:pr-6623
```

Load the board with your Proxmox + 2×TrueNAS Health Monitoring widget and check the server logs — the recurring `getSystemHealthStatus` `NOT_FOUND` should finally be gone.

angst911:
Retested with `ghcr.io/homarr-labs/homarr-test:pr-6623` (freshly pulled, new digest `sha256:1d2b587...`, commit a9960a67e). Confirmed fixed — loaded the board with the Proxmox + 2×TrueNAS Health Monitoring widget several times and checked server logs: no `getSystemHealthStatus` / `NOT_FOUND` errors at all, and no fallthrough call to the System tab with the unfiltered integration list. Widget renders correctly with live data. 🎉

PR #6582, opened 2026-08-13
feat: enhance release v2 experience

ajnart:
## Summary

- unify release-v2 creation, onboarding, widget, integration, Docker, and Kubernetes recovery flows
- add in-widget integration editing with connection testing and diagnostics while preserving unsaved widget settings
- add canonical feature contracts, generators, documentation, accessibility improvements, and resilient multi-endpoint behavior

## User impact

Users can create and repair apps, widgets, and integrations with fewer context switches. Docker and Kubernetes surfaces remain responsive during partial outages, and contributors get one deterministic path for adding and validating features.

## Stack

- base: #6569
- this is the eighth feature PR in the release-v2 stack
- #6545 remains the roll-up PR above this layer

## Validation

- affected-package typechecks: 7/7
- focused cross-layer tests: 119/119
- request-handler Node tests: 74/74
- feature-platform contracts: 8/8
- production documentation build
- formatting and changed-file lint with no errors
- three independent review passes with no remaining blockers

The Docker image build is intentionally left for local verification.

## Summary by CodeRabbit

* **New Features**
  * Added Universal Create from the header, command menu, Spotlight, and boards.
  * Added board empty states, setup checklists, widget recommendations, and inline integration editing.
  * Added multi-endpoint Docker management, reconciliation, health status, capability-aware actions, and safer logs.
  * Added Kubernetes context selection with context-specific resources and graceful metrics availability.
  * Added development workbench and tooling for feature validation and scaffolding.

* **Documentation**
  * Expanded Docker, Kubernetes, board management, integration, and development documentation.

* **Bug Fixes**
  * Improved permission handling, connection errors, modal loading states, and unavailable service feedback.

PR #6503, opened 2026-08-01
feat: rebuild board grid with dnd-kit

ajnart:
## Summary

- replace GridStack with a dnd-kit board editor supporting transactional drag-and-drop, collision handling, overlays, and eight-direction resizing
- replace category and dynamic-section variants with a unified Container model supporting labels, collapsing, nesting, resizing, rails, and cross-container moves
- preserve responsive layouts, keyboard and touch input, cancellation and rollback behavior, migrations, old imports, documentation, and lazy editor loading
- document the actual scaled-canvas behavior and complete tile keyboard movement and resizing controls

## Stack

Layer 2 of 8. Review and merge bottom-to-top:

1. #6356 — Custom Widgets v2
2. **#6503 — dnd-kit board grid**
3. #6502 — Widget modernization
4. #6450 — Responsive board layouts
5. #6482 — Homarr Assistant
6. #6555 — Widget performance
7. #6569 — Onboarding rebuild
8. #6545 — `release/v2` roll-up

This PR targets `v2`; #6502 is based on this branch.

## Validation

Use the checks attached to the current PR head as the source of truth for hosted validation.

PR #6502, opened 2026-08-01
feat: modernize dashboard widgets

ajnart:
## Summary

- modernize the widget catalog with denser responsive layouts, consistent loading/error/stale states, and bounded integration data
- add distinct advanced views for supported widgets while keeping compact-only widgets unchanged
- support Shift-hover previews, a keyboard-accessible advanced-view trigger, context-menu entry, and focus restoration
- centralize viewport tracking so loaded widgets do not each create a resize subscription
- preserve integration identity and permission-aware actions across multi-source widgets
- align Downloads documentation with upload-speed filtering, permissions, client capabilities, and current-view sorting

## Stack

Layer 3 of 8. Review and merge bottom-to-top:

1. #6356 — Custom Widgets v2
2. #6503 — dnd-kit board grid
3. **#6502 — Widget modernization**
4. #6450 — Responsive board layouts
5. #6482 — Homarr Assistant
6. #6555 — Widget performance
7. #6569 — Onboarding rebuild
8. #6545 — `release/v2` roll-up

This PR targets `feat/dnd-kit-grid`; #6450 is based on this branch.

## Validation

Use the checks attached to the current PR head as the source of truth for hosted validation.

PR #6637, opened 2026-08-20
fix: preserve board drafts and widget scroll state

ajnart:
## Summary

Applies two isolated, low-risk UI fixes selected from the eight-PR `release/v2` review. All other review follow-ups are intentionally omitted, so those areas retain the current `release/v2` behavior.

### Fixes

- **Board settings:** Reset from Base resets only the persisted layout and no longer saves unrelated dirty settings.
- **Media Missing:** switching between Missing and Queued restores each tab scroll position for the lifetime of the mounted widget.

### Scope

Two files, 19 additions, and 8 deletions. There are no changes to Custom Widgets, Assistant, backup import, integrations, Next.js configuration or route preloading, board scaling/editing, onboarding, translations, documentation, or the container entrypoint.

### Validation

Formatting and lint pass on both files. Focused typechecks pass for Next.js and Widgets. No broad local suite, Docker build, or E2E suite was run.

## Full stack reviews

The reports below are the original review records and evidence that informed this selection. Most reported findings are intentionally not implemented by this minimal PR. `.reviews/` and its screenshots remain outside the branch.

<details>
<summary><h2>PR #6356 review — Custom Widgets v2</h2></summary>

Merge base: `2513e8454f240caa315b68dd6738b79838eeeecc`  
Reviewed head: `e41d938a5fc223249f767474a7e4744087258225`

### [P1] Put response caching and invalidation in shared storage — `packages/custom-widgets/src/server/request-executor.ts:66`

The response cache, in-flight registry, and invalidation epoch are module-level process memory. `invalidateCustomWidgetResponseCache()` only mutates those local objects, even though this PR documents a shared Redis cache/version contract in `apps/docs/docs/management/custom-widgets/index.mdx:119` and `requests-and-security.mdx:52`. In a normal multi-replica self-hosted deployment, an action or `RefreshButton` handled by replica A leaves replica B's cached response valid until its TTL (up to one hour). A subsequent poll routed to B therefore renders stale state immediately after a successful mutation. Move response storage/invalidation generation into the shared Redis path, or otherwise broadcast a version change that every replica observes.

```mermaid
flowchart LR
  U[Viewer runs action] --> A[Replica A]
  A --> IA[Clear Map A]
  U --> LB[Next poll via load balancer]
  LB --> B[Replica B]
  B --> SB[Map B still contains stale response]
```

### [P1] Schedule load queries within the concurrency budget — `packages/api/src/router/widgets/custom-api.ts:210`

`getData` starts every load query with one `Promise.all`, while the new limiter rejects the fifth concurrent request for the same user/item (`CUSTOM_WIDGET_USER_ITEM_CONCURRENCY_LIMIT = 4`). The manifest schema nevertheless accepts up to 64 requests. A valid widget with five slow load queries therefore renders four results and one `TOO_MANY_REQUESTS` error on every refresh. The authoring path has the inverse problem: `loadPreviewQueries()` awaits each independent request serially at `apps/nextjs/src/app/[locale]/manage/custom-widgets/_custom-widget-form-utils.ts:107`, so eight three-second sources take about 24 seconds instead of two four-wide batches. Use one bounded queue/pool in both runtime and preview so requests are parallel without being rejected by their own safety limit.

```text
Current runtime:  q1 q2 q3 q4 | q5...qN rejected
Current preview:  q1 -> q2 -> q3 -> ... -> qN
Expected:         [q1 q2 q3 q4] -> [q5 q6 q7 q8] -> ...
```

### [P1] Reject anonymous cache-refresh mutations — `packages/api/src/router/widgets/custom-api.ts:198`

The refresh endpoint is a `publicProcedure`, performs no session check or rate limiting, and scans the local cache prefixes. The rendered `RefreshButton` is also enabled whenever an item ID exists (`packages/custom-widgets/src/runtime/refresh-button.tsx:40`). Consequently, an anonymous viewer of a public board can invoke refresh repeatedly—or call the public endpoint directly—to keep evicting shared responses, add database/cache-scan load, and force normal polling viewers back to upstream APIs. This contradicts the explicit public-board security contract in `apps/docs/docs/management/custom-widgets/custom-jsx.mdx:41` and `requests-and-security.mdx:48`. Require authentication server-side and pass authenticated state to the runtime so the control is disabled as documented.

### [P2] Give each anonymous viewer a distinct limiter identity — `packages/api/src/router/widgets/custom-api.ts:155`

`RequestLimitInput` includes `anonymousId`, and the limiter has tests proving different anonymous IDs receive separate user/item buckets, but this only passes `ctx.session?.user.id`. Every public-board visitor therefore becomes `anonymous:unknown`. Five visitors loading the same slow widget at once consume the shared four-slot user/item bucket and one receives `TOO_MANY_REQUESTS`; one busy visitor can likewise exhaust the 60/minute bucket for everyone else. Populate a privacy-safe, stable request identity (the repository already has forwarded-IP extraction utilities) or deliberately key an equivalent per-client token instead of falling back to the global `unknown` bucket.

### [P2] Bound the response cache by bytes, not only entry count — `packages/custom-widgets/src/server/request-executor.ts:289`

The process-local cache allows 1,000 parsed responses while each response may be 1 MiB. That permits roughly 1 GiB of retained payloads before object overhead. Because cache keys include runtime parameters and cache TTLs may be one hour, a public manual query with high-cardinality input can fill those entries in about 17 minutes at the allowed 60 requests/minute, potentially exhausting the Node process on common self-hosted memory limits. Enforce a total byte budget (and preferably per-definition/item budgets) and evict by weight/LRU rather than count alone.

### [P2] Forward `SubFetch` metadata to authored callbacks — `packages/custom-widgets/src/jsx/jsx-emitter.tsx:71`

The emitter wraps a `SubFetch` arrow child as a one-argument function and calls `renderCallback(callback, [value], ...)`. The runtime intentionally invokes children as `(data, metadata)`, and both the authoring prompt and public runtime documentation promise `(result, meta)`. A valid template such as `{(result, meta) => <Text>{meta.status}</Text>}` therefore evaluates `meta` as `undefined` and fails inside the widget after the request succeeds. Accept both runtime arguments and pass `[value, metadata]` into the interpreter callback.

### [P2] Parse an unchanged JSX template only once — `packages/custom-widgets/src/runtime/custom-jsx-renderer.tsx:169`

Every data/status/input update calls `renderSafeJsx`, which reparses the unchanged template with Acorn at `packages/custom-widgets/src/jsx/interpreter.tsx:359`; it then serializes all bindings again to build the error-boundary key. Custom widgets permit 50 KiB templates, 1 MiB responses, and one-second refresh intervals, so several dashboard tiles can repeatedly perform full parse plus full-data serialization on the browser main thread even when only one response field changed. Memoize the AST by template (then interpret it with current bindings), and avoid hashing the entire response object solely to reset the boundary.

## Overall assessment

Changes requested. The new safety model is directionally strong—server-only credentials, DNS pinning, same-origin redirects, request/response budgets, board-permission checks, server-authenticated authoring pages, and lazy loading of the dashboard interpreter are all good foundations. The current cache and scheduling implementations, however, break the PR's explicit shared/bounded guarantees and expose public-board availability risks. The runtime callback and repeated parsing defects also affect ordinary authored widgets.

## UI / browser evidence

The requested command shape, `go run . --pr 6356 --demo --detach`, is not accepted by the current CLI (`unknown flag: --pr`). I used its current equivalent from `tools/homarr-dev`: `go run . run --pr 6356 --demo --detach`. This pulled the published PR image, started `homarr_pr_6356` at `http://localhost:7575`, and passed the live health check. The harness was stopped after review.

Using `agent-browser` with the required core/dogfood guidance and the supplied `demo` / `demo` credentials, I verified:

- login and the seeded five-widget Custom Widgets list;
- the new create workbench, including General/API sources/Requests/Options/JSX/Preview navigation;
- the narrow 390×844 mobile layout and fixed save/preview controls;
- no browser errors reported on the changed create page during the smoke pass.

Evidence:

- **Custom Widgets list** (see screenshot evidence below)
- **Create workbench — desktop** (see screenshot evidence below)
- **Create workbench — mobile** (see screenshot evidence below)
- **Login** (see screenshot evidence below)

The cache, cross-replica, anonymous-concurrency, and upstream-request cases require controlled backend topology and were validated statically rather than by mutating the published demo.

## Material test gaps / residual risks

Per the review request, I did not run broad tests, builds, Docker compilation, E2E, unit suites, or a full typecheck, and I did not modify source or add tests. `git diff --check` is clean.

The focused suite currently lacks coverage for:

- invalidation observed by a second Homarr replica;
- five or more slow load queries under the real limiter;
- distinct simultaneous anonymous public-board viewers and anonymous refresh rejection;
- cache eviction by aggregate payload bytes;
- the documented two-argument `SubFetch` callback through the safe JSX emitter;
- repeated renderer updates without reparsing the template.

The Workshop PocketBase rules, source-secret rebinding transactions, preview/configuration-request Redis state, migrations across SQLite/MySQL/PostgreSQL, generated API/docs assets, and surrounding board-placement guards were inspected. I found no additional actionable regression in those paths.

## Standards vs Spec

- **Standards:** RSC authentication stays server-side, heavy dashboard rendering is dynamically imported, independent database reads are generally parallelized, and SSRF/secret boundaries are explicit. The load-query waterfall/fan-out mismatch and repeated client parsing conflict with the applicable Next.js/React performance guidance.
- **Spec:** The implementation does not meet the PR body's “bounded” execution claim or the docs' shared-Redis and anonymous-refresh guarantees. The `SubFetch` callback also diverges from the authoring contract. The remaining reviewed product journeys match the PR description and repository conventions.

### Screenshot evidence

PR #6356 review — Custom Widgets v2 screenshots

</details>

<details>
<summary><h2>Review: PR #6450 — Responsive board layouts</h2></summary>

### [P2] Resetting a layout also commits unrelated unsaved board settings

`apps/nextjs/src/app/[locale]/boards/[name]/settings/_layout.tsx:94`

The reset handler calls the settings form's broad `saveSettingsAsync()` whenever *any* field is dirty. That function persists every changed partial board setting (page title, background, appearance, behavior, and so on) alongside layout changes before `resetLayout` runs (`_settings-form.tsx:130-149`). A user can therefore edit an unrelated setting, click **Reset from Base**, and have that unrelated draft committed even though the confirmation says only saved positions and sizes will be replaced. It is also committed when the subsequent reset fails. Persist only the layout data needed to resolve the canonical layout ID, require layout changes to be saved first, or explicitly include the additional writes in the confirmation.

## Overall assessment

One actionable regression found. The protected Mobile/Base roles, removable custom layouts, role normalization, independent placement records, proportional previews, and explicit selected-layout editor override otherwise line up with the PR specification and inspected call sites.

The full 63-file layer was reviewed against merge-base `08588da0d1cfa72401a773c1f83e20a106ffe50e` through head `37289a224728ce4ce4591d5ec373e274fe7637eb`, including the API/router, database migrations and seed repair, validation/definitions, import normalization, board providers/editor, settings UI, documentation, and relevant existing tests. No tests or builds were run, per the requested focused-review policy.

## Browser evidence

The published PR image was exercised with demo credentials at `http://localhost:7575`. The requested command is stale for the current CLI (`unknown flag: --pr`); the equivalent supported command, `go run . run --pr 6450 --demo --detach`, launched `ghcr.io/homarr-labs/homarr-test:pr-6450` successfully.

- Mobile and Base rendered as protected; only a locally added custom layout exposed **Remove**. Add/remove updated the form as expected.
- Mobile editing used the representative `390px` width. Explicit Base editing remained selected at a `600px` viewport and used the `1280px` editor width, rather than silently switching to Mobile/custom.
- Reset confirmation copy was verified after making an unrelated setting dirty.
- Mutation/persistence verification was blocked by demo mode as designed. Container logs returned `TRPCError: Mutations are disabled in demo mode` with `code="FORBIDDEN"` for both `board.savePartialBoardSettings` and `board.saveLayouts`; therefore the reset side effect above is demonstrated by the direct handler/call-chain rather than a completed persisted mutation.

Evidence:

- **`layout-settings-initial.png`** (see screenshot evidence below)
- **`reset-step-1-dirty-settings.png`** (see screenshot evidence below)
- **`reset-step-2-confirmation.png`** (see screenshot evidence below)
- **`mobile-selected-layout-editor.png`** (see screenshot evidence below)
- **`base-selected-at-tablet-viewport.png`** (see screenshot evidence below)

## Residual risks and test gaps

- Layout previews synchronously re-project every layout on form rerenders and repeatedly search board items/sections. The representative demo board remained responsive, so this was not elevated to a finding; a profiler run with a large advanced-user board and several custom layouts would be the useful follow-up.
- Demo mode prevented end-to-end verification of saved add/remove/reset results and post-refresh persistence.
- Existing normalization and router tests cover core role and projection cases, but browser-level persistence coverage was removed with the legacy responsive-layout E2E suite in this layer.

## Standards vs Spec

**Standards:** The implementation follows the repository's Mantine/Tabler conventions, keeps browser-only editor state inside client components, and updates the user/API documentation. No demonstrated Next.js RSC, hydration, bundle, or accessibility regression was found. The preview hot path remains a performance risk worth profiling at large-board scale.

**Spec:** Protected Mobile/Base roles, custom layout lifecycle, normalization/import repair, selected-width editing, and independent placement storage match the PR description. The broad save performed by **Reset from Base** violates the action's stated narrow scope and is the sole finding above.

### Screenshot evidence

Review: PR #6450 — Responsive board layouts screenshots

</details>

<details>
<summary><h2>PR #6482 review — Homarr Assistant</h2></summary>

### [P2] Keep the read-only demo from mounting a write-dependent runtime

`packages/api/src/router/assistant.ts:456`

`getAvailability` unconditionally exposes the Assistant when `DEMO_MODE` is set, but the Assistant cannot initialize or persist a conversation without `createThread` and `appendMessage` mutations. `DEMO_READ_ONLY` defaults to `true`, and the existing tRPC middleware rejects every mutation in that mode. The published preview launched with the documented `--demo` option therefore showed the Assistant but every send ended with `Mutations are disabled in demo mode`; the history drawer stayed empty. The canned response at `apps/nextjs/src/app/api/assistant/chat/route.ts:410` does not solve this because the remote-thread/history mutations run around the chat request. This makes the public read-only demo path unusable. Use an in-memory thread/history adapter for the canned demo, permit only private Assistant-history writes, or do not report the Assistant as enabled when `isDemoReadOnly` is true.

### [P2] Route provider traffic through Homarr's trusted HTTP stack

`packages/api/src/router/assistant.ts:211`

Model discovery uses the global `fetch`, and the completion provider created at `apps/nextjs/src/app/api/assistant/chat/route.ts:602` is also left on the AI SDK's default transport. Neither request uses `fetchWithTrustedCertificatesAsync` / `UndiciHttpAgent`, even though that is the repository's path for managed CAs, trusted hostnames, proxy environment handling, DNS initialization, and redacted request logging. A self-hosted administrator can follow the new local-model documentation and point Assistant at an HTTPS Ollama, LM Studio, or custom endpoint signed by a private CA already trusted in Homarr, yet discovery reports the endpoint unreachable and chat streaming fails TLS verification. Both discovery and the SDK provider need the same Homarr-managed fetch/dispatcher.

```mermaid
flowchart LR
  A[Assistant model discovery and chat] --> B[global fetch / AI SDK default fetch]
  B --> C[system trust store only]
  D[Homarr certificate management] --> E[fetchWithTrustedCertificatesAsync]
  E -. currently bypassed .-> A
```

### [P2] Do not auto-load arbitrary model-controlled image URLs

`apps/nextjs/src/components/assistant/assistant-markdown-image.ts:7`

Every absolute HTTP(S) Markdown image returned by the model is accepted, and `AssistantImage` immediately renders it as an `` (`assistant-image.tsx:61`). Model output is untrusted: a prompt-injected tool/source or compromised provider can emit a tracking URL and cause each viewing browser to contact an attacker-controlled host, revealing the user's client IP and probing browser-reachable services. `referrerPolicy="no-referrer"` removes the Referer header but not the request. Restrict automatic loads to same-origin/known Homarr image URLs, proxy them through a hardened image path, or require an explicit click before loading an external image.

### [P2] Make the automatic-approval popover own Escape and keyboard focus

`apps/nextjs/src/components/assistant/assistant-panel.tsx:2094`

This controlled `Popover` has neither the focus/return-focus behavior used by the context popover nor the trigger Escape handler used by conversation history. In the published preview, opening **Automatic approvals** and pressing Escape repeatedly left it at `aria-expanded=true`; pressing Tab moved focus to **New conversation** instead of the popover's **Enable for this conversation** action. The outer panel correctly refuses to close while the visible nested dialog exists, so the result is a keyboard-sticky overlay. Add focused nested-overlay behavior (for example `trapFocus`/`returnFocus` plus reliable Escape dismissal) before suppressing the panel's Escape handler.

Evidence: **`escape-popover-stuck.png`** (see screenshot evidence below)

### [P3] Document the automatic-approval exception to per-change confirmation

`apps/docs/docs/management/assistant.mdx:52`

The security section states without qualification that changes require approval. The shipped control can instead automatically answer every mutation approval affirmatively for the rest of a conversation (`assistant-panel.tsx:888-895`), as well as accept forms and affirmative questions. The UI discloses this, but administrators reading the setup/privacy documentation get a materially stronger approval guarantee than the product provides. Document the opt-in, its conversation scope/reset behavior, that individual changes then run without a fresh click, and that Homarr permissions remain authoritative.

## Overall assessment

**Needs changes.** The per-user ownership checks, encrypted provider secrets, mutation classification, server-side tRPC callers, permission enforcement, RSC parallelization, and enabled-only dynamic Assistant bundle are sound in the reviewed paths. The blocking concerns are the default demo contradiction, bypassing Homarr's self-hosted certificate/network stack, model-controlled client fetches, and the explicit nested-Escape claim not holding for the most security-sensitive popover.

## Assumption / boundary audit

```text
Can the advertised demo run without writes?
└─ No: thread initialization and message history are mutations -> finding 1

Can a configured local provider use Homarr's managed trust/network policy?
└─ No: discovery and chat use default transports -> finding 2

Can untrusted model output create browser-side effects before user consent?
└─ Yes: remote Markdown images auto-load -> finding 3

Does every nested overlay consume Escape and keep focus reachable?
└─ No: automatic approvals stays open and its action is skipped by Tab -> finding 4
```

## Browser evidence and blockers

- Published image: `ghcr.io/homarr-labs/homarr-test:pr-6482` (digest `sha256:54fd57014693d4f46c77fd829a7075f0387176b6fd111ba16b272c641cd69e6d`) at `http://localhost:7577`; login `demo` / `demo`.
- With `go run . run --pr 6482 --demo --detach`, user-menu entry, board widget, floating panel, and composer rendered, but sends failed with `Mutations are disabled in demo mode` and history was empty.
- Relaunched the same image with `--demo -e DEMO_READ_ONLY=false`: canned streaming completed, the thread persisted and appeared in history, user-menu and `Shift+A` entry points worked, the board widget stayed synchronized, and panel close/history/context Escape behavior worked.
- Automatic-approval Escape/focus failure was reproduced on the floating panel. Desktop and 390x844 mobile layouts remained usable. Screenshots: **`desktop-assistant.png`** (see screenshot evidence below), **`mobile-assistant.png`** (see screenshot evidence below).
- The management configuration UI was inspected. Real provider discovery/chat, external-provider failure states, real permission-gated tools, mutation approval execution, source rendering, and quota behavior cannot be exercised in the canned demo: the demo route returns before provider/model/tool execution and no external credentials were supplied. The browser file picker also did not expose a persistent file-input node to the CLI, so attachment selection was not completed.

## Residual risks and test gaps

- No unit, E2E, Docker build, or broad typecheck was run, per instruction; validation was source review plus the targeted published-preview browser session.
- The merge-base diff is unusually large: 308 files, 41,220 additions and 2,657 deletions. It includes Workshop/provider and lower-stack Custom Widget/board overlap, which raises residual integration risk despite full diff inventory and focused call-site review.
- Live multi-user privacy, a second Homarr process, private-CA provider connectivity, malformed provider responses, real streaming interruption, and real tool approval/denial should receive targeted validation outside demo mode.

## Standards vs. spec

- **Standards:** React/Next boundaries generally follow the reviewed Next.js and Vercel guidance: server reads in the root layout are parallel, the expensive Assistant runtime is behind an enabled-only dynamic boundary, Mermaid/Shiki are lazy, private values remain server-side, and context subscriptions are shared. Findings 2-4 are the concrete self-hosting, privacy, and accessibility deviations.
- **Spec:** Private per-user threads and widget synchronization, encrypted credentials, permission-checked server tools, and mutation approval classification are present. The PR's explicit claims that the demo is usable, local/self-hosted providers work in Homarr's environment, and nested overlays retain Escape control are not fully met. The documentation also omits the automatic-approval exception to its approval guarantee.

## Review scope

- Head: `8c42ff9962f44b78182bd1d1bed864ac5de8e111`
- Current remote base: `origin/feat/auto-mobile-boards` at `37289a224728ce4ce4591d5ec373e274fe7637eb`
- Merge-base reviewed: `854b59ede41cc907e8bf45852d4fe48890aa1757`
- The current base is two commits ahead of the merge-base; review used the full `merge-base..HEAD` layer as requested.

### Screenshot evidence

PR #6482 review — Homarr Assistant screenshots

</details>

<details>
<summary><h2>PR #6502 — widget modernization</h2></summary>

### [P1] Closed and ineligible context menus subscribe every widget to the whole query cache

**Location:** `apps/nextjs/src/components/board/items/widget-context-menu.tsx:208`

`WidgetContextMenu` now installs `integration.all.useQuery`, `useIntegrationsWithInteractAccess`, and a predicate-based `useIsFetching` subscription before the eligibility return at line 208. This runs for every tile even when the user is anonymous, right-click menus are disabled, the board is in edit mode, the item is an app, or the menu has never opened. The previous implementation performed the eligibility return in an outer component and did not mount the query/status/action dropdown until the menu was open.

TanStack Query still deduplicates the `integration.all` network request, but it does not deduplicate hook observers: every tile adds a query observer and a QueryCache subscription. Every query-cache notification must consequently reevaluate every widget's predicate against active queries. This is especially costly on the dense, frequently polling dashboards Homarr's advanced self-hosting users build. It also starts the protected `integration.all` query on anonymous public boards even though the menu is immediately discarded.

Restore the cheap outer eligibility guard, keep only the context-menu target/open state mounted, and move integration loading, permission-derived action construction, query matching, and `useIsFetching` into a dropdown component mounted only while `menuOpened` is true.

### [P2] The centralized viewport store broadcasts advanced-focus changes through every widget shell

**Location:** `apps/nextjs/src/components/board/items/item-content.tsx:164`

Every `LoadedBoardItemContent`, including widgets that do not support advanced focus, consumes the complete `{ active, viewportSize, ...actions }` context. Each nested `WidgetContextMenu` consumes it again. Because the provider value changes for every viewport update and every `visible -> closing -> null` transition, opening one advanced widget or resizing the window rerenders all widget shells and menus. `InnerContent` is memoized, so the data-heavy widget implementation is protected, but all shell/card/menu elements and the expensive menu computations above are still repeated.

The published demo board contained 36 widget sources (12 advanced-capable), so one viewport/active update fans out to at least 72 context consumers:

```text
active focus / viewport size
             |
      one context value
        /           \
36 widget shells   36 context menus
```

Split stable commands from changing focus state and use item-scoped selection (or an external store with selectors) so only the active/previous item rerenders. Viewport dimensions are only needed to lay out the active surface and should not be read by every inactive tile.

## Overall assessment

The functional modernization is broad and generally cohesive, but the layer is not ready for a dense self-hosted dashboard until the two subscription/render fan-outs above are addressed. The API partial-failure shapes preserve integration IDs/names, the advanced implementations inspected keep compact and advanced behavior distinct, and dynamic widget component loading remains intact.

## Browser evidence

Used the published PR image via `go run . run --pr 6502 --demo --detach` at `http://localhost:7577` with the demo account; no source image build was performed.

- Calendar: keyboard/manual advanced view opened, close received initial focus, Escape closed it, and focus returned to the originating trigger.
- Calendar: Shift-hover preview appeared after the 500 ms hold and disappeared after Shift release without opening a modal.
- Downloads: context menu exposed advanced view, quick options, scoped refresh status, and settings; the advanced table exposed the complete column set and retained action controls.
- Weather: at a 390 px viewport, the advanced dialog stayed within the viewport (`x=8`, `width=374`, document width `390`) and its wide forecast remained internally scrollable.
- Responsive board at 390 px had no document-level horizontal overflow. No browser console errors were reported during these journeys.

## Test gaps and residual risks

- No focused regression test or instrumentation asserts that closed/ineligible menus do not install QueryCache observers, or that one focus/viewport update rerenders only the affected shell.
- Partial-integration failure and permission-denied action states were reviewed statically but were not reproducible with the all-success mock integrations in the published demo.
- The Downloads `limitPerIntegration` option and API input still have no upper bound. That behavior already exists at the merge base, so it is not filed as an introduced layer finding, but it remains inconsistent with the PR's blanket “bounded integration data” claim.
- No unit, E2E, Docker compilation, full typecheck, or broad test suite was run, per the requested focused-review constraints. `git diff --check` passed.

## Standards vs Spec

- **Standards:** Accessibility and responsive browser paths passed; safe-link handling, stale/error indicators, integration attribution, and lazy widget component loading were consistently applied in the sampled high-risk families. React subscription locality is the material standards failure.
- **Spec:** Advanced triggers, focus restoration, responsive advanced views, partial integration identity, and documentation are present. The implementation centralizes the resize listener, but not the resulting render work; the query-status feature also defeats the intended self-hosted resource discipline while menus are closed.

</details>

<details>
<summary><h2>PR #6503 review</h2></summary>

## Findings

- [P1] Stop inverse-scaling controls once tiles no longer have room — `apps/nextjs/src/components/board/layout/scaled-board-canvas.tsx:60`

  `calculateBoardUiScale` keeps Mantine typography and controls at their normal physical size while the fixed grid continues shrinking to any scale. On the published demo board, a 412 px viewport selected the only available layout and produced `data-canvas-scale="0.14264264264264265"`. The document correctly stayed 412 px wide, but a logical `200 × 200` tile became roughly `29 × 29` physical pixels while its text and controls stayed near desktop size. Labels overlap adjacent widgets, table text draws across columns, and controls are clipped or stacked on top of one another; the board is not operable or legible. This affects any existing or newly created board that does not have a sufficiently narrow responsive layout. The behavior also conflicts with `PRODUCT.md:31`, which requires horizontal scrolling before controls become inaccessible. Clamp the inverse UI scaling / canvas down-scaling at a usable threshold and allow horizontal scrolling (or otherwise reflow) before the tile content becomes smaller than its preserved UI.

  Evidence: **mobile view at 412 px** (see screenshot evidence below) and **mobile edit mode at 412 px** (see screenshot evidence below).

- [P1] Disable native touch panning on widget drag handles — `apps/nextjs/src/components/board/sections/grid/section-grid.module.css:26`

  The documented touch journey says to press and hold a widget and then move it, and the `PointerSensor` uses a 200 ms delay for touch. However, `.editorEntry`, its outer `.board-grid-entry`, and the portal mount all retain computed `touch-action: auto`. Browsers decide native pan handling when the gesture begins, so moving after the delay can hand the gesture to page scrolling and emit `pointercancel`, canceling the grid transaction instead of dragging the widget. Container grips and resize handles explicitly use `touch-action: none`, so this omission is specific to widget dragging. Apply the touch-action rule to the actual widget drag handle while edit mode is active (without suppressing normal page scrolling outside the handle/edit mode).

- [P1] Unmount portal widget trees when edit mode ends — `apps/nextjs/src/components/board/sections/grid/grid-portal-host.tsx:110`

  `BoardGridPortalRenderer` renders every acquired portal indefinitely, while the registry only prunes entities deleted from the board. Exiting edit mode removes the editor shells but neither clears the registry nor gates the renderer. The static grid then mounts a fresh visible widget tree and the old portal tree remains mounted in detached DOM. In the demo, entering and exiting edit mode retained 36 widget trees in disconnected portal elements while another 36 widget trees were live in the document. Polling widgets, observers, timers, subscriptions, and query hooks therefore continue running in duplicate for the rest of the board session. Nested containers amplify this because detached container portals render their descendants statically while descendant portals are also retained. Clear/unmount portal content when the editor runtime becomes idle, or render the portal map only while edit mode is active.

  ```mermaid
  flowchart LR
    A[View mode: visible static widget tree] --> B[Edit mode: portal widget tree]
    B --> C[Exit edit mode]
    C --> D[New visible static widget tree]
    C --> E[Detached portal widget tree remains mounted]
  ```

## Overall assessment

Changes requested. The local layout engine, collision transaction model, responsive-layout persistence, migration transaction, lazy dnd-kit import boundary, and keyboard fallback are thoughtfully structured, but the three regressions above affect core self-hosted dashboard use: narrow-screen readability, touch editing, and long-lived widget runtime cost after editing.

## UI / browser evidence

QA used the published `ghcr.io/homarr-labs/homarr-test:pr-6503` preview through `go run . run --pr 6503 --demo --detach` at `http://localhost:7576`, authenticated with the provided demo credentials.

- Desktop pointer drag moved a tile from row 1 to row 2; southeast resize changed it from `5 × 3` to `6 × 4`.
- Keyboard edit moved the focused tile down one row; Escape stopped keyboard editing as documented.
- Entering edit mode mounted 36 portal widget trees. After exit, all 36 portal elements were disconnected but still contained their widget trees, alongside 36 visible widget trees.
- At a 412 px viewport the canvas scale was `0.14264264264264265`, horizontal overflow was disabled, and the captured view showed pervasive clipping and overlap.
- A real touch drag could not be dispatched by this Chromium agent-browser session (`navigator.maxTouchPoints === 0` and `(pointer: coarse) === false`). Static inspection plus computed styles confirmed `touch-action: auto` on the widget handle, mount, and shell. The demo board had no containers or rails, so nested/cross-container and rail journeys were reviewed statically rather than exercised in the preview.
- The only console error observed was the expected demo-mode rejection of a widget options mutation; it was not attributed to this PR.

## Test gaps and residual risks

No automated tests, Docker builds, E2E suites, or broad typechecks were run, per the requested focused-review constraints. No source or tests were changed.

Coverage is missing for the edit-to-view portal lifecycle, a real coarse-pointer drag that can receive native `pointercancel`, and a visual/interaction threshold for one-layout boards on narrow viewports. The demo fixture also does not cover nested containers, rails, container removal flattening, or cross-container pointer moves; those paths were inspected statically, but remain higher-risk integration areas.

## Standards vs Spec

- **Standards:** The narrow-canvas behavior violates the PR's own `PRODUCT.md` requirement to prefer horizontal scrolling before controls become inaccessible. The retained detached trees conflict with the React performance guidance to avoid unnecessary mounted work. The Next.js client/server boundary and dynamic editor import are otherwise appropriate for the self-hosted app.
- **Spec:** Desktop pointer, resize, and keyboard editing match the PR description. The documented touch drag is not supported by the widget handle's current touch-action behavior, and the documentation claim that typography remains readable is contradicted by the 412 px preview. Documentation was otherwise updated alongside the new board/container model.

### Screenshot evidence

PR #6503 review screenshots

</details>

<details>
<summary><h2>PR #6545 review — `release/v2` roll-up</h2></summary>

### [P1] Synchronize #6598 before publishing `release/v2` as the v2 image

**Changed path:** `.github/workflows/ci.yml:327` (publication condition), `.github/workflows/ci.yml:337` (the `v2` tag), `.github/workflows/ci.yml:377` (manifest publication)  
**Inherited faulty path:** `scripts/entrypoint.sh:9-18`, especially `scripts/entrypoint.sh:12`

Every push to `release/v2` now publishes the branch as `ghcr.io/homarr-labs/homarr-test:v2`, but the image still contains the old custom-UID startup path:

```sh
find . -name 'node_modules' -prune -o -mindepth 1 -maxdepth 1 -exec chown -R $PUID:$PGID {} +
```

The recursive `chown` descends through the standalone Next.js tree and nested `node_modules`. For the TrueNAS/custom `PUID`/`PGID` deployment called out by #6598, this can turn each update/restart into more than 15 minutes of downtime. The merged fix commit from #6598 (`f1c49d1fbf6f51cd39904e59e8fc03235cb77147`) is not an ancestor of this PR head, and a direct diff confirms its restricted `/appdata`, `.next/cache`, and nginx ownership logic is absent.

This is the unchecked TODO in the PR description. Port #6598 (or its equivalent) before allowing the release workflow to promote this branch.

```mermaid
flowchart LR
  A[push release/v2] --> B[CI publishes :v2]
  B --> C[self-hosted update with PUID/PGID]
  C --> D[recursive chown of /app]
  D --> E[large startup outage]
  F[#6598 targeted-runtime-path fix] -. missing from head .-> D
```

### [P2] Do not preload every production route in the self-hosted server

**Changed path:** `apps/nextjs/next.config.ts:49`

This unrelated dashboard-interaction merge changes `experimental.preloadEntriesOnStart` from `false` to `true`, undoing the explicit idle-memory optimization introduced earlier in the same layer by `8eb673538` (`perf: reduce widget and runtime overhead`). In Next 16.3.1, the production server checks this flag at startup and calls `unstable_preloadEntries()`, which iterates both the Pages and App Router manifests and loads every route module. Homarr has a large management/API surface and is primarily self-hosted, so this front-loads route modules into memory even when a small installation never visits them; low-memory Docker installations can pay higher startup RSS or be pushed toward OOM for no user-facing benefit.

Restore `preloadEntriesOnStart: false`. This preserves lazy route loading and is consistent with the release's stated widget/runtime performance work.

## Overall assessment

**Request changes.** The release roll-up is broadly coherent, the published preview is usable with the documented v2 Workshop endpoint, and I did not find an additional React hydration, RSC-boundary, client-bundle, or request-waterfall regression in the high-risk paths. The missing #6598 synchronization is a release blocker for an important self-hosted deployment class; the production route-preload reversal should also be corrected.

## Scope and ancestry verification

- Reviewed exact PR head `a2219b72f6264a7740e8f69c49b210888b45fb0c` against current `origin/feat/onboarding-rebuild` (`6e078c360642453fa7ac52012e3a7b013a4ed7c5`).
- Current merge-base is `c92e074a6f5d2f7698c33b6c925a27c21aeb5aae` (`chore: update Next.js to 16.3.1`); this release layer contains 73 commits and 438 changed paths.
- The target branch is two commits ahead of the merge-base. Those commits only remove `packages/widgets/src/advanced-focus-support.spec.ts` and `e2e/board-advanced-interactions.spec.ts`; both files are already absent from the PR head through release-side commits, so the structural ancestry drift has no remaining tree omission.
- PR #6582 head `7bafe1ce75f85636ac3707556f968f32dafd8195` is not an ancestor of the PR head. The sibling PR is therefore excluded as stated. The release intentionally includes the separately reviewed audited subset from #6609 (`e842177d9`), so this is not a claim that no ideas from #6582 were ported.
- PR #6598 was merged to `dev` as `f1c49d1f...` and is absent from `release/v2`; see the P1 finding.
- The PR description's CI SHA (`8a44451c...`) is stale. Hosted checks are green on the current exact head, including Fast gate, Container and E2E, both preview architectures, multi-platform publication, Workshop validation/publication, CodeQL, semantic-title validation, and CodeRabbit.
- `gh stack view --json` could not operate in this detached review worktree (`not on any branch`), so ancestry was verified directly with fetched refs, `merge-base`, and `merge-base --is-ancestor`.

## Published-preview browser evidence

The published digest under test was `sha256:d31479f9251e38436526b706578ba123b1aca52f7a491a125d5e7f6427715989` from `ghcr.io/homarr-labs/homarr-test:pr-6545`, running at `http://localhost:7576` with demo credentials `demo` / `demo`.

The final representative harness used the release's documented v2 staging contract:

```sh
go run . --pr 6545 --demo --detach \
  -e HOMARR_WEBSITE_URL=https://v2.preview.homarr.dev \
  -e WORKSHOP_API_URL=https://v2.preview.homarr.dev \
  -e WORKSHOP_WEB_URL=https://v2.preview.homarr.dev/workshop
```

- Login completed and landed on the seeded dashboard.
- Installed Custom Widgets rendered five seeded widgets with Edit, Configure sources, duplicate/export, and publish actions.
- Workshop browse returned live listings through a successful CORS preflight and PocketBase request; the listing detail rendered screenshots, request/network configuration, export, and install controls.
- The page-based publish form rendered the exported widget metadata and screenshot inputs. `Publish now` stayed disabled without a Workshop identity; completing the external GitHub OAuth flow and performing community mutations was outside the demo harness.
- The first run used only `--demo`, so the image's production default targeted `https://homarr.dev`. That host currently serves the production docs without the PocketBase `/api`; browser preflights returned 405 and the UI correctly showed “Homarr Workshop seems to be unavailable.” This is a preview-environment blocker, not a source finding: the operator guide explicitly assigns `https://v2.preview.homarr.dev` to v2 staging, and the configured run worked. Production promotion still depends on deploying the combined Workshop service at the production default.
- Chromium initially needed the documented `--no-sandbox` host workaround. No application console exception or hydration error was observed after the browser daemon connected.

Evidence:

- `evidence/after-login.png`
- `evidence/custom-widgets-installed.png`
- `evidence/workshop-v2-preview.png`
- `evidence/workshop-detail.png`
- `evidence/workshop-publish.png`
- `evidence/workshop-github-demo-block.png`

## Standards review

- Read `AGENTS.md` and `apps/docs/README.md`; evaluated Homarr as a Docker-first, self-hosted application for advanced users.
- Checked release CI/publication, Workshop runtime URL configuration and docs, auth pages, backup import/custom migrations, instrumentation/Next 16.3.1, permission recalculation, Docker endpoint setup, advanced-widget compatibility, media/widget fixes, and later release merges.
- Applied the Next and Vercel React review criteria to RSC/client boundaries, parallel requests, hydration, serialization, bundle loading, rerender patterns, and self-hosted startup behavior. The production preload flag is the concrete framework-performance regression found.
- Documentation is present for Workshop operation/runtime URLs, Docker endpoint behavior, permissions, widgets, and other user-facing release changes. The v2 staging/production Workshop distinction is documented consistently.
- `git diff --check` passed. No source files were edited.

## Spec review

- Release-specific application and Workshop image publication is present for `release/v2`.
- Workshop browse/detail/publish routes, advanced Custom Widget compatibility, auth-shell updates, backup migration handling, Next 16.3.1, and the later audited release fixes are present.
- Full #6582 ancestry is excluded; #6609 is a deliberate audited subset.
- The explicit #6598 synchronization TODO is not satisfied and is the P1 spec failure.

## Test gaps and residual risk

- Per instruction, no source build, broad typecheck, unit suite, Docker compilation, or E2E suite was run. Existing hosted checks are green on the exact head.
- Demo mode and external OAuth prevented authenticated vote/report/publish mutations. Onboarding restore could not be re-entered safely on the already seeded demo volume, and no real integration target was available for an end-to-end embedded-integration save.
- Backup restore/custom migrations were reviewed statically only; multi-dialect historical backups remain the highest unexercised compatibility surface.
- Before production promotion, deploy the `ghcr.io/homarr-labs/workshop:v2` digest at `homarr.dev`, then repeat the operator guide's health, public-listing, GitHub sign-in, and install smoke checks against the production origin.

### Screenshot evidence

PR #6545 review — `release/v2` roll-up screenshots

</details>

<details>
<summary><h2>PR #6555 review — widget performance (stack layer 6/8)</h2></summary>

### [P2] Bound Umami event concurrency instead of awaiting every event serially — `packages/integrations/src/umami/umami-integration.ts:249`

`getMultiEventTimeSeriesAsync` changed from `Promise.all` to a `for` loop with an `await` in every iteration. The event picker has no selection cap, so response latency is now the sum of every selected event request (and each Cloud request can itself page through raw events). For example, eight selected events against a 200 ms self-hosted Umami instance move from roughly one round-trip to at least 1.6 seconds; 20 events take at least four seconds before Homarr can return any series. This directly introduces the request waterfall that this PR says it removes.

```mermaid
flowchart LR
  Q[Homarr widget query] --> E1[Event A request]
  E1 --> E2[Event B request]
  E2 --> EN[Event N request]
  EN --> R[Return all series]
```

Use Umami's aggregate event-series endpoint when compatible, or run a small bounded pool so a modest self-hosted server is protected without making widget latency scale linearly with the number of selected events.

### [P2] Retain each Media Missing tab's scroll position when unmounting it — `packages/widgets/src/media-missing/component.tsx:171`

`keepMounted={false}` removes the inactive `Tabs.Panel`, including its `ScrollArea` DOM and scroll offset. On a long missing-media list, a user can scroll deep into Missing, inspect Queued, then return to Missing and be dropped at the first item. The previous mounted panels retained both list positions. This is an introduced lazy-lifecycle/UX regression in the exact high-item-count case where the optimization matters most. Cache and restore each panel's offset, virtualize the lists, or keep the panel mounted while deferring its expensive children without discarding navigation state.

## Assessment

The layer is otherwise internally consistent and keeps the reviewed all-or-nothing error semantics. Parallel integration calls use independent requests; TrueNAS adds correlation for concurrent WebSocket responses; observers/listeners/timers have cleanup; the 10-second polling definitions match the handler cache window; query inputs remain structurally distinct; and the Anchor Note editor split does not broaden the initial client boundary. The two findings are performance/UX regressions caused by optimization tradeoffs, not unrelated pre-existing behavior.

| Standards review | Spec review |
| --- | --- |
| No additional repository-style, React boundary, Next.js convention, or documentation-sync violation found. `git diff --check` passes. | Most listener, observer, allocation, payload, polling, and lazy-load changes preserve behavior. The Umami loop conflicts with the stated waterfall goal, and Media Missing does not preserve visibility lifecycle state. |

## Optimization audit

| Area | Assumption challenged | Result |
| --- | --- | --- |
| Backend parallelism | Concurrent calls must preserve dependency and failure semantics. | Passed for AdGuard, DashDot, NZBGet, SABnzbd, media requests, Tdarr, and TrueNAS; dependent TrueNAS network work remains ordered internally. |
| Umami requests | Reducing fanout must not create user-visible linear latency. | Finding: unbounded serial waterfall. Official Umami metrics responses are direct arrays, so the event-name endpoint-shape candidate was discarded. |
| Polling/cache | The browser must not poll cached handlers faster than their 10-second refresh window. | Passed; published-preview response dates showed DNS/downloads/health/media requests at approximately 10-second cadence. |
| List/tab lazy lifecycle | Reclaimed inactive DOM must not discard navigation state. | Finding: Media Missing loses the inactive list's scroll offset. |
| Editor/media lazy loading | Deferred modules/assets must retain compact rendering, hydration, and advanced-view behavior. | Static paths passed. The demo board did not configure a changed iframe, Immich carousel, RSS enclosure, or release poster, so those network deferrals remain a runtime gap. |
| Render propagation | Hoisted components and removed observers must reduce churn without stale values. | Passed statically. The Notebook refactor still has one selection listener per indent control, so it is not a measurable listener reduction, but it is not an introduced regression. |
| Next startup preload | Lower self-hosted startup memory must justify first-hit route latency. | Intentional tradeoff: `preloadEntriesOnStart: false` matches current Next server behavior, and the PR explicitly disclaims board-load latency improvement. No correctness finding. |

## Browser and performance evidence

- Published image only: `ghcr.io/homarr-labs/homarr-test:pr-6555`, launched with `go run . run --pr 6555 --demo --detach`; container `homarr_pr_6555` exposes `http://localhost:7576`, with `DEMO_MODE=true` and `demo` / `demo` credentials. No source image was built.
- Representative demo board loaded calendar, downloads, system/storage, apps, media, DNS, releases, RSS, notebook, Docker, weather, and assistant surfaces. A clean reload produced no page errors or console errors.
- Calendar advanced view opened and closed with keyboard focus restoration at desktop size and at `390x700`; the mobile surface remained usable. Evidence: **initial board** (see screenshot evidence below), **mobile advanced Calendar** (see screenshot evidence below).
- Core load profile from the published production image: TTFB `114.5 ms`, FCP `204 ms`, LCP `1,312 ms` (provisional), CLS `0.19`. These are single local samples, not a base/head comparison.
- React DevTools initialization succeeded for Suspense/render commands. Suspense reported 39 dynamic holes and no primary blocker. A settled five-second demo-board sample reported 104,702 fiber renders across 200 component names, but the board intentionally changes mock stats every second and the production bundle is minified; without a base sample this cannot be attributed to this layer and is not a finding. `react tree` returned no rows, and hydration timing was unavailable because the published image is not a React profiling build.
- Compact Media Releases rendered the seven changed items with `content-visibility: auto`. Native lazy-loading behavior could not be exercised because the demo board had no changed lazy iframe, Immich thumbnail strip, RSS enclosure, or enabled release poster.

## Residual and test gaps

- No reproducible benchmark artifacts or base/head measurements are committed or linked; the PR-body memory/CPU figures cannot be independently audited from this layer.
- The 509-line advanced-interactions E2E file and the advanced-focus support contract spec are deleted without replacement. Published-preview smoke coverage passed, but automated regression coverage for hover/keyboard/focus/portal lifecycle is reduced.
- Umami Cloud pagination and truncation behavior is unit-covered with mocks, but no live Cloud account was available.
- No local unit, E2E, full typecheck, Docker build, or broad test suite was run, per the focused-review instruction. Current GitHub checks were green at review time.

## Review scope

- PR head: `bfd0bb7f809baa86bc39b680ad65d157e24c3af2`
- Resolved target: `origin/feat/homarr-assistant`
- Merge-base reviewed: `111a72c28349893b0f84b101e1c5239fb53ff2fd`
- Entire 61-file merge-base diff, changed tests, call sites, PR description, relevant Vercel React rules, relevant Next.js references/source, repository rules, and `apps/docs/README.md` were reviewed.
- These internal performance changes do not add or change a documented widget, integration, API, environment variable, CLI command, auth flow, or cron job, so no user-documentation delta is required.

### Screenshot evidence

PR #6555 review — widget performance (stack layer 6/8) screenshots

</details>

<details>
<summary><h2>PR #6569 review — onboarding rebuild (stack layer 7/8)</h2></summary>

**Assessment: Request changes.** Four concrete P2 regressions were found: two onboarding state/error UX bugs and two self-hosted performance risks.

Review range: merge-base `d091362db2f853e67fad2089eab6dadf95b3d5c3` through PR head `6e078c360642453fa7ac52012e3a7b013a4ed7c5` (265 changed paths). The current remote target tip `origin/feat/widget-performance-v2` is `bfd0bb7f809baa86bc39b680ad65d157e24c3af2`; it has diverged from the merge-base by a corresponding final test commit, so findings are based on the PR's merge-base diff rather than `base..head` alone.

### [P2] Do not silently discard selected integrations when Review is opened from the rail

**Path:** `packages/onboarding/src/setup-studio.tsx:436`

`applySetupAsync` reduces all drafts to `drafts.filter(isIntegrationDraftComplete)`. The explicit “skip unfinished integrations” confirmation only runs from the Connect section's Continue button (`:743`), while every timeline button can select Review directly (`:697`) and the keyboard handler can jump there with `End`. This creates a deterministic bypass:

```text
Select Sonarr (API key still blank)
├─ Connect → Continue → explicit Cancel / Confirm dialog  ✅
└─ Click Review rail item (or press End)
   └─ Review says “Integrations: None selected”
      └─ Build my board → Sonarr is filtered out silently  ❌
```

The published preview reproduced both branches. Advanced users are especially likely to navigate through the rail and can finish setup believing the service they selected will be configured. Enforce the same confirmation at every transition into Review or at Build time, or keep incomplete selections visible in Review and require an explicit decision there.

Evidence: `evidence/incomplete-integration-skipped.png`.

### [P2] Surface completion errors instead of returning to an unchanged Review screen

**Path:** `packages/onboarding/src/setup-studio.tsx:537`

The catch block discards the error, resets progress, and only plays an error sound. There is no notification or inline error state. Any server failure—stale/renamed board, transaction error, conflict, or unavailable mutation—therefore leaves the user on an enabled “Build my board” button with no explanation or recovery guidance.

The preview supplied a concrete failure: the demo server rejected `onboard.completeSetup` with `Mutations are disabled in demo mode`; the spinner stopped and the Review page remained unchanged with no visible or accessible error message. Catch the error value and render a localized inline alert or notification containing the actionable server message plus a translated fallback.

### [P2] Do not fetch the entire integration catalog for decorative onboarding chrome

**Path:** `packages/ui/src/components/integration-marquee.tsx:25`

The marquee eagerly renders every integration icon twice, and `OnboardingBackdrop` mounts two marquees (`packages/onboarding/src/onboarding-backdrop.tsx:8`). On a fresh authenticated `/init` load, browser instrumentation measured:

- 205 eager `` elements, none with `loading="lazy"`;
- 52 unique remote dashboard-icon URLs and 52 image requests;
- 437,453 compressed response bytes for those decorative icons (56 images / 532,719 bytes total on the route).

The images are `aria-hidden` decoration, but their work scales with the integration catalog and is paid on Homarr's most latency-sensitive first-run route. It also creates a CDN dependency for a self-hosted setup experience. Use a small curated/local icon set, a sprite/composited asset, or otherwise cap and defer the decoration instead of mapping the full catalog.

### [P2] The “bounded” restore parser still multiplies a 256 MB upload in memory

**Path:** `apps/nextjs/src/app/api/backup/import/route.ts:87`

The new request guard keeps every stream chunk, copies all chunks into another contiguous `Uint8Array`, then asks `Response.formData()` to materialize a `File`. The route subsequently calls `file.arrayBuffer()` for `AdmZip` at `:260`. Near the accepted limit, the process can retain several full-size representations of the same ~256 MB upload before ZIP extraction (whose accepted database can be 512 MB):

```text
request chunks (~257 MB)
        ↓ full copy
contiguous Uint8Array (~257 MB)
        ↓ multipart materialization
FormData File (~256 MB)
        ↓ file.arrayBuffer()
AdmZip input (~256 MB) + extracted DB (up to 512 MB)
```

This can exceed common 512 MB–1 GB Homarr container limits and kill the server during the recovery operation. The guard prevents unbounded input size, but not unsafe peak memory. Stream multipart data to a bounded temporary file (and open the ZIP from disk), or lower the advertised limits to a peak proven safe under the supported container memory floor.

## Grill decision-tree assessment

The review stress-tested the settled product assumptions rather than reopening the redesign:

- **Selection intent:** selecting a service is durable intent until the user explicitly confirms dropping it. Finding 1 breaks that invariant on rail/keyboard paths.
- **Recovery visibility:** every first-run failure must explain what failed and leave a clear recovery action. Finding 2 breaks that invariant.
- **Self-hosted constraints:** first-run decoration must not grow linearly with the integration catalog, and a bounded upload must remain bounded in peak memory. Findings 3 and 4 break those invariants.
- **Destructive action:** database restore must require an explicit phrase after preview. The preview correctly kept this gate.
- **Atomic setup:** discovery, board creation/update, apps, integrations, secrets, layouts, settings, everyone-group defaults, onboarding transition, and claim deletion are protected by a compare-and-set transaction. The async and SQLite implementations roll back together on failure; no transaction/idempotency defect was found.

## Browser evidence and blockers

- Published image only: `ghcr.io/homarr-labs/homarr-test:pr-6569`, container `homarr_pr_6569`, `DEMO_MODE=true`, `http://localhost:7575`, credentials `demo` / `demo`.
- The current CLI requires the `run` subcommand, so the working invocation was `go run . run --pr 6569 --demo --detach`.
- Demo mode seeds onboarding at `finish`. To inspect the studio without modifying the image, the disposable demo database was used as a harness by changing only its onboarding marker back to `setup`; the published code and demo environment remained unchanged.
- Exercised desktop and 390×844 responsive layouts; keyboard timeline navigation (`Arrow`/`Home`/`End`); Connect confirmation and rail bypass; Discover partial/unavailable states; board target/appearance preview; Workshop, Assistant, and MCP tabs; manage apps/integrations/search engines/boards; debounced search; backup preview; and the exact `I understand` restore gate.
- No JavaScript exceptions occurred in the changed manage/backup paths. Completion testing was blocked exactly by demo read-only behavior (`Mutations are disabled in demo mode`), which also exposed Finding 2. Docker discovery reported `EACCES /var/run/docker.sock`, so only its partial/unavailable UX and static server logic were reviewed.
- Evidence files: `evidence/incomplete-integration-skipped.png`, `evidence/onboarding-review-mobile.png`. The temporary demo backup used to drive the restore preview was moved to trash and remains recoverable there.

## Standards review

- `git diff --check` is clean.
- Next.js 16 async request APIs, RSC/client boundaries, server-side auth checks, claim-cookie validation, hydration handling, and route-handler authorization were applied correctly in the inspected changes.
- Independent capability probes and integration validations are parallelized. The setup transaction uses compare-and-set transition guards and stable source-derived IDs.
- React/Next performance is not acceptable for the all-catalog decorative marquee (Finding 3). The 2,022-line client studio and eager `canvas-confetti` imports remain bundle/rerender risks, but no additional concrete regression was asserted without route bundle attribution.
- No unrelated source edits or tests were added.

## Spec review

- The new setup studio, Docker-label discovery, first-board automation, Workshop/Assistant/MCP exposure, manage collections, board preview, explicit database restore confirmation, localized color reset, board name/target validation, and Next.js 16.3.1 migration are represented in code and docs.
- Removal of the pre-1.0 importer is deliberate and documented in the migration guide/blog; it is not an accidental missing upgrade path in this layer.
- The explicit unfinished-integration confirmation is not applied consistently (Finding 1), and completion failures are not communicated (Finding 2).
- Documentation correctly describes atomic board/setup persistence and rollback. It does not mitigate the restore route's unsafe peak-memory implementation (Finding 4).

## Test gaps and residual risk

Per instruction, no unit, E2E, Docker source build, broad typecheck, or test suite was run. Existing changed tests and their call sites were inspected only.

- A true empty-volume claim → account creation → setup completion could not be exercised under demo mode because demo seeds a user and the `finish` step.
- Successful commit/rollback was not runtime-tested against SQLite, MySQL, and PostgreSQL; confidence there comes from transaction and targeted test inspection.
- Concurrent completion, restore restart/readiness, a maximum-size restore, live Docker/Kubernetes discovery, Assistant provider writes, and MCP key creation remain runtime residual risks.
- The tests cover the normal Connect confirmation but do not appear to assert the rail/keyboard bypass or visible completion-error feedback. A focused UI test for each would prevent recurrence, but no tests were added because they were not requested.

### Screenshot evidence

PR #6569 review — onboarding rebuild (stack layer 7/8) screenshots

</details>


ajnart:
@coderabbitai review

ajnart:
@coderabbitai review

ajnart:
@coderabbitai review

PR #6759, opened 2026-09-02
feat(plugins): add llama.cpp integration and widget

leafyelin:
**Description**

Adds a first-class llama.cpp (llama-server) integration and a matching widget, so dashboards can monitor any OpenAI-compatible llama.cpp server over its plain HTTP endpoints — no extra sidecar required.

**What it does**

- New integration (`packages/integrations/src/llama-cpp/`): connects to llama-server's built-in endpoints
  - `/health` — liveness
  - `/v1/loads` — model name, context size, KV cache
  - `/metrics` — Prometheus gauges: token generation rate, prompt processing rate, load
  - `/slots` — per-slot KV cache usage and in-flight request state
- New widget (`packages/widgets/src/llama-cpp/`): model name + status, generation speed with a three-tier display (in-request running average → live gauge → process-lifetime average, with tooltip explaining which is shown), context/KV cache usage with optional per-slot breakdown
- API route, request handler, definitions registration and translations (en/zh/cn) wired in
- 21 unit tests for the integration's response mapping, including per-request speed calculation

**Why per-request running average?**

llama-server's `/metrics` rate gauge resets to 0 while idle and reflects a rolling bucket while busy, so a dashboard polling it cannot tell "fast now" from "fast earlier". The widget instead diffs `next_token[0].n_decoded` across polls for the slot's current `id_task` (verified against llama.cpp source, `server-context.cpp`), giving the true average speed since the current request started — matching what the llama.cpp Web UI shows during generation.

**Manual testing performed**

- [x] Integration tested against a real llama-server (b10679-1) running a Qwen 27B GGUF (n_ctx 131072): health, loads, metrics and slots endpoints verified
- [x] Per-request speed verified end-to-end: `id_task` stable during a generation, `n_decoded` increases 0→137 over ~4.4s (~31 t/s), resets on request end
- [x] Widget deployed on a production Homarr instance (Docker, NAS) — KV usage, speed tiers and tooltips render correctly across busy/idle transitions
- [x] `pnpm turbo run build` (3/3), typecheck (4 packages), lint and format all green; 21/21 integration unit tests pass

**Additional context**

None — this is self-contained: it only adds new files plus the standard registration points (definitions, translations, exports).

## Summary by CodeRabbit

* **New Features**
  * Added a llama.cpp integration for monitoring local LLM server health, loaded models, request activity, generation speed, context usage, and token metrics.
  * Added a configurable llama.cpp dashboard widget with model details, status, performance, and cache statistics.
  * Added an AI integration category with English, Simplified Chinese, and Traditional Chinese translations.
* **Documentation**
  * Added setup and configuration guides for the llama.cpp integration and widget.


ajnart:
@leafyelin It's a shame you didn't open an issue for this earlier to discuss it... We already have an OpenAI-compatible way of interacting with llama.cpp in v2 (see pinned issue) I would really appreciate it if you could rebase/re-shape this PR based off the `release/v2` branch instead https://github.com/homarr-labs/homarr/issues/6600 

It would be nice to get your implementation into shape to enhance the "homarr assistant" feature to suit your needs instead of making a whole new integration & widgets. What do you think ?

leafyelin:
Thanks for the quick review @ajnart. Two things:

Rebasing onto release/v2— agreed, I'll re-target this. (Happy to move it.)

On reshaping it into the assistant— I think there's a small mismatch worth clarifying: the assistant's OpenAI-compatible path (custom provider) lets you chat withllama.cpp, which is great. But this PR is about monitoringllama-server — KV cache usage, per-request token speed, and Prometheus load. Those only exist on llama-server's private /slotsand /metricsendpoints; the OpenAI /v1channel doesn't expose them. So the monitoring widget is complementary to the assistant, not a duplicate.

Suggestion: I'll rebase onto release/v2and keep the monitoring integration, andadd a llama.cpppreset to the assistant provider list (alongside Ollama / LM Studio) so the chat path is first-class too. Does that split work for you?

leafyelin:
@ajnart Done — rebased onto `release/v2` and reshaped as discussed. New PR: **#6765** (this one is closed because the head branch was recreated during the rebase, so it can't be reopened).

One scoping note on the assistant reshape: the assistant's OpenAI-compatible path lets you *chat with* llama.cpp, but this PR is about *monitoring* llama-server — KV cache usage per slot, per-request token speed, Prometheus `/metrics` load — which only exists on the private `/slots` and `/metrics` endpoints, not `/v1`. So the monitoring widget complements the assistant rather than duplicating it, and I've added a first-party `llama-cpp` preset to the assistant's provider list (alongside Ollama / LM Studio) so the chat side is covered too.

All checks are green locally (typecheck, 261 tests, assistant suite, lint) and smoke-tested against a live llama-server. Happy to adjust if you'd rather I drop the monitoring part and keep only the assistant preset here.

PR #6801, opened 2026-09-07
feat(widgets): add downloads speed unit toggle for bits vs bytes

Alien10140:
## Summary
- Adds a **Speed unit** option on the Download Client widget: **bytes** (KiB/s, MiB/s, default) or **bits** (SI kbps/Mbps).
- Integrations still report byte/s; `formatBitRate()` multiplies by 8 and scales by 1000. Size / received / sent columns stay on `formatBytes`.
- Default is bytes so existing boards do not change. Oldmarr import maps the new option to `undefined` (then the default).

## Test plan
- [x] Unit tests: `formatBitRate(125)` → `1.0 kbps`, `formatBitRate(125_000)` → `1.0 Mbps`
- [x] Built a local image and ran the widget against live qBittorrent + NZBGet
- [x] With **bits** selected, seeding rates render as kbps (e.g. 22154 B/s → 177.2 kbps); size columns remain in GiB
- [x] Widget edit UI shows "Speed unit" with Bytes / Bits labels
- [ ] Switch the option back to bytes and confirm KiB/s / MiB/s
- [ ] Pause / resume still work from the widget footer

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added a Downloads widget setting to display transfer speeds in either bytes per second (KiB/s, MiB/s) or bits per second (kbps, Mbps).
  * Applied the selected speed unit consistently across tables, tooltips, expanded views, statistics, and footers.
  * The setting defaults to bytes per second and includes descriptive labels for both formats.

### DIFF
--- a/apps/docs/docs/widgets/downloads/index.ts
+++ b/apps/docs/docs/widgets/downloads/index.ts
@@ -47,6 +47,13 @@ export const downloadsWidget: WidgetDefinition = {
         defaultValue: "no",
       },
       {
+        name: "Speed unit",
+        description:
+          "How download and upload rates are displayed. Bits convert byte/s from the client to SI kilobits/megabits (kbps, Mbps).",
+        values: "bytes (KiB/s, MiB/s) or bits (kbps, Mbps)",
+        defaultValue: "bytes",
+      },
+      {
         name: "Show usenet entries marked as completed",
         description: "This will show entries that have been completed in your Usenet client.",
         values: { type: "boolean" },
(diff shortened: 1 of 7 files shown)


ajnart:
Thanks for the PR @Alien10140 ! I think I would rather have this option as a personal preference for user settings and make it used globally instead of doing a per-widget preference 

Would you mind seeing if you can rebase this on the release/v2 branch ? I've done explorations on having this as a preference

Alien10140:
Sounds good. I'll work on that change!

ajnart:
@Alien10140 I have integrated this into v2 as a global user preference. feel free to check it out there instead

PR #6848, opened 2026-09-14
feat(plex): surface newly added episodes in media releases widget

### LINKED ISSUE #6847: feat(plex): media releases widget should list newly added episodes, not seasons

The Plex "Media releases" widget never shows new episodes added to an existing season.

`PlexIntegration.getMediaReleasesAsync()` calls `/library/recentlyAdded`, which returns **seasons** (type `season`) with the season's own `addedAt`. Adding episodes to a season that already exists does not change that value, so the season never moves up and the widget looks stale.

Example from my server (Plex 1.42, Homarr 1.77.1):

```
/library/sections/1/all?type=4&sort=addedAt:desc   (episodes)
09-14 09:19 | Futurama | Oceans Three
09-14 09:19 | Special Ops: Lioness | Kiss the Girls

/library/recentlyAdded   (what Homarr uses, newest first)
09-09 13:41 | Cambio de marcha | Season 2
09-09 13:41 | Cambio de marcha | Season 1
```

Proposal: for show libraries, fetch episodes instead of seasons, e.g. `/library/sections/{id}/all?type=4&sort=addedAt:desc` (or `/hubs/home/recentlyAdded?type=4`, which is what the Plex apps use), and map them to `MediaRelease` with `type: "tv"`, title = `grandparentTitle`, subtitle = `S{parentIndex}E{index} – {title}`. Optionally group consecutive episodes of the same season added in the same scan.

Related: the widget also uses `originallyAvailableAt` as `releaseDate`, so an old title added yesterday sorts/labels as old. Using `addedAt` (or exposing both) would make "recently added" match what Plex shows.

### Additional information

Related past issues: #3885, #3692, #6083 (all about season handling in this widget).

GreenMonito:
Closes #6847

## Problem

`PlexIntegration.getMediaReleasesAsync()` only reads `/library/recentlyAdded`, which returns **seasons** with the season's own `addedAt`. Episodes added to a season that already exists never change that value, so the media releases widget never shows them (the widget looks stale for anyone following ongoing shows).

## Change

- Additionally fetch the most recently added episodes of every `show` library (`/library/sections/{id}/all?type=4&sort=addedAt:desc`, one request per show library, 50 items each).
- Seasons already in `/library/recentlyAdded` that received newer episodes are bumped to the episode's `addedAt` and use its `originallyAvailableAt` as release date.
- Seasons that fell out of `/library/recentlyAdded` but got new episodes are rebuilt from the episode metadata (`grandparentTitle` / `parentTitle`, `grandparentThumb` + `art` as images, link to the season).
- Result is sorted by `addedAt` desc and capped at 50 (Plex's default page size), so the widget behaves like Plex's own "Recently added" and like the Jellyfin integration (`Items/Latest` groups episodes into their series).
- Any failure while fetching episodes is logged and falls back to the previous behaviour, so servers/tokens without library access are unaffected.
- Small refactor: `fetchJsonAsync()` helper shared by the JSON endpoints, `recentlyAddedItemSchema` extracted so it can be reused.

## Tests

- Existing spec unchanged and passing (episode fetch fails in its mock → fallback path).
- New spec: bumped season ordering + release date, season rebuilt from episode metadata (title, images, href), and fallback when the episode endpoints error.

`pnpm --filter @homarr/integrations typecheck`, `lint` and `vitest run packages/integrations/src/plex` pass.

## Summary by CodeRabbit

- **New Features**
  - Plex recently added content now includes the latest episodes for each show.
  - Seasons missing from Plex’s recent-items list can now be recovered from episode information.
  - Season dates and ordering reflect the newest available episode, with results limited to the 50 most recent items.
  - Season detail links now use cleaner, more reliable URLs.

- **Bug Fixes**
  - Plex integrations now gracefully fall back to the available recent-items data when episode details cannot be retrieved.


ajnart:
Would you mind to rebase this on release/v2 ? it's the newest changes for homarr v2

GreenMonito:
Rebased onto `release/v2` and retargeted the PR (single commit 3597695a4, no functional changes; only the spec's trailing newline conflict). Typecheck, lint and `vitest run packages/integrations/src/plex` pass on top of `release/v2`.

packages/integrations/src/plex/plex-integration.ts
```diff
@@ -242,20 +243,109 @@ export class PlexIntegration extends Integration implements IMediaServerIntegrat
...
+  }
+
+  private async getRecentlyAddedEpisodesAsync(): Promise<RecentlyAddedEpisode[]> {
+    const sections = await librarySectionsSchema.parseAsync(await this.fetchJsonAsync("/library/sections"));
+    const showSections = sections.MediaContainer.Directory?.filter((section) => section.type === "show") ?? [];
+
+    const episodesPerSection = await Promise.all(
+      showSections.map(async (section) => {
+        const data = await recentlyAddedEpisodesSchema.parseAsync(
+          await this.fetchJsonAsync(
```
ajnart, line 326: Why not use `fetchWithTrustedCertificatesAsync` ?

GreenMonito: It does: `fetchJsonAsync` is a small private wrapper around `fetchWithTrustedCertificatesAsync` (see the definition further down in the diff) that only adds the `X-Plex-Token` / `Accept: application/json` headers and the `response.json()` call, since the same three lines were now needed for `/identity`, `/library/recentlyAdded`, `/library/sections` and every show library. I added a doc comment on it in 926251c2f to make that obvious. If you'd rather keep the calls inlined like the original code, happy to drop the helper.

packages/integrations/src/plex/plex-integration.ts
```diff
@@ -240,20 +241,109 @@ export class PlexIntegration extends Integration implements IMediaServerIntegrat
...
+                  ...(poster ? [{ type: "coverPoster", url: poster }] : []),
+                  ...(backdrop ? [{ type: "background", url: backdrop }] : []),
+                ]
+              : undefined,
+        };
+      });
+
+    return [...bumpedItems, ...missingSeasons]
+      .toSorted((itemA, itemB) => itemB.addedAt - itemA.addedAt)
+      .slice(0, Math.max(items.length, RECENTLY_ADDED_LIMIT));
```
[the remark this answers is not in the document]
GreenMonito: Done in 926251c2f, sliced at `RECENTLY_ADDED_LIMIT`.

packages/integrations/src/plex/plex-integration.ts
```diff
@@ -240,20 +241,109 @@ export class PlexIntegration extends Integration implements IMediaServerIntegrat
...
+      }),
+    );
+
+    return episodesPerSection.flat().toSorted((episodeA, episodeB) => episodeB.addedAt - episodeA.addedAt);
+  }
+
+  private async fetchJsonAsync(path: `/${string}`): Promise<unknown> {
     const token = super.getSecretValue("apiKey");
-    const response = await fetchWithTrustedCertificatesAsync(super.url("/identity"), {
+    const response = await fetchWithTrustedCertificatesAsync(super.url(path), {
```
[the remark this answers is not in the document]
GreenMonito: Done in 926251c2f: `fetchJsonAsync` now forwards a `timeout` and the `/library/sections` + per-library episode requests use a 10s timeout, so a stalled library rejects and the seasons-only fallback runs.

packages/integrations/src/plex/plex-integration.ts
ajnart: you're right, i forgot about that

ohmzi:
@GreenMonito  can you please take a look at some edge cases, specially the connectivity error handling ones. 

- [ ] Duplicate cards. seasonKeyOf() (:48, :287) ignores episode and show items that /library/recentlyAdded can return. So the same content shows up twice, e.g. "Futurama / Oceans Three" and "Futurama / Season 11".
- [ ] One failing library disables the feature for all libraries. Promise.all (:323) drops every library's episodes if one returns an error or times out, and the widget silently falls back to the old behaviour.
- [ ] Timeout covers the headers only. withTimeoutAsync clears its timer once fetch() resolves, so reading the body (response.json()) is only limited by undici's 300s default. /identity and /library/recentlyAdded have no timeout at all.
- [ ] Added latency on every refresh. The episode requests run after the main calls instead of alongside them (:167-169), adding up to about 20s before the fallback kicks in.
- [ ] No response.ok check. In fetchJsonAsync (:341-352), a 401 or 404 is reported as "Failed to parse integration data" instead of a response error. The PR's fallback test only passes because the resulting SyntaxError gets swallowed.
- [ ] Inconsistent cap. Results are cut to 50 only when episodes are found; the fallback path returns everything (:312-314). Jellyfin and Emby use 100.
- [ ] There's no episode info (SxxEyy – title) and cards still show air dates.
- [ ] Misleading dates. The list is sorted by date added but labelled with air date, so a backfilled old episode puts "2008-01-20" at the top. Date-only values also show the previous day with a fake time for users west of UTC.
- [ ] Cost per refresh. It's now 3+N requests, and each one re-reads the certificate folder, runs a DB query and opens a new TLS connection.

GreenMonito:
Thanks for the thorough pass @ohmzi, all addressed in aeaf81426 except one I'd like your opinion on.

- [x] **Duplicate cards.** `seasonKeyOf` now maps `episode` entries through their `parentKey`, and synthesized seasons are skipped when `/library/recentlyAdded` already lists that season or the show (`grandparentKey`). An episode entry that got a newer sibling is bumped in place instead of gaining a second card. Covered by a new test.
- [x] **One failing library.** `Promise.allSettled` per library, each rejection logged with its section key, the libraries that answered still contribute. New test with one library returning 500.
- [x] **Timeout covers the headers only.** All Plex JSON requests now go through one dispatcher created with `bodyTimeout: 10s`, plus the existing 10s header timeout, and `/identity` and `/library/recentlyAdded` use it too.
- [x] **Added latency.** `/identity`, `/library/recentlyAdded` and the episode requests now run in one `Promise.all`, so the episodes cost nothing on top of the existing calls.
- [x] **No `response.ok` check.** `fetchJsonAsync` throws `ResponseError` like the Emby integration. The fallback test now exercises that path instead of a swallowed `SyntaxError`.
- [x] **Inconsistent cap.** The cap moved to `getMediaReleasesAsync`, so it applies to both paths, and is now 100 to match Jellyfin and Emby. Episodes are still paged at 50 per library, which is enough to find the newest episode of every recently updated season.
- [x] **Episode info.** Entries surfaced or bumped by an episode are labelled `S11E05 – Oceans Three` (falls back to the season title when Plex omits the indices).
- [x] **Fake time / previous day.** Date only values such as `2008-01-20` are parsed as local calendar dates instead of UTC midnight. The widget still renders `hour`/`minute` for every integration, which is where the remaining `00:00` comes from — happy to open a separate PR for that if you want it dropped when the source has no time.
- [ ] **Sorted by added, labelled with air date.** Left as is for now: `originallyAvailableAt ?? addedAt` is what Jellyfin (`PremiereDate ?? DateCreated`) and Emby use, so switching Plex to `addedAt` would make it the odd one out and would also change movies, not just episodes. If you'd rather have the label match the sort I'll switch it here and can follow up with the other two.
- [x] **Cost per refresh.** It was 3+N agents, not just requests: every `fetchWithTrustedCertificatesAsync` call without a dispatcher reads the certificate folder and queries `trustedCertificateHostnames`. There is now a single agent per refresh, closed in a `finally`, so it is one certificate read, one DB query and a pooled connection for all requests.

`pnpm --filter @homarr/integrations typecheck`, `lint`, `oxfmt --check` and `vitest run packages/integrations/src/plex` (6 tests) pass.

